PR #77, opened 2016-07-28
Update BlocktradesDeposit.jsx input type="tel"

pkattera:
A quick UX tweak, to enable the numeric keypad to pop up on mobile (instead of the normal qwerty keyboard).


roadscape:
Is `type="number"` more appropriate than `type="tel"`?

pkattera:
type="number" is a a better solution that type="text". but type="tel" will give users the full number pad which makes it super simple to enter numbers. For more context, see http://mobileinputtypes.com/

PR #78, opened 2016-07-29
Pull Request: Adding Cancel All Button 

JasonMcz:
**Goal:**
```
-an easier cancellation process for GUI traders on internal market 
-cancelling 10+ orders one by one can be difficult on the GUI level especially with the each reload the order sequence changes after the table gets  re-constructed 
```

**Changed:**
```
-looping through open_orders array 
-cancel each individual orders
-notify users about cancellation ( might be better just send one message after map is done?)
-reload the front-end
```

Note: Please take a look, I didn't have `STEEMD` on my environment so I wasn't able to test the function by running it locally.

### DIFF
--- a/app/components/pages/Post.jsx
+++ b/app/components/pages/Post.jsx
@@ -80,7 +80,7 @@ class Post extends React.Component {
                 </div>
 
                 //@Jasonmcz
-                //Comment: Perhaps it would be good to havd a render/timeout
+                //Comment: Perhaps it would be good to HAVE a render/timeout
                 //Because it takes about 5-15 seconds right now to load an account
                 //which pretty much shows up the promotion of $10.00 STEEM every single time
                 //perhaps add a timeout here prior to render this?
--- a/app/components/pages/Post.jsx
+++ b/app/components/pages/Post.jsx
@@ -58,7 +58,7 @@ class Post extends React.Component {
         let sort_menu = [];
 
         let selflink = '/' + rout_params.category +'/@'+ rout_params.username + '/' + rout_params.slug;
-        for( let o = 0; o < sort_orders.length; ++o ){ 
+        for( let o = 0; o < sort_orders.length; ++o ){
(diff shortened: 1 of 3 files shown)


JasonMcz:
**Comment:**
`The $10 STEEM sign up promo banner shows up even when my account is logged in everytime I refresh the page. Adding a timeout to this display? or check the browser local storage/cookie?`

app/components/pages/Post.jsx
```diff
@@ -78,6 +78,13 @@ class Post extends React.Component {
                         <PostFull post={post} global={g} />
                     </div>
                 </div>
+
+                //@Jasonmcz
+                //Comment: Perhaps it would be good to havd a render/timeout
```
noisy: > havd 

typo

svk31:
@JasonMcz Would you mind making a pull request with the same content but to the develop branch instead?

JasonMcz:
@svk31 yep, will do. I will do a pull req later.

svk31:
So I tested this and besides some minor syntax errors (missing `)` for the .map in cancelAllOrderClick and missing `this.` in the onClick for Cancel All), the main issue is it won't actually cancel all orders. The reason is that there's a confirmation dialog, so only the first call to cancelOrder is executed properly.

What we need in order for this to work is support for multiple operations in one transaction, that should be added soon though, maybe in a week or two. I'll leave this open till then and revisit once we have that ready.

roadscape:
What's with the `// Changed toFixex(8) because of rounding issue caused orderbook display?` comments that seem to be floating between unchanged lines of code?

And the "UI suggestion comment" is not relevant to this pull request, and doesn't belong in the code in general. It should be a separate github issue.

Why did you commit a whitespace change to `app/components/pages/PostsIndex.jsx` in this request?

Please keep the diffs *minimal and relevant.* Closing this issue, resubmit on develop after fixing.

PR #90, opened 2016-07-30
Enable soundcloud player

TigerND:


TigerND:
Testing mirror:

https://steemit.vlj.lt/steemit/@teego/enable-soundcloud-player

jcalfee:
Interesting, this works?  src.indexOf(reYoutube) .. I don't see this feature in the docs.

Do you have some soundcloud url docs so we can see what we are whitelisting?

svk31:
@jcalfee reYoutube is simply a variable containing the original whitelist url, he just moved it to the top together with the soundcloud and vimeo urls.

svk31:
I've just merged this manually in the `develop` branch, closing.

jcalfee:
I reverted this for now it has bugs it was not tested.  @roadscape was going to submit something he found.  Please fix and create another merge request and let us have more time to review.  This code touches user provided content so it needs to have extra testing .. We do this internally if we have to modify code like this.

roadscape:
We found some issues with this commit and ran out of time this week to fix...

`if (src.indexOf(reYoutube) === 0)` is not valid, it always returns -1, which would allow all annoying youtube parameters through. `if(re === reYoutube)` would fix this.

Also, in soundcloud embed I see there are multiple query params which need to be sanitized and whitelisted:

`<iframe ... src="https://w.soundcloud.com/player/?url=https%3A//api.soundcloud.com/tracks/205462171&amp;auto_play=false&amp;hide_related=false&amp;show_comments=true&amp;show_user=true&amp;show_reposts=false&amp;visual=true">`

We definitely want *auto_play* to be forced to false. I don't know about the other params but to start with we should force-set all params to reasonable defaults (except for 'url' of course). I'd recommend parsing out the url param and building the `src` in a strict manner. As stated above, this code concerns user-generated content so we need to be very careful.

svk31:
Yea I should've caught that, sorry. The indexOf won't work like that with a regex expression, it should be replaced with a regex test instead, or a srcYoutube string could be added.

PR #92, opened 2016-07-31
Pull Request: Adding Cancel All Button

JasonMcz:
**Goal:**

```
-an easier cancellation process for GUI traders on internal market 
-cancelling 10+ orders one by one can be difficult on the GUI level especially with the each reload the order sequence changes after the table gets  re-constructed 
Changed:
```

```
-looping through open_orders array 
-cancel each individual orders
-notify users about cancellation ( might be better just send one message after map is done?)
-reload the front-end
```

Note: Please take a look, I didn't have STEEMD on my environment so I wasn't able to test the function by running it locally.

**Comment:**

> The $10 STEEM sign up promo banner shows up even when my account is logged in everytime I refresh the page. Adding a timeout to this display? or check the browser local storage/cookie?


JasonMcz:
@svk31  is this good?

svk31:
Sorry for not getting back to you sooner. This has quite a few commits that aren't currently in develop branch that might cause issues, I would prefer a clean PR with only your changes. I can try to cherry-pick your commits, not sure that works from forks though.

JasonMcz:
That should be fine!

svk31:
Closing this since I figured out how to merge manually to whatever branch I want, sorry about the inconvenience. Will make more comments in the other PR.

JasonMcz:
Sounds good!

PR #104, opened 2016-08-06
Missing dependencies

TigerND:
The project cannot be installed to fresh linux systems due to missing dependencies.


TigerND:
Build details:
https://hub.docker.com/r/teego/steemit-bundle/builds/bahcn7vp4vtfhqtfygrvkvx/
Dockerfile and patches:
https://github.com/TigerND/docker-steemit-bundle

valzav:
Hmm.. strange it probably depends on linux distro and node version. We never had any issues with ubuntu 14.04 and 16.04 and node 6.3.

jcalfee:
Yes, is working on Ubuntu.  But we are definitely missing dependancies.  Look at immutable, we use that everywhere and it is not n our package.json..  Keep in mind we are using npm shrink-wrap so that would need to be updated too.

Someguy123:
@TigerND Your "system" is using Node 4.4.7

You need to use Node 6. I've found Steemit to be incompatible with the 4.4 LTS (for some reason)

roadscape:
Please resubmit to develop

PR #123, opened 2016-08-11
Added Following/Followed pages

aaroncox:
Links to the pages are in the user header. Should complete the bare
minimum requirements for steemit/steemit.com#112.

### DIFF
--- a/app/ResolveRoute.js
+++ b/app/ResolveRoute.js
@@ -49,7 +49,7 @@ export default function resolveRoute(path)
         return {page: 'SubmitPost'};
     }
     let match = path.match(/^\/(@[\w\.\d-]+)\/?$/) ||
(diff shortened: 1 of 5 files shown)


valzav:
Thank you, it's merged and most probably it will be in next release. Feel free to post a link to this pull request on steemit.com with some description of the work you've done so others could upvote your work.

jcalfee:
It is looks stable now.. At one point I had a short list of "followed" (maybe 10) but the Next and Previous buttons were lit up. I could like on them but the list would not change (nothing to see, it was a short list).  I can not re-create it now though.

aaroncox:
Appreciate the assistance! I'm really glad you caught the issue with the current user not being referenced properly, that would have caused some confusion for users.

PR #151, opened 2016-08-19
Show the additional witnesses you have voted for outside of the top 50

aaroncox:
After talking with a few people, there was some confusion as to whether
votes were actually being applied to people outside of the top 50. This
change adds those users below the input box, and also lets them upcast
their votes by clicking the button.

- The additional witnesses are loaded off the current user's `witness_votes` profile field. 
- Any witnesses in the top 50 are filtered out.
- Entering a username and voting will add the username to the list below.
- Clicking the blue vote will uncast the vote, removing the user.

image

### DIFF
--- a/app/components/pages/Witnesses.jsx
+++ b/app/components/pages/Witnesses.jsx
@@ -73,6 +73,29 @@ class Witnesses extends React.Component {
             )
         });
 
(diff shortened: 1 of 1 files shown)


roadscape:
Looks good, works well.

Issue #135, opened 2016-08-15
Dev environment for running steemit locally -- updates to README required


bmann:
I got steemit.com and steemd running locally on my Mac (the latter thanks to @TigerND and his homebrew items. The README is pretty good, I've made some changes and will do a pull request for OS X.

I'm up and running and looking at localhost:3002. How do I seed the database with a local dev login?

roadscape:
A local dev login for what? Currently there's no testnet, so there's no such thing "dev login" for steem or steemit.com. You'll just need to create an account for testing via the faucet or CLI.

bmann:
I have Steemit running in my Dev environment locally with a blank database. I want to do local development for visual fixes and pull requests. 

Can you point me at how to create an account for testing?

Happy to document.

roadscape:
There are 3 ways to create a steem account:

1) cli_wallet: `create_account string creator, string new_account_name, string json_meta, bool broadcast` command
2) Mining an account
3) Faucet: https://steemit.com/create_account

roadscape:
Were you able to create an account?

bmann:
@roadscape I'm not sure I'm asking for the right thing.

I have an account on steemit.com (http://steemit.com/@bmann) -- I can login, and know what all my passwords / keys are. I had that before, no problems there.

For local development, I have a copy of the steemit.com running at localhost:3002

I'm running steemd locally with ```steemd --rpc-endpoint 127.0.0.1:8090``` as listed in the README.

I can view localhost:3002 and see an "empty" instance running. I try and login with "bmann", and get "Username does not exist" (because it's connecting to the local steemd node, which isn't synced with anything).

So, does the local steemd need to be connected to different seed nodes so I can login using my existing account?

aaroncox:
@bmann Is local node running on testnet or something? I think most of us while doing development have nodes that are connected to the live network and all of the live accounts work fine for testing.

bmann:
You tell me @aaroncox :P I'm trying to follow the README successfully, and ```steemd --rpc-endpoint 127.0.0.1:8090``` is the only thing I have to go on. Do I need to also connect to seed nodes in configuration?

bmann:
OK, got some Twitter DMs from @TigerND 

In the README, we should be saying that you need to be running a full node locally. The command should be something like:

[...]

PR #156, opened 2016-08-20
Updates to README fleshing out OS X instructions

bmann:
Includes using @TigerND brew for steemd, addresses #135


bmann:
After discussion and help from #135, I've updated this to just tell people to use node.steem.ws.

PR #161, opened 2016-08-22
Added conditional header to posts for posts with depth > 0

aaroncox:
[@bitcoiner posted this feature request](https://steemit.com/tag/@bitcoiner/steemit-improvement-idea-give-a-link-to-the-parent-post) in response to some of the links I've been using on steemstats.com. These links leave steemstats.com and go directly to a reply, like so:

[https://steemit.com/tag/@jesta/re-bitcoiner-steemit-improvement-idea-give-a-link-to-the-parent-post-20160822t051520448z](https://steemit.com/tag/@jesta/re-bitcoiner-steemit-improvement-idea-give-a-link-to-the-parent-post-20160822t051520448z)

It's been incredibly useful to leave steemstats, immediately land on a reply, and be able to write a response. The only pain point is that there's no navigation on this page like you would get on reddit. So I propose the following change if `content.depth > 0`:

image

The two links will either:

- Bring you to the full context of the post.
- Load the direct parent of the reply you're reading.

This will likely also be useful in the future if steemit.com implements a notification system - as it will be easier to drop them on URLs like this than load the full post + scroll.

### DIFF
--- a/app/components/cards/PostFull.jsx
+++ b/app/components/cards/PostFull.jsx
@@ -195,6 +195,14 @@ export default class PostFull extends React.Component {
         let post_header = <h1 className="entry-title">{content.title}</h1>
         if(content.depth > 0) {
             let parent_link = `/@${content.parent_author}/${content.parent_permlink}`;
+            let direct_parent_link
+            if(content.depth > 1) {
+                direct_parent_link = <li>
+                    <Link to={parent_link}>
+                        View the direct parent
+                    </Link>
+                </li>
+            }
             if (content.category) parent_link = `/${content.category}${parent_link}`;
             post_header = <div className="callout">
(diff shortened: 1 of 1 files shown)


aaroncox:
**Question**: is there ever a situation where the post depth is greater than 0 and we would not want this header to be displayed?

roadscape:
I'd suggest not showing "View the direct parent" when its target is identical to "View the full context".

> is there ever a situation where the post depth is greater than 0 and we would not want this header to be displayed?

Nope

aaroncox:
> I'd suggest not showing "View the direct parent" when its target is identical to "View the full context".

Good call, just added that logic in.

roadscape:
Perfect, thank you!

PR #166, opened 2016-08-23
Webtorrent and docker-compose

faddat:
This PR adds 2 notable things:

1. A full docker-compose.yml that launches the front and back end.
2. the beginnings of [webtorrent](github.com/feross/webtorrent) support.  I may have borked the webtorrent support a little, so please let me know if/how I have.


valzav:
@faddat thanks for the time devoted to this pull request, but sorry, I think we are going to reject this pr at this point for several reasons:
1. We are already working on docker support and it's going to be in separate repo.
2. Adding webtorrent support has a cost: bigger code base, time for the a security audit, CSP headers change - I'm not sure there would be a lot of users for this technology to leverage this.
3. (minor) There should be no pull requests to master - master is only for production ready code.

PR #176, opened 2016-08-25
fixed direct parent link

aaroncox:
I believe my last commit to conditionally show and hide the direct link actually broke the direct link. This fixes it.


aaroncox:
Is there ever a situation where a comment isn't going to have a category?

roadscape:
Thanks, and not that I'm aware of.

Issue #203, opened 2016-08-30
implement reblog/repost feature


valzav:

faddat:
What's the functionality that this adds, exactly?

jcalfee:
Good question.. Can you make since of this?  This is how I figure it out..

https://github.com/steemit/steem/search?q=reblog

faddat:
aha, I've got it.

It's twitter retweets :).  

Thank you for the link!  I do remember something called reblogs but seriously we're talking mid-late 1990's intarwebs.....

roadscape:
todo:

 - styling: make it blend better
 - users ask for ability to undo Reblog
 - remove from PostSummary? (otherwise, add to mobile view)

faddat:
@roadscape 

Just throwing in my $.02 here:  I'm not so certain about this feature, overall, because it doesn't seem to fit the format.  Possibly needs tweakage, possibly just me being strange.

PR #254, opened 2016-09-15
Reblogging functionality #203

roadscape:
Working well, waiting for backend.

This should not be merged until https://github.com/steemit/steem/issues/401 is resolved and merged into develop.

### DIFF
--- a/app/assets/icons/reblog.svg
+++ b/app/assets/icons/reblog.svg
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 16.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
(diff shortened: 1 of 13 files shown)


jcalfee:
Can user's reblog twice?  Either way, we should probably know if a post was recently re-blogged.

roadscape:
Users "can" reblog twice, and I'm not sure how the backend disposes of duplicate reblogs. I don't believe it's intended to be able to reblog multiple times.

Afaik, there's no way to know (at least from any steemd API) if a post was recently re-blogged.

In the near future we may have a get_state2 call which will take the current user context into account and return more useful info. Until then, if it's important to save the reblog state we may have to do it in local storage?

jcalfee:
`localStorage.getItem("reblogged") could be localStorage.getItem("reblogged_" + username)`

roadscape:
I added namespacing.. it would be a lot smoother if we had currently logged in user in localStorage and didn't have to wait for the login event. Is there anything preventing us from doing this?

roadscape:
Once this PR gets merged we can test & deploy: https://github.com/steemit/steem/pull/428

jcalfee:
>I added namespacing.. it would be a lot smoother if we had currently logged in user in localStorage and didn't have to wait for the login event. Is there anything preventing us from doing this?

I believe it is in `state.offchain.get('user')` .. Notice, when you go to steemit.com it forwards to your feed before you are logged in.  In incognito you instead see a banner and the mainpage.  Look at App.jsx 204.  @valzav is this the way to do it?

jcalfee:
state.offchain comes out of the mysql database.. So it seems the session is left open and this is loaded on first render.

valzav:
you can use state.offchain.account - this is user's account name stored in cookie

roadscape:
As of this merge, **steemd must be built on or after https://github.com/steemit/steem/commit/354cd6bd353b3b6359d28f878cbb3e375173b661**.

PR #258, opened 2016-09-15
257 use webpack extract text plugin in dev

valzav:
see https://github.com/steemit/steemit.com/issues/257


valzav:
got more files changed than needed. closing

PR #287, opened 2016-09-20
No errors if HtmlReady is given a blank string

keverw:
A small little edit so no errors are made if given a blank string. I was playing with this module directly for a project I'm working on and noticed that empty string caused errors because `doc` would be undefined when calling `DOMParser.parseFromString` with a blank string.

So this checks if doc is defined before calling `serializeToString` and within `proxifyImages` before the forEach loop.


roadscape:
Looks good, thanks.

Issue #311, opened 2016-09-21
Frozen Discussions Shouldn't show an input box


faddat:
Current user workflow is like this:

Reads article, 
scrolls down and sees:
![image](https://cloud.githubusercontent.com/assets/7142025/18728500/92f37ab0-8077-11e6-988c-e392558bec81.png)

Types comment:

![image](https://cloud.githubusercontent.com/assets/7142025/18728560/df117bea-8077-11e6-97cb-4170e19a3ed1.png)

Clicks "post":

![image](https://cloud.githubusercontent.com/assets/7142025/18728583/f9fbd0a4-8077-11e6-92a7-89e44f4c1ca0.png)

Oh, noes!  It is frozen!  

Better workflow:

![image](https://cloud.githubusercontent.com/assets/7142025/18728640/5941233e-8078-11e6-9f30-e0994ce223c7.png)

roadscape:
Thanks, fixed in https://github.com/steemit/steemit.com/pull/321

PR #321, opened 2016-09-22
Remove actions from frozen discussions #311

roadscape:


jcalfee:
In Comment.jsx, looks like your missing `&& depth < 6` now..

roadscape:
Great catch, thanks!

roadscape:
On closer look, actually I see it's in place: https://github.com/steemit/steemit.com/pull/321/files#diff-e49ee0fecdc197dfe80ad1c25050bcceR263

PR #353, opened 2016-09-28
This refactor allows for post list items to have hyperlinked categories based on sort order(Improved Navigation)

mcifani:


roadscape:
There may be a simpler way to do this using TagList component, please take a look:
https://github.com/steemit/steemit.com/blob/develop/app/components/elements/TagList.jsx#L9

So instead of `<strong>{post.category}</strong>` you could use `<TagList post={post} single />`

mcifani:
@roadscape Good call, updated

PR #369, opened 2016-09-30
add newrelic support; rebuild shrinkwrap; update react to 15.3.2; remove browser-sync

valzav:


sneak:
From newrelic: Some modules were uninstrumented during the current time window: generic-pool, mysql, bluebird, connect. Make sure require('newrelic'); is the first line of the application's main module.

Is something loading those modules before the require? I thought you put it at the top, I'm not sure why it's saying that...

PR #416, opened 2016-10-05
Fix various page titles

thebitcoiner:
Similar to https://steemit.com/steemit-bug/@ash/steemit-bug-fix-feed-ttitle


roadscape:
Thanks, looks good.

PR #418, opened 2016-10-05
Markup numbered lists with more than 9 items restart at "0"

thebitcoiner:
From @TimCliff's https://steemit.com/steemit-ideas/@timcliff/the-steemit-wish-list-avatars-notifications-multi-language-support-and-more-oh-my-v2-0

image

The bug was showing as

image

This was caused by a default normalize.css rule for the left margin on ol and ul tags.

Corrected:

image


roadscape:
Thank you for reporting this, I will make this change in `app/assets/stylesheets/markdown.scss` as it will be a more targeted rule and I'm working with that file atm.

Issue #262, opened 2016-09-16
Decline payout reward option


roadscape:
Certain posts (e.g. platform announcements) do not need to collect a payout.

We should have an optional checkbox when creating a post to "decline rewards". It would set the `max_accepted_payout` to 0.000 SBD.

Posts which are not receiving a payout still collect rshares and get ranked on the trending page.

randomSteemit:
Another alternative to this would be something like, If you decline payment it could go into a temporary account for the day and then at the end of the day it would be split between all active users for the day. This would give incentive to be active every day.

faddat:
@randomSteemit I think this would need to be worked out on both the back end and the front end, but I think it's got merit.  DAU matters overall as a metric, and besides, upvoting is our content filtering system-- if activity stops, the site breaks.  

I like this concept.

randomSteemit:
@faddat of course, and very true that activity is one of the most important metrics. The best way to keep that metric up is to give the users something to keep them coming back every day. The content on steem is good, the interest on SP and SBD is great, why not throw in another incentive to remain active everyday and give people the chance to "Give Back" to the community by spreading their wealth amongst the active ones trying to grow the platform.

faddat:
Alright people, I think that Steemit, Inc. and the steem community deserve a pat on the back for being the spark for such discussions.  I just had one of those moments of clarity where I... "wait this is like inverse Facebook!  It's a social network that intentionally enriches and empowers its users!" 

Totally OT.  Felt it was worthy of mention anyway.

theoreticalbts:
I think [this post](https://steemit.com/spam/@test-safari/m8vdx-test) is not right.  `max_accepted_payouts` is not a field in `comment_operation`, rather it is in a separate `comment_options_operation` in the tx.

roadscape:
@theoreticalbts It does need more testing, but it appears to be reflected as expected:
![image](https://cloud.githubusercontent.com/assets/5168676/18885899/9ef95aa8-84bb-11e6-9c27-8799995095a6.png)
https://steemd.com/spam/@test-safari/m8vdx-test

roadscape:
I have confirmed it's using `comment_options_operation` to set this value

theoreticalbts:
Why doesn't it show up in the block?

I do `get_block 5341475` and see:

    {
      "ref_block_num": 33041,
      "ref_block_prefix": 1379079997,
      "expiration": "2016-09-27T17:34:21",
      "operations": [[
          "comment",{
[...]

PR #420, opened 2016-10-05
Update payout options #262

roadscape:


jcalfee:
Are you really sure about putting defaultPayoutType in local storage?  I think it should be a decision on every post.  This affects the author's income.

jcalfee:
At the very least, make sure `decline payout` is not stored in local storage.  If a user wants to set to 100% SP as a payout that would not be so bad to save.  You could verify by giving them a "remember" link in the form.  If the setting gets changed on a later post after "remember" it could auto-forget and offer to remember again.  But I would not want to save `decline payout`..

roadscape:
How this PR addresses your concerns:

1. The default option is 50%/50%
2. It's a decision on every post
3. When 'Decline Payout' is selected, the color of the text turns Orange to draw attention to it. I don't expect this option will be used often, but when it is selected, it should be clear to the user.

jcalfee:
It looks good.. I think that if you get "decline payout" that is should not save in localstorage.  So the next post will default to 50/50 or 100 depending on the post before it..   or just remove local storage from this drop down.

jcalfee:
If val thinks it is ok he can approve it as is (I don't mind) .. ;)

jcalfee:
The big reason is: this option can't be reversed.  If you accidentally put your hard work in under decline and someone votes you can't remove the post, or change that option.  Do you have users that want to consistently decline payouts?

roadscape:
I suspect users who use the 100% option would prefer the option to be sticky. The concern was avoiding unexpected/inconsistent behavior -- e.g. sometimes the option saves between forms, sometimes it doesn't. I understand that this could be not ideal to users who want to decline payout once, but I expect such cases to be rare overall, and this is why a colored label was added.

The main use case is platform and witness updates -- posts which don't warrant a payout yet deserve to be voted on & still ranked for visibility.

roadscape:
Due to a bug in the backend, posts which do not receive a payout still count against a user's optimal max of 4 posts per day. So it's (currently) not all that useful for users who want to share frequently without any reward while still being eligible for the occasional full reward post.

roadscape:
Also, there is one missing piece of functionality -- a signal that a post has reached its maximum payout. I will update Voting.jsx to check for this case so the UI reflects it.

roadscape:
This PR will conflict with #422. #422 is higher priority.

roadscape:
Since 'Decline Payout' will not be used too often, and it still counts against the optimal 4 posts per day, I concur that simply not saving it is the best way to go for now. I've made that change in my last commit.

I also fixed a bug where comments use the last setting (even though they don't get a select box).

More issues to consider for the future:
 - Payout type should be saved as part of the draft status -- returning to a draft will clear "Decline payment" currently.
 - The split between 100% and 50-50 maybe should probably be *account-level*, and apply to comments as well. "Decline payment" could be a per-post setting.
 - For now it's forcing all comments to 50-50. Might be better to just go 100% SP for all comments.

jcalfee:
Just had an idea..  I think removing decline payout from local storage is still fine.  An alternative was to use the __confirm feature in the transaction if the user were about to decline payout.

roadscape:
> Just had an idea.. I think removing decline payout from local storage is still fine. An alternative was to use the __confirm feature in the transaction if the user were about to decline payout.

I like this idea.. for the first iteration I think either approach is good. A confirm dialog could contain a brief explanation of this feature, but I've also stubbed out a posting help dialog in branch `153-posting-guide` and I envision it explaining all the payment options.

I have rebased this branch with develop now that the ReplyEditor refactor has been merged in, so it's good to go unless you'd like to add __confirm logic!

Issue #185, opened 2016-08-25
Medium-like Editor Integration


jcalfee:
This [branch](https://github.com/steemit/steemit.com/tree/medium-editor) replaces the Rich Text editor with the [yabwe/medium-editor](https://github.com/yabwe/medium-editor) ..

Packages used:

https://www.npmjs.com/package/react-medium-editor
https://www.npmjs.com/package/medium-editor-insert-plugin

JasonMcz:
When is this gonna get merged to the master branch and get deployed?

jcalfee:
Something to consider: https://app.classeur.io provides a real-time view of markdown and styling together.  Also interesting how they can support uploading of images..

sneak:
![](https://github.trello.services/images/mini-trello-icon.png) [Editor](https://trello.com/c/hF9VD7vj/35-editor)

sneak:
lol editor

PR #422, opened 2016-10-05
Refactor of ReplyEditor #185

roadscape:
I've been testing this extensively, but I'd like to get it merged into develop soon so any issues can be caught quickly.

Main changes:
 - `<html>` is no longer used during the editing process. It's stripped in `componentWillMount` and the `rte` flag is set appropriately. It's only added back during `reply()`.
 - No longer updating `rte_value` in `componentWillReceiveProps`. I believe this was a workaround and it should no longer be required.

I removed the `if(editorData.formId === formId)` check in [ReplyEditor.jsx#L116](  https://github.com/steemit/steemit.com/blob/e9574285d6761f330e4e1e801fc4ed2e4839f359/app/components/elements/ReplyEditor.jsx#L116) as it seems it could never be false.. let me know if this is not the case.

### DIFF
--- a/app/components/cards/Comment.jsx
+++ b/app/components/cards/Comment.jsx
@@ -246,7 +246,7 @@ class CommentImpl extends React.Component {
         // const steem_supply = this.props.global.getIn(['props','current_supply']);
 
         const showDeleteOption = username === author && !hasReplies && netVoteSign <= 0
-        const showEditOption = username === author && comment.mode == 'first_payout'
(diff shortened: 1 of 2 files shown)


roadscape:
It looks like one of the words in your comment disappeared, if you're referring to posts with `<html>` tags then yes.

I made one more change -- drafts are now saved with the `<html>` wrapper, this makes them compatible with current drafts and simplifies `componentWillMount` logic.

PR #426, opened 2016-10-06
Show Resteemed on Blog

thebitcoiner:
This will allow users to go to a blog and be able to easily tell which posts are from the author and which were resteemed.

The pattern from the feed was followed.

Here is an example of what it is currently:

image

Here is what it will look like:

image

### DIFF
--- a/app/components/cards/PostSummary.jsx
+++ b/app/components/cards/PostSummary.jsx
@@ -52,6 +52,7 @@ export default class PostSummary extends React.Component {
     render() {
         const {currentCategory, thumbSize, ignore, onClick} = this.props;
         const {post, content, pending_payout, total_payout} = this.props;
+        const {account} = this.props;
         if (!content) return null;
 
         const archived = content.get('mode') === 'archived'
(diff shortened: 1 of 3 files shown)


roadscape:
This may be better solved by having the backend return `first_reblogged_by` for user blog feeds, as this would not require any workarounds on the front end

roadscape:
But if we want to go this route for now, I'd recommend changing

```
 +        if(account && account != content.get('author')) {
 +          reblogged_by = <div className="PostSummary__reblogged_by">
 +                             <Icon name="reblog" /> Resteemed
 +                         </div>
 +        }
 +
```

to

```
          let reblogged_by = content.get('first_reblogged_by')
 +        if(!reblogged_by && account && account != content.get('author')) {
 +            reblogged_by = account // workaround #426
 +        }
          if(reblogged_by) {
```

to leverage the existing html generation logic

thebitcoiner:
Yes, having a way to get first_reblogged_by from backend would be better and will allow for the next step, showing the Resteemed on another tab.

I agree the workaround makes for cleaner code, but getting feedback from others on chat, most have said they don't want to see the " by author" text, just Resteemed, on the blog.

thebitcoiner:
Further feedback? Can we leave as is?

PR #427, opened 2016-10-06
Fix Witness Vote Login Prompting for Posting Key

thebitcoiner:
This is in response to a dialog in Steemit Chat #help channel.

image

The issue arised from the check on the authType: _vote_ was being matched in _account_witness_vote_.

Corrected:

image


roadscape:
Looks good, merging

Issue #233, opened 2016-09-11
Add i18n support


f3rnando:
I'm creating this issue in order to begin discussions on i18n support for steemit.com given that Im working into a clean, updated fork of this repo to include this feature.

Core points, questions and insights:
1. Adding i18n support is quite easy, the biggest efforts reside on translation and on keeping translation files updated.
2. I will PR as soon as spanish translations are done, in order to make merging relevant and useful right away.
3. The modules/plugins that I chose are: [react-intl](https://github.com/yahoo/react-intl) and [babel-plugin-react-intl](https://github.com/yahoo/babel-plugin-react-intl). Are these fine? if any other should be considered, please propose it ASAP.
4. The current i18n branch can be found here: [f3rnando/steemit.com/tree/i18n](https://github.com/f3rnando/steemit.com/tree/i18n) 
5. The current status compare of steemit.com:master > f3rnando:i18n [here](https://github.com/steemit/steemit.com/compare/master...f3rnando:i18n?expand=1)
6. Language selection should be automatic based on user locale
7. A chinese speaking user might access from a browser asking en-US, and he might want to override that language selection. I will not develop this UI feature, I believe it is subjective and the steemit team will take over this task.
8. I may give access to this repo to whoever wishes to collaborate and/or audit it.

Please reply referring to the list number, if any.

Updates coming soon.

f3rnando:
If this proposal is given green light, I will rename the branch to reflect this issue's number.

mynameiskeo:
Nice initiative... Was trying to work on this for a while, and bring developers to it. Heimindanger started something with me on http://94.23.30.172:3002/  (FR), and i'm finally quite BUSY on another project :) 
Very glad you work on it

https://github.com/yahoo/react-intl looks Ok
you can also have a look to https://www.npmjs.com/package/i18n-react

skzap:
Hi @f3rnando, we want to make a French website for steem with @mynameiskeo 
I would be able to help you for porting the code.

1. Agreed, there will be a lot of edits but it should be pretty straightforward.
6. We could also store the language preference of a user in the json_metadata, but I agree by default use the locale.
7. Different language clones actually serve a bigger purpose than just translating. They also promote certain tags (#ru , #cn , etc). If we have a UI to change language directly inside steemit.com I'm afraid these communities might die.
8. Please do, also can I contact you on steemit.chat? I am @heimindanger there

mynameiskeo:
[...]

PR #430, opened 2016-10-06
add i18n support

valzav:
see #233 for details


roadscape:
IMO, translations need namespacing otherwise they will get out of control.. here's examples of what I consider developer/translator-friendly structure:

https://github.com/plataformatec/devise/blob/master/config/locales/en.yml
https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml

It would be better to do this sooner than later

sneak:
i agree, but it can be a different PR

sneak:
there are people who want to begin contributing translations and this allows them to do that. it's easy enough to refactor the translated strings later in a single move.

PR #477, opened 2016-10-13
Fix for External Witness URLs to Open in New

thebitcoiner:
Some witnesses set their url with an external links. This will allow users to view them from the Witnesses page (instead of directing to steemit.com homepage due to it not found).

In addition, added a space between the two sentences:

image

Also, added `target=_blank` to another page's link that had showed the extlink icon.

### DIFF
--- a/app/components/modules/BlocktradesDeposit.jsx
+++ b/app/components/modules/BlocktradesDeposit.jsx
@@ -222,7 +222,7 @@ export default class BlocktradesDeposit extends React.Component {
                 <div className="row">
                     <div className="column small-12">
                         {sendTo}
-                        {paymentLink && !hasError && <a href={paymentLink}>&nbsp;<Icon name="extlink" /></a>}
+                        {paymentLink && !hasError && <a href={paymentLink} target="_blank">&nbsp;<Icon name="extlink" /></a>}
                         <div className="de-empasize">{depositLimit && `Suggested limit ${depositLimit}`}&nbsp;</div>
                     </div>
                 </div>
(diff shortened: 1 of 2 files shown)


roadscape:
There is a potential vulnerability with using target=_blank, so for now it's best to avoid using them. The other changes look good.

thebitcoiner:
It would be a vulnerability to the opened and not the opener? Would this be fine? There are links already that do this like opening steemit.chat in new from the sidebar.

roadscape:
In some cases/browsers, the new tab has access to its opener. This is mainly an issue with user-generated links, less so with hardcoded links to trusted sites (though we should at some point make sure we are handling these cases consistently).

thebitcoiner:
If rel="noopener noreferrer" was added, can it be accepted?

roadscape:
There are still browser support issues with this feature (namely Safari), so current practice is to leave out target=_blank for user-generated links. We may implement a more advanced workaround in the future.

Details: https://mathiasbynens.github.io/rel-noopener/

roadscape:
Regarding the  `target="_blank"` in BlocktradesDeposit.jsx .. it seems this is for a bitcoin protocol URL, so I'm not sure it makes sense to open in a new window if it's going to be launching an external app

thebitcoiner:
Oh, I didn't notice it was a `bitcoin:` link, in that case, the browser should handle it and the target="_blank" should be removed. Remove?

roadscape:
Yes, remove please

Issue #511, opened 2016-10-20
resteemed icon falls out of metadata portion on right-hand-side


sneak:
<img width="985" alt="screen shot 2016-10-20 at 00 03 10" src="https://cloud.githubusercontent.com/assets/408977/19545909/7e7c7b8a-9658-11e6-856f-6f1bd2f6f38d.png">

The little reverse/repeat arrow icon is falling too far left, intruding on the thumbnail part.  The icon and the label text for the icon should be contained in the right-hand side; that is, left-aligned to the same left edge as the the title, subhead, and upvote button.

mcifani:
![](https://github.trello.services/images/mini-trello-icon.png) [resteemed icon falls out of metadata portion on right-hand-side #511](https://trello.com/c/xW9clwOn/83-resteemed-icon-falls-out-of-metadata-portion-on-right-hand-side-511)

PR #522, opened 2016-10-24
resteemed icon falls out of metadata portion on right-hand-side #511

mcifani:
- adjust alignment for resteem icon to match post snippet


roadscape:
IMO, this line of text is out of place but barring any significant design changes it still looks better when the gap between icon/text is aligned with the gap between thumb/text below. 

It didn't start out this way though, it was added for non-mobile view in https://github.com/steemit/steemit.com/commit/46b66f72e87410112d3da986c58a2d7c07ecc1ab
If we're going to do away with it, the entire rule block can be removed.

PR #569, opened 2016-11-01
536 on a mobile wallet malform

mcifani:


mcifani:
screen shot 2016-11-01 at 6 09 55 pm

Fixes malformed drop down on wallet balance options(mobile/tablets)

PR #570, opened 2016-11-02
remove ReplyEditor__title padding

thewisenerd:
this was brought in by 647bcfe8 and breaks the title input
by hiding text (in firefox 49.0.2).

surprisingly does not break things in chrome/opera, making
this a browser quirk. also, padding on input elements should
not really be necessary, i guess. browsers will take care of
things on its own.

### DIFF
--- a/app/components/elements/ReplyEditor.scss
+++ b/app/components/elements/ReplyEditor.scss
@@ -17,7 +17,6 @@
 .ReplyEditor__title {
   font-size: 1rem;
   font-weight: bold;
-  padding: 1rem 0.5rem;
 }
 
 .ReplyEditor__draft {


thewisenerd:
screenshot, firefox 49.0.1.

thewisenerd:
@roadscape, please have a look at this (fixes a bug introduced in 647bcfe8).

roadscape:
The goal with padding was to get this input field closer to resembling an `<h1>` and give it some space. I addressed this bug in https://github.com/steemit/steemit.com/commit/014cb18d5ec390bc518b973e76c2f0a850126ab7 but missed the fact it was introduced on develop.

Your fix will work until we merge that branch, thanks!

PR #571, opened 2016-11-02
No post results ui bug

bigbhowell:
**Added feedback text that will display when there are no results while browsing posts.**

To see the UI bug in action, [check out this link](https://steemit.com/promoted/i-am-a-tag-that-will-not-be-found). As you can see there are no post results and the viewable posts area is left blank. This scenario occurs often when browsing the *promoted topic* for a *specific tag*, but it is not limited to this case.

This commit updates the empty text when there are no posts so that the user is not left wondering if posts are loading or simply not available.


roadscape:
lgtm.. thanks!

Issue #404, opened 2016-10-04
Create a "Welcome to Steemit" page


TimCliff:
There should be an informative landing page for new users when they sign up

sneak:
Want to write one and send in a Pull Request? :)

TimCliff:
Sure. I'll probably solicit quite a bit of help, but I'll take it on :)

jcalfee:
Great, call me out in github (@jcalfee) if you have any questions..

TimCliff:
Thanks!

TimCliff:
If anyone is interested in contributing content for the welcome page, I created a post to gather input from the community:
https://steemit.com/help/@timcliff/help-us-create-the-official-welcome-to-steemit-landing-page-community-input-needed-part-1

TimCliff:
Hi @sneak and @jcalfee - We have the content for the page pretty much ready. What is the process as far as having someone from Steemit review it? Should someone take a look at it before we start to code it into the page, or would the best way be to just submit a pull request with our first draft? I've asked Ned to take a look at it too, but I don't know if he will have time.

The content is here:
https://steemit.com/help/@timcliff/an-onboarding-solution-in-the-works-creating-the-official-welcome-to-steemit-landing-page-community-draft-content-review

sneak:
Send a PR with a WIP title with just a draft... we can pull it into a branch. Does github let you make a PR to a new in-repo branch? That would be best.

TimCliff:
I don't think we are able to create a new branch. Is there one that already exists that we could use?

Also, can we submit a text file or word document with the content, or would it be better to convert it into code first? (We will have an actual page / code file for the final pull request, but right now we are just looking for content review.)

sneak:
branch created: https://github.com/steemit/steemit.com/tree/404-welcome-page

Ultimately you're going to have to drive this - don't use word, but if you actually want to *ship* this, it's going to have to be in the same format as the existing pages.  You could do a draft in plain text or markdown but you're going to need to integrate it, too, so you might as well put the draft into the right format for integration.

TimCliff:
Sounds good, thanks!

roadscape:
Added in #585

PR #585, opened 2016-11-04
Add welcome page and link in hamburger and empty blog

thebitcoiner:
For issue #404

Added welcome page that new users redirect to.

A link has been added in the hamburger and in an empty blog.

Related post: https://steemit.com/help/@timcliff/the-official-welcome-to-steemit-landing-page-has-been-submitted-to-github-as-a-pull-request

### DIFF
--- a/app/ResolveRoute.js
+++ b/app/ResolveRoute.js
@@ -6,7 +6,7 @@ export default function resolveRoute(path)
     if (path === '/about.html') {
         return {page: 'About'};
     }
-    if (path === '/welcome.html') {
+    if (path === '/welcome') {
         return {page: 'Welcome'};
     }
     if (path === '/faq.html') {
--- a/app/ResolveRoute.js
+++ b/app/ResolveRoute.js
@@ -6,6 +6,9 @@ export default function resolveRoute(path)
     if (path === '/about.html') {
         return {page: 'About'};
     }
+    if (path === '/welcome.html') {
+        return {page: 'Welcome'};
+    }
     if (path === '/faq.html') {
         return {page: 'Faq'};
     }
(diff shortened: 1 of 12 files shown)


mcifani:
Looks good.  One suggestion is we try to get away from .html routes/pages.  Seems a bit cleaner with just "welcome".

valzav:
ok, I will merge it in after today's release

thebitcoiner:
@roadscape The conditional disable of the sanitize filter is done only for the HelpContent, which is the FAQ and welcome pages. As these are from markdown files in the repo and not from a user, it should be safe. The redirect to the welcome page happens on the login page only, which the user gets to after a sign up. If a user logs in through the login modal, there will be no redirect.

@mcifani Agreed. I was following the current naming scheme for other pages on the site. Refactoring these in the future would be a good idea.

@valzav Thank you.

roadscape:
Requested changes:
 - Remove references to SvgImage
 - Rename 'welcome.html' to 'welcome' (we are moving away from this legacy convention)

thebitcoiner:
The requested changes have been implemented.

thebitcoiner:
I believe this is ready to merge and not WIP.

roadscape:
@thebitcoiner - tested and works well! My concern about the new welcome page is that it's a bit monochromatic/dry compared to the current new user landing (but certainly more helpful). I hope you might find time to spruce it up a bit with images and maybe some "quick links" at the top to jump straight into good content. Thanks!

---

image

---

image

thebitcoiner:
Thank you @roadscape! That is a good point. After a new user signs up, they will be brought to this new page instead of their blog.

PR #599, opened 2016-11-07
translations & account_update

roadscape:


roadscape:
do we need the new files in `app/assets/images/favicons/`?

todo

- [ ] app/components/pages/Market.jsx (several issues spotted, incl fixes reverted)
- [ ] app/components/elements/DepthChart.jsx

Issue #624, opened 2016-11-14
Joined date on mobile


Tadasu85:
Joined field does not display correctly on mobile device. Android chrome.

roadscape:
Fixed by @thebitcoiner with #629

PR #629, opened 2016-11-15
Fix joined date display

thebitcoiner:
For issue #624

With a small browser width like those on mobile, the Joined Date is displayed as follows:

image
After fix:

image

A CSS rule from before the user pics were implemented needed to be removed. In the past, this was to shrink the banner div as the h2 was responsively shrunk too.


mcifani:
Looks good.

PR #633, opened 2016-11-16
Fix various UI glitches on mobile

thebitcoiner:
**Currently, users on mobile will see:**

image

image

***

**With these changes:**

image

image


jcalfee:
Looks good .. Wider version is unchanged:

image

PR #634, opened 2016-11-16
Thebitcoiner 616 resteemed tab

jcalfee:


roadscape:
In some cases this seems to break pagination. This is possibly due to pages not being scrollable when the first 20 items only contain a few blog posts, resulting in not being able to trigger `loadMore`.

Here's one account that exhibits this behavior:
http://localhost:3002/@thecryptofiend

By going to the 'Resteemed' tab and getting more pages to load, then going back to 'Blog', pagination works again.

Also, splitting the posts in this way creates cases when new pages load a random number of items (as opposed to a predictable 20 with every successive request) which is not ideal ux.

Aside from these 2 points, the feature works quite well. We could give users the option of combining their blogs/resteems through account settings (json_metadata). But I suspect we will need to have 2 separate APIs to implement this feature without more workarounds.

thebitcoiner:
This is an interesting case. Maybe it would be ideal to have an optional argument like `include_resteemed `to the API call `get_discussions_by_blog`, or separate calls like `get_discussions_by_blog_with_resteemed` and `get_discussions_by_blog_without_resteemed`. Can someone help me do this or point me in the right direction?

jcalfee:
You can try to make a request here https://github.com/steemit/steem and see if they will add the extra parameter.

thebitcoiner:
Thanks, @jcalfee. I have submitted an issue there at https://github.com/steemit/steem/issues/581

valzav:
Rather than adding Resteemed tab and separate drop down menu entry I would add `Show Resteemed` checkbox to Blog tab.

jcalfee:
@valzav Should we figure out how to fix paging first?  Seems like the checkbox or a similar layout enhancer could be done later so that would not have to hold up merging.  In any case, I'm interested in the paging fix first as that is the show stopper.

valzav:
@jcalfee yes, pagination needs to be fixed but this is unrelated to this PR. 
Regarding this PR we need to decide whether it should be implemented as separate Resteemed tab or as "Show Resteemed" checkbox.

thebitcoiner:
There is good and bad about either option. With the API change, pagination will be fixed as loading more resteems or blog posts will give a consistent number of posts back. But one potentially negative thing is that users will not be able to see chronologically when the resteems were done by the author in the midst of author posts.

With the Show Resteemed checkbox, there could also be a pagination issue. If the viewing user has chosen to hide the resteemed posts, and the next load request were, say, all 20 resteems, then the viewer will not see anything new loaded until they trigger another load.

Also, there may need to be some clarification, the "Show Resteemed" checkbox suggested by @valzav only affects what the visitor sees and not what others say. The suggestion from @roadscape about "the option of combining their blogs/resteems through account settings (json_metadata)" would affect what other visitors see. Is this correct?

sneak:
@jcalfee I'm assigning this to you - please figure out if we want to proceed with this and either close the PR or merge it by 1/1.

thebitcoiner:
Please close. Continued with PR #819

Issue #614, opened 2016-11-11
UX related: move some links out from the side menu to top menu


abitmore:
Something like "Steem App Tools", docs and perhaps whitepaper need more eye balls.
I'm not sure how many people even know there is a side menu, since it will only appear when clicking on the icon in the right top corner.
In the meanwhile we have nothing at the top.

Tadasu85:
The whitepaper certainly needs more eyes.

abitmore:
Actually "Steem App Tools" links to http://steemtools.com/, but that site is uncategorized, new users would get lost when be directed there (by the way it should be opened in a new window/tab).

What we need are links to "official" or "certified" or "featured" mobile apps, browser extensions, that can make users easier to use Steem the blockchain (by "use" I mean basic functions e.g. read posts, reply, vote, submit posts, transfer Steem Dollars) then perhaps "advanced tools" at somewhere deeper.

Many items on the right menu aren't noobs-friendly, or are outdated. Only piles of functions, lack of explanations. 
* About? Most contents on the first page are outdated.
* Explorer? For what? A list of tags. Why not just show the numbers on the main page?
* WhitePaper is something. But content is outdated.
* FAQ is something. Lack of an index. I'm not sure whether an average Joe will ask these questions, or understand the answers.
* Buy Steem? Why? Also the page has too little info, although it works. IMO we need at least a fat disclaimer or agreements to be accepted.
* Steemit App Center is something, issues mentioned above.
* Market? A place for shopping with Steem Dollar? No it's not..
* Password related links are easy to be understood. OK. Chat? OK. API docs? OK
* Witnesses? No explanation. 
* Privacy Policy? OK. TOS? OK.

What we need to be easily accessed (show them directly on the main page but not hidden somewhere):
* links to "official" mobile apps
* a link to official blog / news release / press
* a link to official career opportunities

By the way, the google custom search page doesn't look well. Can we at least get rid of the ads (usually one full screen)? Just try search "steem" or "steemit" or "art" or anything else.

abitmore:
More related info here https://steemit.com/steemit/@ats-david/steemit-we-have-a-problem

mcifani:
Some suggestions have been implemented in #635 -  Looking into disabling the ads(agree they are distracting).

mcifani:
Thanks abitmore for your suggestions. Closing ticket for brevity along with many suggestions are in our triage/future considerations or have been updated.

PR #635, opened 2016-11-16
Hammer menu changes

thebitcoiner:
**Relating to issue #614** 

* Steemit App Center to open in new - roelandp can be trusted to not mess with the opener
* Added strings for "Currency Market" instead of "Market" for clarity
* Steemit API Docs to have the extlink icon
* Added strings for "Vote for Witnesses" instead of "Witnesses" for clarity


roadscape:
Looks good, thanks!

PR #660, opened 2016-11-17
650 submit story trx op order

jcalfee:


roadscape:
`shared/HtmlReady.js` is extraneous here but it seems to be identical to what was already merged into develop

PR #661, opened 2016-11-17
Ensure lowercase on user name formatting for page title profile routes

mcifani:


app/components/modules/Header.jsx
```diff
@@ -113,7 +113,7 @@ class Header extends React.Component {
             page_title = `Stolen Account Recovery`;
         } else if (route.page === 'UserProfile') {
             user_name = route.params[0].slice(1);
-            page_title = user_name;
+            page_title = user_name.toLowerCase();
```
roadscape: Is this change necessary? I believe username will always be lowercase in the route

app/components/modules/Header.jsx
```diff
@@ -137,8 +137,13 @@ class Header extends React.Component {
             page_name = ''; //page_title = route.page.replace( /([a-z])([A-Z])/g, '$1 $2' ).toLowerCase();
         }
 
-        // Always format first letter of all titles capitalized for consistency & readability
-        page_title = page_title.charAt(0).toUpperCase() + page_title.slice(1);
+        // Format first letter of all titles and lowercase user name
+        if (route.page === 'UserProfile') {
+            page_title = page_title.charAt(0) + page_title.slice(1);
```
roadscape: This line can be removed, we only need to handle `route.page !== 'UserProfile'`

app/components/modules/Header.jsx
mcifani: Yep, correct the username always is lowercase so it's not needed.

PR #692, opened 2016-11-22
625 fix account creation issue that could potentially allow attacker to create multiple accounts per single verification

valzav:


jcalfee:
I would like to move unlock to a finally block.  In the process I'm testing the code.

jcalfee:
@valzav please review my commit and merge if that works for you...

PR #709, opened 2016-11-23
Display previous title when closing post modal, keep previous title

mcifani:


roadscape:
Made a small change so that modal closing logic happens in just the 1 function

PR #713, opened 2016-11-23
Fixed show more low value posts text (low was hidden)

jcalfee:


roadscape:
@jcalfee, what happened here, did we make the same change? The only difference I see here is how the space is inserted.

jcalfee:
It was hiding the word "low" ..  This fix is fine, but this needs some more researching.  There are other cases where spaces are being concatenated directly with the return value of `translate`.  I'm not sure why that messed though.  The DOM was actually missing the word "low"...

jcalfee:
I thought this was deployed already. Steemit.com is not showing this issue anymore.

roadscape:
@jcalfee the issue was w/ the locale file, here's the commit which adjusted it: https://github.com/steemit/steemit.com/commit/fbb5084725a0da151e7267adb290ef6be7f8d043

jcalfee:
@roadscape thanks..  Closing: do not merge...

Issue #641, opened 2016-11-17
middle-clicking (command-clicking) usernames in the list of users who upvoted a post should not close the list


sneak:

roadscape:
@mcifani I had relevant changes stashed away, I pushed them to `641-middle-click-dropdowns` (https://github.com/steemit/steemit.com/commit/bc25170d35c3f1508b630f59b2ddaf5362066382) Would you be willing to review/test that code?

mcifani:
@roadscape Tested and looks good.  Fixes both tickets #641 and #642.  Feel free to do a PR and will get it merged in.

PR #726, opened 2016-11-28
switch to using onclick, remove menu close event. resolves #722

### LINKED ISSUE #722: Dropdowns not working on mobile
When I am browsing my feed, the dropdown next to "home" that I would use to switch to "new" and "trending" does not work. I click on it, but the menu does not appear.

Also, the dropdown to see which users have voted on a post does not work either.

roadscape:
mousedown was used for the #641 fix because of an onClick conflict.

It appears the conflict was due to `closeMenu` in `VerticalMenu` -- as far as I can tell, we do not need this logic.

### DIFF
--- a/app/components/elements/DropdownMenu.jsx
+++ b/app/components/elements/DropdownMenu.jsx
@@ -24,7 +24,7 @@ export default class DropdownMenu extends React.Component {
     }
 
     componentWillUnmount() {
-        document.removeEventListener('mousedown', this.hide);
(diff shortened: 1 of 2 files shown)


mcifani:
Looks good.. also tested iOS/iphone 6 no issues.

roadscape:
Bug: some dropdowns do need to be force-closed on click. For instance the sort order dropdown on the homepage (in mobile view). Need to evaluate what is the most intuitive default behavior.

PR #738, opened 2016-12-01
Performance tweaks

svk31:
This is a collection of performance tweaks mostly focused on refactoring components to improve or add the shouldComponentUpdate checks, as well as some changes to which API methods trigger a global `loading` state in appReducer.

It also includes a css fix for the jumping 'back button header' in the overlay post view on mobile.

Some results:

Comment.jsx: 50 renders down to 6
Post.jsx: 23 renders down to 2
PostsIndex.js: 12 renders down to 6


app/components/pages/UserProfile.jsx
```diff
@@ -84,26 +113,21 @@ export default class UserProfile extends React.Component {
...
+        const followers = follow ? follow.getIn( ['get_followers', accountname] ) : null;
+        const following = follow ? follow.getIn( ['get_following', accountname] ) : null;
         if(followers && followers.has('result') && followers.has('blog')) {
             const status_followers = followers.get('blog')
             const followers_loaded = status_followers.get('loading') === false && status_followers.get('error') == null
             if (followers_loaded) {
-                followerCount = followers.get('result').filter(a => {
-                    return a.get(0) === "blog";
-                }).size;
+                followerCount = followers.get('count');
```
roadscape: Is this using 'count' from the new API? `a.get(0)` could be `blog` or `ignore` (mute), need to look into this

app/redux/FollowSaga.js
```diff
@@ -25,14 +24,17 @@ export function* loadFollows(method, follower, type, start = '', limit = 100) {
                 })
                 return m2
             })
-            return m.merge({[type]: {loading: true, error: null}})
+            const count = m.get('result').filter(a => {
+                return a.get(0) === "blog";
+            }).size;
+            return m.merge({count, [type]: {loading: true, error: null}})
```
roadscape: This answers my question from above, but needs testing

app/components/cards/PostsList.jsx
```diff
@@ -124,11 +138,11 @@ class PostsList extends React.Component {
...
             const {loadMore, posts, category} = this.props;
-            if (loadMore && posts && posts.length > 0) loadMore(posts[posts.length - 1], category);
+            if (loadMore && posts && posts.size) loadMore(posts.last(), category);
         }
 
         // Detect if we're in mobile mode (renders larger preview imgs)
-        var mq = window.matchMedia('screen and (max-width: 39.9375em)');
+        const mq = window.matchMedia('screen and (max-width: 39.9375em)');
         if(mq.matches) {
             this.setState({thumbSize: 'mobile'})
```
roadscape: May need to check for this in shouldComponentUpdate -- this changes the size of loaded thumbs when window is resized between mobile/desktop view

roadscape:
Code looks good, I left a few notes. Let's thoroughly test these changes.

@jcalfee could you please look over the commits?

app/components/cards/Comment.jsx
```diff
@@ -48,8 +47,9 @@ export function sortComments( g, comments, sort_order ) {
                 return bactive.getTime() - aactive.getTime();
               },
       new:  (a,b) =>  {
-                let acontent = g.get('content').get(a);
-                let bcontent = g.get('content').get(b);
+                let acontent = cont.get(a);
+                console.log("acontent", acontent);
```
svk31: Hm that's weird that was definitely removed in my local code, I'll see if I can add a commit for that.

app/components/pages/UserProfile.jsx
svk31: I just moved the counting logic from the component to the reducer, so it's still the old api.

app/components/cards/PostsList.jsx
svk31: `thumbSize` is part of `this.state` which is checked in `shouldComponentUpdate` so that should be fine still.

svk31: ignore that previous comment, was looking at the current version which uses the mixin, I'll add thumbSize to the new shouldComponentUpdate list.

app/components/pages/UserProfile.jsx
jcalfee: It should say `followers.get('block_count')`

jcalfee:
Sure, I'll have a look.

Issue #51, opened 2016-07-18
Wallet balance should reflect open market orders


jcalfee:
https://steemit.com/market/@logicwins/a-bug-in-the-steem-market

> When I put out an order, the dollar value of my steem and steem dollars does not show up in my wallet. The amount of steem dollars or steem used for the order should still be counted in my total, especially for interest calculations.

roadscape:
Related: #260 (conversion requests not reflected in wallet balance)

roadscape:
![](https://github.trello.services/images/mini-trello-icon.png) [Wallet balance needs to reflect conversions in progress and open orders #260, #51](https://trello.com/c/21P3w0dU/111-wallet-balance-needs-to-reflect-conversions-in-progress-and-open-orders-260-51)

PR #740, opened 2016-12-01
Add current open orders to UserWallet balances for logged in user, fi…

svk31:
…x #51


roadscape:
Looks great, do you think the amount should be preceded by a plus (+) sign to be more clear? I like that it links to the market page but at a glance it's not clear if the main balance reflects the order balance.

image

PR #741, opened 2016-12-01
Also update global account state on MarketSaga get_accounts, fixes #437

### LINKED ISSUE #437: Witness page not loading with votes selected
When I load the witnesss page, it loads without any of my selected witnesses selected and says I have 30 votes left. When I refresh the page it updates correctly. I was able to consistently reproduce with both Firefox and Chrome.

https://steemit.com/~witnesses

svk31:
I made the state.market.account object immutable as well which is why Market.jsx needed an update.

### DIFF
--- a/app/redux/MarketSaga.js
+++ b/app/redux/MarketSaga.js
@@ -73,7 +73,7 @@ export function* fetchOpenOrders(set_user_action) {
        if(account) {
            account = fromJS(account)
            yield put(MarketReducer.actions.receiveAccount({ account }))
-           yield put(g.actions.receiveAccount({ account }))
+           yield put(g.actions.receiveAccount({ account })) // TODO: move out of MarketSaga. See notes in #741
        }
     } catch (error) {
         console.error('~~ Saga fetchOpenOrders error ~~>', error);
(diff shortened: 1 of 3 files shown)


app/redux/MarketSaga.js
```diff
@@ -68,9 +69,12 @@ export function* fetchOpenOrders(set_user_action) {
...
         yield put(MarketReducer.actions.receiveOpenOrders(state));
 
-       const [account] = yield call(Apis.db_api, 'get_accounts', [username])
-       yield put(MarketReducer.actions.receiveAccount({ account }))
-
+       let [account] = yield call(Apis.db_api, 'get_accounts', [username])
+       if(account) {
+           account = fromJS(account)
+           yield put(MarketReducer.actions.receiveAccount({ account }))
+           yield put(g.actions.receiveAccount({ account }))
```
roadscape: Is `yield put(g.actions.receiveAccount({ account }))` required for the witness page fix? Is there possibly a more relevant place to put this line, just in case `fetchOpenOrders` is no longer called on every page load in the future?

svk31: Hm you're right actually. The issue is that fetchOpenOrders is the only place that calls get_accounts with your account when you log in. If we can put that in `loginWatch` in UserSaga for example that would be better, or we can add a special case in `fetchState`.

PR #742, opened 2016-12-01
Handle for malformed case on top categories

mcifani:
Handle cased category filtering


jcalfee:
@mcifani please review and merge if you like it

PR #751, opened 2016-12-02
Sort transfer log by date fix #750

### LINKED ISSUE #750: Transfer history appears in reverse order

svk31:


roadscape:
The only other place I can think of that uses this list is the author/curation rewards. But those appear to already be sorted in the correct order.

Also, running develop against the current production nodes, I see the transaction history is still in reverse. This suggests it was not an API change like I thought it was

svk31:
I see why this happened now: I removed the fromJS() conversion of the account object, so the result of the map() is now an immutable object, which means that the `transfer_log.reverse()` called just below the map loop doesn't do anything. Fixed in my new commit.

PR #774, opened 2016-12-05
745 refactor follow data struc

jcalfee:


app/components/pages/UserProfile.jsx
```diff
@@ -152,46 +138,46 @@ export default class UserProfile extends React.Component {
         if( section === 'transfers' ) {
             walletClass = 'active'
             tab_content = <div>
-                <UserWallet
-                          account={accountImm}
-                          showTransfer={this.props.showTransfer}
-                          current_user={current_user}
-                          withdrawVesting={this.props.withdrawVesting} />
+                <UserWallet global={this.props.global}
```
roadscape: Is using the global here necessary? @svk31 has been cutting down on passing around the whole state for performance.

svk31: Good catch, global is not used in UserWallet anymore so isn't needed there.

jcalfee:
I did think of one more change.. I moved _inprogress completely outside of the follow tree.  The components tend to pass around `follow` at the root, moving in-progress out of there prevents extra re-renders in that case.

I removed global from UserWallet.

PR #785, opened 2016-12-06
783 meta links

svk31:
Adds 'http://' to urls in the case it's missing, and opens the links in a new tab.

### DIFF
--- a/app/components/pages/UserProfile.jsx
+++ b/app/components/pages/UserProfile.jsx
@@ -352,11 +352,9 @@ export default class UserProfile extends React.Component {
             </div>
          </div>;
 
-        let {name, location, about, website} = normalizeProfile(account);
-        if (website && website.indexOf("http") === -1) {
-            website = 'http://' + website;
-        }
+        const {name, location, about, website} = normalizeProfile(account);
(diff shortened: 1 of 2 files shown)


app/components/pages/UserProfile.jsx
```diff
@@ -352,9 +352,11 @@ export default class UserProfile extends React.Component {
             </div>
          </div>;
 
-        const {name, location, about, website} = normalizeProfile(account);
+        let {name, location, about, website} = normalizeProfile(account);
+        if (website && website.indexOf("http") === -1) {
+            website = 'http://' + website;
```
roadscape: Let's move this logic to normalizeProfile

app/components/pages/UserProfile.jsx
```diff
@@ -387,7 +389,7 @@ export default class UserProfile extends React.Component {
                             </div>
                             <p className="UserProfile__info">
                                 {location && <span><Icon name="location" /> {location}</span>}
-                                {website && <span><Icon name="link" /> <a href={website}>{website_label}</a></span>}
+                                {website && <span><Icon name="link" /> <a target="_blank" href={website}>{website_label}</a></span>}
```
roadscape: We can't use target=_blank for security reasons

PR #786, opened 2016-12-06
608 @author/permlink redirect

svk31:
This PR adds a redirect component that looks up the post data using `get_content` and then redirects to /category/@user/permlink. If the call goes to the server `get_content` is used there also which makes the redirect happen quicker on the client side.

Since we can now link to /@author/permlink this PR also includes a fix for the curation reward links.


svk31:
Thanks. That was the initial idea but it was necessary to do it client side also.

Issue #464, opened 2016-10-11
Add ability to proxy vote for witness


valzav:

roadscape:
![](https://github.trello.services/images/mini-trello-icon.png) [Add ability to proxy vote for witness #464](https://trello.com/c/3fXmp50S/130-add-ability-to-proxy-vote-for-witness-464)

PR #797, opened 2016-12-06
Add ability to set witness proxy voter, #464

svk31:


sneak:
@svk31 please add the translation keys to the other localization files too with their english texts and tag them with a FIXME comment.

sneak:
@svk31 i love additional functionality but this one is only useful to a very small community of whales, witnesses, bitshares holdovers, et c. and doesn't move the needle much on user acquisition/retention which is our primary focus.  See Trello for priorities regarding those.

app/locales/en.js
```diff
@@ -388,6 +388,12 @@ const en = 	{
 	witness_thread: 'witness thread',
 	you_have_votes_remaining: 'You have {votesCount} votes remaining',
 	you_can_vote_for_maximum_of_witnesses: 'You can vote for a maximum of 30 witnesses',
+	set_witness_proxy: "You can also choose a proxy that will vote for witnesses for you. This will reset your current witness selection.",
```
sneak: please add these (as english, tagged with a `FIXME` comment) to the other language files too.

svk31:
I've added the strings as requested.

PR #798, opened 2016-12-06
Login form transfer updates

svk31:
image

@valzav asked me to add the @ prefixes to account inputs and I noticed there was room for improvement in the asset selection dropdown so I restyled it in the same vein.

The LoginForm would not set the initial value for the account for users logging in with private keys, so I've added that to the connect method initialisation. The account input also gets an @ prefix.

### DIFF
--- a/app/components/modules/Transfer.jsx
+++ b/app/components/modules/Transfer.jsx
@@ -151,7 +151,7 @@ class TransferForm extends Component {
                 <div className="row">
                     <div className="column small-2" style={{paddingTop: 5}}>From</div>
                     <div className="column small-10">
-                        <div className="input-group">
+                        <div className="input-group" style={{marginBottom: "1.25rem"}}>
                             <span className="input-group-label">@</span>
                             <input
                                 className="input-group-field bold"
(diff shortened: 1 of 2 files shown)


sneak:
Please try to avoid inline styles whenever possible, instead favoring CSS.

app/components/modules/Transfer.jsx
```diff
@@ -170,24 +187,30 @@ class TransferForm extends Component {
...
+                            </span>}
+                        </div>
+                        <div style={{marginBottom: "0.6rem"}}>
+                            <AssetBalance balanceValue={this.balanceValue()} onClick={this.assetBalanceClick} />
+                        </div>
+                        {(asset && asset.touched && asset.error ) || (amount.touched && amount.error) ?
+                        <div className="error">
+                            {asset && asset.touched && asset.error && asset.error}&nbsp;
+                            {amount.touched && amount.error && amount.error}&nbsp;
+                        </div> : null}
```
svk31: I agree, I don't like this unary syntax that's being used here.

PR #801, opened 2016-12-07
380 unknown account flash

svk31:


sneak:
@valzav in the future when you are approving (second set of eyeballs) please go ahead and merge as well.  Merging this myself without review based on @svk31 and @valzav approvals.

PR #802, opened 2016-12-07
Add checks for loading and loaded in FollowSaga. (close #799)

### LINKED ISSUE #799: Possible Follower Count Bug
steemit/steem#671

> As stated I just noticed the number of people I am following is now showing as 61 (instead of 753). If I click on the link it only allows me to view 61 names. I have already had one user asking me why I unfollowed. I am assuming this is a bug? My follower number has not changed. Here is a post I made about this. It seems I am not the only experiencer of this issue:

> https://steemit.com/steemit/@thecryptofiend/possible-bug-over-90-of-people-i-am-following-have-vanished-since-the-fork-down-from-753-to-61

> I thought I should let you all know ASAP. Thanks.

---

> I see the same issue. My profile cycles between 143 and 1034 people followed.
> If I hard refresh my profile page it shows 1034 for a moment, before changing back to 143

jcalfee:


roadscape:
👍

PR #804, opened 2016-12-07
Display conversions in progress, add them to account value, fix #260

### LINKED ISSUE #260: Wallet page should indicate in-progress conversions
Showing at least the total amount of SBD conversions in progress (if not the entire list detail), and adding it to total estimated value, would address the current issue where in-progress conversions appear to simply disappear from the balance and then suddenly reappear when the conversion is complete. This is confusing and has caused some concern by users.

svk31:


app/components/modules/UserWallet.jsx
```diff
@@ -87,19 +87,36 @@ class UserWallet extends React.Component {
             })
         }
 
+        // Sum conversions
+        let conversionValue = 0;
+        const conversions = account.get('other_history').filter(a => {
+            return a.getIn([1, 'op', 0]) === 'convert';
+        }).map(c => {
+            const timestamp = new Date(c.getIn([1, 'timestamp'])).getTime();
+            const finishTime = timestamp + (86400000 * (timestamp <= 1481040000000 ? 7 : 3.5)); // add conversion delay before/after hardfork change
```
sneak: please pull logic like this out into appropriately-named functions so that someone reading this code gets a semantic hint (i know what it does today, but in 8 months i will absolutely not remember the specifics of HF16).

svk31: In principle I agree, but in this case I intend to just remove this logic next week since it won't be needed anymore. It only applies in the first week following HF16.

PR #818, opened 2016-12-08
Add zebra stripes to wallet page fix #813

### LINKED ISSUE #813: zebra stripes for wallet table list
Please configure the rows corresponding to "steem", "steem power", "steem dollars", and "savings" to alternate between a very-light-grey and a medium-grey.

It is super difficult to visually determine which balance (far over on the right side) matches up with which label (far over on the left hand side).  Just the other day when looking at the `@steem` account I misread the Steem Power value as the STEEM value.

svk31:


svk31:
Feel free to adjust the color, I just picked one I felt was fairly neutral.

PR #604, opened 2016-11-08
translation of markets

roadscape:

PR #821, opened 2016-12-08
translation of blocktrades deposit

roadscape:
pulled out of #604.


roadscape:
Should be good to go.

PR #832, opened 2016-12-12
Disable follow buttons while busy. close #830

### LINKED ISSUE #830: unfollowing then muting someone gives a "user has already transacted this block" error

jcalfee:


roadscape:
Tested, works well. I added a small change to prevent broadcasting while `busy == true` -- even though the label looks disabled, it was still clickable. Cool?

PR #842, opened 2016-12-13
Update react and babel

valzav:


roadscape:
continued in #843 due to stray commit

PR #851, opened 2016-12-14
844 username routes 404

mcifani:


roadscape:
Someone recently refactored route logic so the useprofile regex only lives in `app/ResolveRoute.js:4`:

```
UserProfile2: /^\/(@[\w\.\d-]+)\/(blog|posts|comments|recommended|transfers|curation-rewards|author-rewards|permissions|created|recent-replies|feed|password|followed|followers|settings)\/?$/,
```

Can we leverage this instead of using `userEndpoints`?

mcifani:
@roadscape Did a bit of refactoring, going forward all the route paths/matches can live here.   1.  moved settings into the resolve route(also refactored some other patterns we had for good practice).  2. For the particular matching it's helpful to have the additional variation, but it now lives in the global route area.

PR #861, opened 2016-12-15
827 nsfw handling

roadscape:
The purpose of this PR is to increase the visibility of NSFW content while allowing the user some degree of control over how it's displayed.

Related: https://github.com/steemit/steem/pull/705


app/utils/StateFunctions.js
```diff
@@ -76,5 +76,16 @@ export function contentStats(content) {
...
+    // Combine tags+category to check nsfw status
+    let json = content.get('json_metadata')
+    let tags = []
+    try {
+        tags = json && JSON.parse(json).tags || [];
+    } catch(e) {
+        tags = []
+    }
+    tags.push(content.get('category'))
+    const isNsfw = tags.includes('nsfw');
```
roadscape: If we're going the non-regex route we will need to make sure all tags get downcased first.. I  believe they can still come through with caps at times.

jcalfee: Ok .. I reverted back to the case insensitive version just in case..

sneak:
Couple of issues:

1) it doesn't change the 'create an account' prompt on the stub if you log in, you have to reload the page.

2) on the settings page, the user profile data requires a save button to save, the nsfw hide/prompt/show popup does not.  i understand why that is, but the switch in the modes of preference saving is not clear or explained to the user.  Ultimately, this should be saved in our db and not put onto a user to configure on each device they log in from.

PR #871, opened 2016-12-16
Refactor class name for best practice conventional naming, tweak prom…

mcifani:
…ote for flush alignment with parent container


mcifani:
New(bottom)/Old screenshot comparison. 
screen shot 2016-12-16 at 4 19 43 pm

PR #883, opened 2016-12-20
Patch route user

mcifani:


jcalfee:
>	Tweak translate setting for followed/follower count
What does this look like before the follow count is loaded?

jcalfee:
I see.. still works .. ok nevermind .. thanks

Issue #894, opened 2016-12-22
account page 500 error


roadscape:
https://steemit.com/@rmach

https://steemit.com/steemit/@rmach/some-kind-of-error-when-accessing-my-blog-did-it-happen-to-anyone

```
SteemApi exec 0 get_state ( [ '/@rmach' ] )
State/store error:  TypeError: Cannot read property 'match' of null TypeError: Cannot read property 'match' of null
    at app/utils/StateFunctions.js:93:39
    at Array.filter (native)
```

```
    const isNsfw = tags.filter(tag => tag.match(/^nsfw$/i)).length > 0;
```

PR #897, opened 2016-12-22
fix account page 500 error #894

roadscape:
ensure tag is not null before attempting to match


mcifani:
@roadscape I believe this now resolves https://github.com/steemit/steemit.com/issues/894 - can you just confirm?  Thanks.

roadscape:
@mcifani yep, it does!

PR #904, opened 2016-12-23
prepping for dockerization.

sneak:


config/defaults.json
```diff
@@ -0,0 +1,135 @@
...
+  "cli_wallet_connection": "",
+  "db": {
+    "dialect": "sqlite",
+    "language": "en",
+    "storage": "/tmp/db.sqlite",
+    "sync": {
+      "force": true
+    }
+  },
+  "db-prod-example": {
```
valzav: What is the purpose of db-prod-example entry?

config/defaults.json
```diff
@@ -0,0 +1,135 @@
+{
+  "cli_wallet_connection": "",
+  "db": {
```
valzav: Not sure it will work with sqlite. Did anybody test is with sqlite?

config/defaults.json
sneak: Documentation for people who want to see how it works with mysql. JSON does not allow comments.

config/defaults.json
sneak: I did basic testing. API event logging works. Page view counters work. Haven't tested signups etc.

valzav: That's nice - sqlite is much easier to install

config/defaults.json
sneak: It ultimately is temporary as the defaults will be set in js soon enough.

config/defaults.json
sneak: Yeah I figure fewer developer deps is a win. I think we should update your raw queries to use orm methods though for max compatibility.

sneak: (And even if we don't, we should use prepared statements with placeholders instead of string interpolation/concatenation.)

sneak: I would love to switch to Postgres in prod eventually, it is vastly preferable to mysql.

sneak:
different effort

PR #906, opened 2016-12-25
Christmas Steem Icon

roelandp:
Just in time christmas steem icon. 

Updated: Header.jsx, Icon.jsx, and a new xmas icon file


roadscape:
I guess no Santa this year 😉

PR #917, opened 2016-12-29
Add a last-minute check to ensure one phone per account

jcalfee:


jcalfee:
This is not the issue, don't merge..

PR #922, opened 2016-12-30
Fix phone verification issue

valzav:


server/server_pages/enter_confirm_mobile.jsx
```diff
@@ -43,10 +43,18 @@ function *confirmMobileHandler() {
         return;
     }
 
-    const used_phone = yield models.sequelize.query(`SELECT a.id FROM accounts a JOIN identities i ON i.user_id=a.user_id WHERE i.phone='${mid.phone}'`, { type: models.Sequelize.QueryTypes.SELECT})
-    if (used_phone && used_phone.length > 0) {
-        this.flash = {error: 'This phone number has already been used'};
-        this.redirect('/enter_mobile');
+    // const used_phone = yield models.sequelize.query(`SELECT a.id FROM accounts a JOIN identities i ON i.user_id=a.user_id WHERE i.phone='${mid.phone}'`, { type: models.Sequelize.QueryTypes.SELECT})
+    const used_phone = yield models.Identity.findOne(
+        {attributes: ['id', 'user_id'], where: {phone: mid.phone, provider: 'phone', verified: true}, order: 'id DESC'}
```
sneak: thank you for using the ORM as intended! 😄

server/server_pages/enter_confirm_mobile.jsx
```diff
@@ -43,10 +43,18 @@ function *confirmMobileHandler() {
...
+    );
+    if (used_phone) {
+        if (used_phone.user_id === this.session.user) {
```
sneak: I don't think you need to do separate flash messages here - you can just throw a single error without the nested if and redirect.  I'll approve it anyway but if you want to take that out, feel free.

PR #934, opened 2017-01-02
914 user thumbnails

mcifani:
screen shot 2017-01-02 at 2 29 54 pm
 Comment thumbnails(and also profile page thumnail) now match sharpness of profile snippet image quality on retina and 4k displays.

New example below:  
screen shot 2017-01-02 at 2 28 58 pm

### DIFF
--- a/app/components/elements/Userpic.jsx
+++ b/app/components/elements/Userpic.jsx
@@ -20,7 +20,7 @@ class Userpic extends Component {
         } catch (e) {}
 
         if (url && /^(https?:)\/\//.test(url)) {
-            const size = width && width > 48 ? '320x320' : '90x90';
+            const size = width && width > 48 ? '320x320' : '120x120';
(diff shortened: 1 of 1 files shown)


app/components/elements/Userpic.jsx
```diff
@@ -20,7 +20,7 @@ class Userpic extends Component {
         } catch (e) {}
 
         if (url && /^(https?:)\/\//.test(url)) {
-            const size = width && width > 48 ? '320x320' : '72x72'
+            const size = width && width > 48 ? '320x320' : '120x120';
```
roadscape: Can we try 96x96? Since 48px is the cutoff, with retina @2x the optimal size would be 96px

mcifani: 96x96 works good for the retina 13/15", but on 4k still comes soft since it's around 108.79 PPI/27" @4x.  At the current dimensions it has full sharpness on a 32" 4k.

PR #937, opened 2017-01-03
915 resteem button feeds mobile

mcifani:


mcifani:
Preview re-steem now in collapsed state post preview area
screen shot 2017-01-03 at 12 37 36 pm

PR #940, opened 2017-01-03
Add 3x support for thumbnail previews for post previews to remove fuz…

mcifani:
…zy display

No longer "fuzzy" post thumbnail previews up to 5K displays.  Preview below. 
screen shot 2017-01-03 at 2 34 14 pm
screen shot 2017-01-03 at 2 34 16 pm

### DIFF
--- a/app/components/cards/PostSummary.jsx
+++ b/app/components/cards/PostSummary.jsx
@@ -149,7 +149,7 @@ class PostSummary extends React.Component {
         let thumb = null;
         if(pictures && p.image_link) {
           const prox = $STM_Config.img_proxy_prefix
-          const size = (thumbSize == 'mobile') ? '640x480' : '128x256'
+          const size = (thumbSize == 'mobile') ? '640x480' : '256x384';
           const url = (prox ? prox + size + '/' : '') + p.image_link
           if(thumbSize == 'mobile') {
             thumb = <a href={p.link} onClick={e => navigate(e, onClick, post, p.link)} className="PostSummary__image-mobile"><img src={url} /></a>


app/components/cards/PostSummary.jsx
```diff
@@ -149,7 +149,7 @@ class PostSummary extends React.Component {
         let thumb = null;
         if(pictures && p.image_link) {
           const prox = $STM_Config.img_proxy_prefix
-          const size = (thumbSize == 'mobile') ? '640x480' : '128x256'
+          const size = (thumbSize == 'mobile') ? '640x480' : '256x384';
```
roadscape: Important to keep the aspect ratio; this may need to be `256x512` if I understand the change correctly

mcifani: The res size here shouldn't be altering the aspect/cropping of the thumbnail(check out the examples).  Theoretically if we can use the smaller resolution, benefit with a smaller size img to serve, and eliminate the bad quality thumbnails on the higher displays.

mcifani: Here's a few examples as well(you can load them side by side)

https://img1.steemit.com/256x584/http://www.impawards.com/2016/posters/moana_ver5.jpg
https://img1.steemit.com/256x384/http://www.impawards.com/2016/posters/moana_ver5.jpg

https://img1.steemit.com/256x584/http://i.imgsafe.org/501676a6b1.jpg
https://img1.steemit.com/256x384/http://i.imgsafe.org/501676a6b1.jpg

PR #942, opened 2017-01-03
prepend nested comments with dots

valzav:


valzav:
image

sneak:
:(

valzav:
@sneak it helps to distinguish root comments from replies, the dots are very subtle so they will not add much clutter. I will revert this if there will be some negative feedback.

sneak:
I understand the purpose and the problem it attempts to solve is a real one; i just don't think this is the right solution. 

The correct long term answer is to redesign the comment tree and layout, but in the short term, there are better ways than putting all the visual clutter in there.

PR #953, opened 2017-01-05
restore thumb ratio for crops #940

roadscape:
Reference: https://github.com/steemit/steemit.com/pull/940/files

The crop bounds was 2:1 for thumbnails to account for cases of tall images. The current CSS-based cropping attempts to fit images widthwise -- when the bounding box is more of a square, tall images become very low res and in some cases stretched out.

@mcifani the example image URLs you listed all had a height that was less than 2x the width, so there is no difference in the generated image. But it does make an important difference for images which do not fit this case.


mcifani:
Makes sense and looks good.

PR #968, opened 2017-01-09
Allow for comments to render voter list, unify post/comment footer vo…

### LINKED ISSUE #439: upvoters not displayed on user's profile's comments tab
If I'm reviewing my own comments, I can't see who upvoted them or when.

mcifani:
…te list

PR fixes #439. Also seemingly for best UI practice this also adds uniformity in all the preview footer area(payout amount, vote number(with list)).  Few changes to review: 

1.  All post previews snippets have consistent style/UI footer.  Payout amount and voter list dropdown.
2. Prior 0 votes no indicator was rendered/now indicated with "0 votes". 

Before/after

screen shot 2017-01-09 at 5 21 27 pm
screen shot 2017-01-09 at 5 24 54 pm

### DIFF
--- a/app/components/cards/PostSummary.jsx
+++ b/app/components/cards/PostSummary.jsx
@@ -125,7 +125,7 @@ class PostSummary extends React.Component {
         </span>
 
         const content_footer = <div className="PostSummary__footer">
-            <Voting post={post} showList={false} />
+            <Voting post={post} showList={true} />
             <VotesAndComments post={post} commentsLink={comments_link} />
             <span className="PostSummary__time_author_category">
                 {!archived && <Reblog author={p.author} permlink={p.permlink} />}
--- a/app/components/elements/VotesAndComments.jsx
+++ b/app/components/elements/VotesAndComments.jsx
@@ -32,9 +32,6 @@ class VotesAndComments extends React.Component {
 
         return (
             <span className="VotesAndComments">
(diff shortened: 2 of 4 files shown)


roadscape:
I originally removed the "0 votes" display for comments because it seemed to become quite repetitive/useless:

image

What do you think about adding a parameter so we can configure how it's displayed on different views? Also it would be good to keep the icon on PostSummary IMO.

mcifani:
Agree on the repetitive nature.  Perhaps another elegant route would be to revert the "0 votes" and also keep the "vote" icon removed.  This would give the uniformity on all the scenarios, we could also consolidate some code(instead of the two components for the footer). The uniformity in the footer seems ideal also with the proposed comment reward changes.

valzav:
@mcifani this adds comments drop down to all previews including those on index pages - I would try to avoid this because it adds visual noice to index pages and also will add up in render time because now you need to render a lot of drop downs per page.

jcalfee:
Just hide when votes === 0

valzav:
@jcalfee I don't think it would be enough to hide drop down - it should render old 
                <span className="VotesAndComments__votes" title={translate('vote_count', {voteCount: voters_count})}>
                    <Icon name={voters_count > 1 ? 'voters' : 'voter'} />&nbsp;{voters_count}
                </span>
on all index pages

roadscape:
Let's put this on the backburner for now; there are some incoming changes which make this trickier. With SBDS a lot of this code will need to change and it will be easier.

PR #846, opened 2016-12-13
Translator improvements

Undeadlol1:
Adds:

- stand alone translate() function which can translate anything outside of react context
- user’s browser language detection and auto selection of language (currently disabled to avoid errors due to non-fixed translation files (it.js, fr.js, es.js)
- fixes client_config
- fixes locale data polyfills

PR #972, opened 2017-01-10
restore a const which was removed in #846

roadscape:
Was causing bugs on pages which rely on VESTS and assetFloat method


roadscape:
https://github.com/steemit/steemit.com/pull/846/files#diff-56dd992e1534ccf66830a782703819d0L14

PR #1026, opened 2017-01-19
396 markdown image upload

jcalfee:
Configure to use the imagehoster via: `"uploadImage": "https://..."`


sneak:
I didn't see it in this PR - but if the client key-based auth to the imagehoster is signing the challenge and not a hash of the challenge, the challenge needs to be prefixed with a constant to make sure the server can't easily make the client sign a transaction doing something else. @jcalfee

jcalfee:
Prefix challenge is in production and stable...

app/redux/UserSaga.js
```diff
@@ -429,25 +431,31 @@ function* uploadImage({payload: {file, dataUrl, filename = 'image.txt', progress
     const sig = Signature.signBufferSha256(bufSha, d)
     const postUrl = `${$STM_Config.uploadImage}/${username}/${sig.toHex()}`
```
sneak: all the other keys in $STM_Config are snake_case and not camelCase.

roadscape:
Tested working, just needs some server-side config.

PR #1028, opened 2017-01-20
Update Resteem Confirm to "Resteem This Post as @username"

TimCliff:
For users who use multiple accounts, it is easy to accidentally resteem under the wrong account.

The current Resteem confirmation looks like this:
image

This update will change it so that the confirmation includes the username of the logged in user:
image

If there are any changes to the logic/formatting/etc. that would be preferred over this implementation, just let me know and I can make the necessary updates.

### DIFF
--- a/app/components/elements/Reblog.jsx
+++ b/app/components/elements/Reblog.jsx
@@ -87,7 +87,7 @@ module.exports = connect(
                     id: 'follow',
                     required_posting_auths: [account],
                     json: JSON.stringify(json),
-                    __config: {title: 'Resteem This Post'}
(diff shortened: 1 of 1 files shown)


valzav:
steemit.com currently doesn't support multiple accounts - we are targeting regular, not power users. Adding this support would require to add @account names to virtually all dialogs and confirmations - I believe this will clutter UI providing value to less than 1% of our users. Just imagine if FB or Twitter would add Share or Like as SomeUserName to their dialogs.

TimCliff:
Ok, makes sense.

PR #1037, opened 2017-01-20
Parse metadata when returning user JSON

mcifani:


server/json/user_json.jsx
```diff
@@ -18,6 +18,7 @@ export default function useUserJson(app) {
 
         if(chainAccount) {
             user = chainAccount;
+            user.json_metadata = user.json_metadata ? JSON.parse(user.json_metadata) : "";
```
sneak: @mcifani will this barf if the user has put invalid json into their profile? should this be in a try/catch?

mcifani: @sneak This conditional will actually catch for users with/without JSON metadata.  If the user has no JSON it won't parse/empty string.  In the case of malformed data it would seem the backend should be validating on the save. But I added the try/catch.

Issue #1032, opened 2017-01-20
brand logo disappears from top navbar when viewport is medium-sized on post view


sneak:
There's plenty of empty space there that isn't being used, the logo should make an appearance there.

mcifani:
Resolved in https://github.com/steemit/steemit.com/pull/1043/

PR #1043, opened 2017-01-20
#1032 brand logo disappears from top navbar when viewport is medium-s…

mcifani:
…ized on post view


roadscape:
One thing I noticed (and I confirmed this happens on `master` as well) is that the top bar sometimes scrolls a bit--not sure if it's worth fixing

image

Issue #872, opened 2016-12-17
iframes not hidden like images on partially-shown posts due to non-positive rep or negative rshares


pfunks:
Just like images are hidden after clicking "show", iframes such as youtube should probably be as well.

jcalfee:
Do you have a link?

pfunks:
Sure https://steemit.com/testing/@yak/a-handful-of-iframe-embeds-for-your-perusal

PR #1087, opened 2017-02-01
unlinkify exe and zip links

roadscape:


roadscape:
input:
```
<a href="http://google.com/test.exe">download</a>

<a href="http://google.com/test.exe">ssss&lt;strong&gt;<strong>dddd</strong></a>

Test test http://google.com/test.exe
```

output:
image

roadscape:
from james:
 - escape `.`
 - now that linkify happens first, urls may be added twice to `links` list
  - check iframe handling too (gets rid of dupe logic, and simplifies #872)

roadscape:
Another issue is that if @-mentions and #-tags are linkified, they will be put into `links` list. These cases might be best handled in a separate/final stage

roadscape:
htmlready clearly needs some work. I've put the beginning changes in https://github.com/steemit/steemit.com/tree/update-htmlready

for the sake of this PR I've gone with a simpler approach (avoid linkifying raw URLs only) as a refactor is out of scope

PR #1101, opened 2017-02-03
Adjust author heading

roadscape:
before:

image

after:

image


valzav:
It wouldn't be nice if you could vertically align left and right parts and match the font size

valzav:
This was incomplete - not adjusted for mobile. Also I think now there is less balance between image and text than it used to be.

sneak:
Then revert it back to before anyone touched it, please.

sneak:
FWIW our article-view overlay modals are pretty terribly broken on mobile as well.  I would like to see us remove the two different views for the article view and just have one.  (Users know how to use the back button, or to middle-click.)

PR #1104, opened 2017-02-04
Move witness proxy voting to the bottom of the witness voting page

TimCliff:
The witness voting proxy is an excellent feature to have in the GUI, but it is a feature that is only useful to a subset of users. Having at the top of the witness voting page does not seem like the best placement, since (hopefully) the "average" user is going to want to be involved in the witness voting process and vote for witnesses themselves.

image

This pull request moves the witness voting proxy to the bottom of the witness voting page.

image

I have verified that the functionality still works as expected when placed at the bottom of the page.

image

### DIFF
--- a/app/components/pages/Witnesses.jsx
+++ b/app/components/pages/Witnesses.jsx
@@ -134,35 +134,6 @@ class Witnesses extends React.Component {
                     </div>
                 </div>
 
-                <div className="row">
-                    <div className="column">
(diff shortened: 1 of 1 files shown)


Transisto:
A small link to a page "What are a Steem witnesses" or "What are the roles of witness" might also help a lot the average users.

TimCliff:
It's a great suggestion! I'll work on something for that. I'll handle it in a separate pull request though.

PR #1106, opened 2017-02-05
(WIP) Adding Table of Contents to FAQ Page - Feedback Requested

TimCliff:
@pfunks and I are are working on changes to the FAQ page to add a lot of additional content. As part of the changes, we will be adding a table of contents. The purpose of this PR is to get feedback on the TOC part of the change, so that we can determine the correct conventions/format/structure/style to follow for the larger FAQ.

The TOC will be at the top of the page, and will have links to each question:

image

The questions will be organized into categories, but for the purpose of this PR I just used fake/sample categories. This is so I didn't have to move the questions around, which would have introduced a bunch of unnecessary differences to the file. (In the final version, we will decide on actual categories and move the questions around to be in the right ones.)

Clinking on a link brings the user to that question in the TOC. At the bottom of each question's content, there is a link back to the TOC:

image

Please provide feedback on the changes.

### DIFF
--- a/app/assets/stylesheets/app.scss
+++ b/app/assets/stylesheets/app.scss
@@ -123,3 +123,4 @@ a, path, circle {
   }
 }
 
+.anchor { padding-top: 90px; }
--- a/app/help/en/faq.md
+++ b/app/help/en/faq.md
@@ -1,23 +1,69 @@
 # Steemit FAQ 
 
+<a class="anchor" name="Table_of_Contents"></a>
+## Table of Contents
+
+### Sample Category 1
+
+- <a href="/faq.html#What_is_Steemit_com">What is Steemit.com?</a>
+- <a href="/faq.html#What_is_the_Steem_blockchain">What is the Steem blockchain?</a>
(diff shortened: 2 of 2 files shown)


TimCliff:
I am mainly looking for feedback on the following items:
- Is the solution to use anchor links an acceptable way to add the TOC to the page?
- What naming convention should I use for the anchor links?
- Is using the `padding-top: 90px;` CSS class an acceptable way to deal with the shift that occurs on the page due to the top level menu?

valzav:
@TimCliff everything looked good to me so I've merged it, but didn't notice "sample categories", could you add real categories via another PR? 
Aside note: please use rem units instead of px.

TimCliff:
Sure, will do

TimCliff:
@valzav - I wasn't able to come up with a good set of categories with the content that was there currently. It would have ended up with lots of categories with only one item, or several very generic ones with tons of semi-related items. It made the most sense to just remove the categories for now until we can add more content to balance them out.

PR #1120, opened 2017-02-07
1118 update json profile

mcifani:
#1118 Updates


roadscape:
Which bug is this fixing?

mcifani:
This wasn't a bug but changing the user profile JSON route to .json requested by @sneak

roadscape:
👌

PR #1137, opened 2017-02-10
add payout sort [HF17]

roadscape:
this will be implemented with HF17 release

needs testing as posts + comments will be shown alongside


roadscape:
Need to evaluate placement of these links in the header. @valzav any suggestions?

image

I'm not sure they need to be so prominent. The content on them won't cycle very fast and it's useful mainly just for those needing to review payouts.

mcifani:
Perhaps putting just "payout" in the top level submenu, with a 2nd tier menu below the hr with posts/comments toggles.

PR #1144, opened 2017-02-13
show fill_convert_request and fill_order ops in wallet

roadscape:
this works with steemd/master. not 0.16.3.


roadscape:
Safe to merge

Issue #1194, opened 2017-02-19
Post quote should not be medium gray


thrize:
I know it's a small thing but I think that the quote text should be just a little bit darker than #8B8B8B.

It's hard for the eyes to latch onto the text and I worry that quotes get skipped a lot because of this.

A bit darker while still gray would be great, for example I think #525252 would look great.

Thanks 🙂

mcifani:
I think this is a good idea, but recommend something slight changed to rgba(0,0,0,.6).  Also, I think  if the quote text was "slightly" larger it would be beneficial.  See screenshots for proposal @roadscape your thoughts?  
<img width="881" alt="screen shot 2017-02-20 at 11 03 23 am" src="https://cloud.githubusercontent.com/assets/1158463/23133571/96adc642-f75f-11e6-83c3-16ca3f2fd8be.png">
<img width="883" alt="screen shot 2017-02-20 at 11 27 27 am" src="https://cloud.githubusercontent.com/assets/1158463/23133574/9a413136-f75f-11e6-8f21-93829d066620.png">

roadscape:
@mcifani looks great.

thrize:
The color is good like that, great 😁

I'm not so sure about making it bigger, I'd actually go the other way! Like 2 points smaller.

thrize:
@mcifani why did you close this?

mcifani:
@thrize We pushed a pending PR for updates. Thanks for your original suggestion.

thrize:
Wonderful, thank you!

sneak:
Be careful with larger blockquotes, frequently they are not used only as pull-quotes but inline replies to e.g. parent comments.

mcifani:
@sneak Agree, it's a bit of a catch.  Larger blockquotes are impactful when used well and highlight a thoughtful quote - but ultimately agreeing that should be styled by the poster as they are used in different context.

PR #1198, opened 2017-02-20
1194 post quote update

mcifani:


mcifani:
Handles 	#1194, a few updates:

- sets a slight darker tone to blockquotes.
- sets a slightly larger size(full post only) to blockquotes.  In context of an article/post this gives them slightly more importance.

Issue #1161, opened 2017-02-16
follow/mute buttons missing on logged out profile view


sneak:
<img width="793" alt="screen shot 2017-02-15 at 20 33 51" src="https://cloud.githubusercontent.com/assets/408977/23003335/0dbb6ac8-f3be-11e6-95c1-a0c89513102f.png">

The follow/mute buttons should remain on logged out views, popping up a big "login/signup" modal.  It's an opportunity to convert people.

sneak:
Oh, also, the padding between the bottom of the profile image and the top of the user's descriptive text should be increased 50-100%.

mcifani:
Updates in PR https://github.com/steemit/condenser/pull/1217

PR #1217, opened 2017-02-28
1161 follow mute missing

mcifani:


mcifani:
This addresses ticket #1161 

-  New padding in user preview card and userpic/bio
-  Follow button is now included in "logged out state" on both preview card and profile page.  This will try to authenticate/or convert new users.  This should help increase conversions of new users who want to follow users, but otherwise wouldn't see this option if they weren't logged in.

Issue #1159, opened 2017-02-16
command-clicking "submit a post" does not open it in a new tab like middle clicking does.


sneak:
same for links in the right hand tray

mcifani:
Fix in PR https://github.com/steemit/condenser/pull/1218

PR #1218, opened 2017-02-28
#1159 conditional to prevent anchor tags breaking on default behavior…

mcifani:
… for cmd clicking


mcifani:
Fix for ticket #1159 

- Conditional to protect from break in anchor click

Issue #1158, opened 2017-02-16
middle-clicking/command-clicking an author's username in a post view does not open their blog/profile in a new tab.


sneak:

mcifani:
Updates in PR #1220 - Closing ticket.

PR #1220, opened 2017-03-01
#1158 add support for ctrl clicking on author name, misc eslint cleanup

mcifani:


mcifani:
Updates to handle ticket #1158 

- Ctrl clicking is enabled for author

app/components/elements/Author.jsx
```diff
@@ -11,32 +11,56 @@ import Userpic from 'app/components/elements/Userpic';
...
+    }
+
+    componentWillUnmount() {
+        const element = document.querySelector('.FoundationDropdownMenu__label');
+        if (element) element.removeEventListener("click", this.showProfileCtrl, false);
     }
-    shouldComponentUpdate = shouldComponentUpdate(this, 'Author')
+
+    showProfileCtrl(e) {
+        if (e.metaKey || e.ctrlKey) { // handle edge case for ctrl clicks
```
sneak: is this the right way to be doing this?  control-clicks and middle-clicks should probably be handled via the same code path versus special-casing them?

Issue #1167, opened 2017-02-16
clicking the user's userpic on the post view does not popup the user info popup like clicking on their username does


sneak:

mcifani:
Updates in PR https://github.com/steemit/condenser/pull/1223 - closing ticket.

PR #1223, opened 2017-03-03
#1167 add behavior for post full header userpic to be clickable for a…

mcifani:
…uthor preview, misc eslint cleanup


mcifani:
New features:

- Makes avatar/userpic in post full header clickable to reveal author preview snippet

app/components/cards/PostFull.jsx
```diff
@@ -134,6 +134,20 @@ class PostFull extends React.Component {
             this.state !== nextState
     }
 
+    componentDidMount() {
+        const headerPic = document.querySelector('.PostFull__header .Userpic');
+        if (headerPic) headerPic.addEventListener("click", this.showAuthorPreview, false);
+    }
```
valzav: The react's way to do it is to wrap Userpick with <a> and use onClick event:
<a href="#" onClick={}><Userpic/></a>

PR #1239, opened 2017-03-15
1119 conversion to steem-js

jcalfee:
This started out to be a websocket to http conversion but evolved into simply integrating steem-js and removal of duplicate code in the condenser repo.  Steem-js handles the connection to steemd so that conversion will be done in that repo later..


roadscape:
Waiting on https://github.com/steemit/steem-js/pull/118

PR #1252, opened 2017-03-21
increase comment depth to 255 [HF17]

roadscape:


sneak:
I was hoping to chat about this before we made this change in the frontend. I don't think this is the way to do it yet. Maybe 8 for now, but deep threading requires a special UI for the frontend to not totally suck.

roadscape:
It's a simple change for now, we'll never render more than a depth of 6 on one page (I can up it to 8 if needed). Beyond this, it links to the subtree. Just want to make sure we're not without a paddle for now.

roadscape:
Do NOT merge until HF17 is applied

sneak:
@roadscape in the future for stuff like this, please put it behind a has_hardfork() feature flag that checks the version so that it doesn't have to be synchronized with a backend version.

PR #1253, opened 2017-03-21
show delegated SP in wallet #1234 [HF17]

roadscape:


roadscape:
It's safe to release this as long as prod nodes are running 0.17

Issue #1254, opened 2017-03-22
Wallet Power Up does not handle commas


TimCliff:
I tried to power up 1,340.143 STEEM, and it ended up powering up 1 STEEM.

mcifani:
PR updates in https://github.com/steemit/condenser/pull/1268 to handle better, closing.

PR #1268, opened 2017-03-24
#1254 Improve match for not allow commas in transfer, not allow malfo…

mcifani:
…rmed amounts with more then one .

### DIFF
--- a/app/components/modules/Transfer.jsx
+++ b/app/components/modules/Transfer.jsx
@@ -22,7 +22,7 @@ class TransferForm extends Component {
     }
 
     constructor(props) {
-        super()
+        super();
         const {transferToSelf} = props;
         this.state = {advanced: !transferToSelf};
         this.initForm(props)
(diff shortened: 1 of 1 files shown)


app/components/modules/Transfer.jsx
```diff
@@ -75,9 +75,9 @@ class TransferForm extends Component {
                     ! values.to ? 'Required' : validate_account_name(values.to),
                 amount:
                     ! values.amount ? 'Required' :
-                    ! /^[0-9]*\.?[0-9]*/.test(values.amount) ? 'Amount is in the form 99999.999' :
+                    ! /^\d+(?:[\.\.]\d+)?$/.test(values.amount) ? 'Amount is in the form 99999.999' :
```
roadscape: This regex doesn't look right. Shouldn't it be `/^\d+(\.\d+)?$/` ?

mcifani: They actually work identical in the validation after testing, but prefer this as it's cleaner.  Updated.

PR #1270, opened 2017-03-27
change flag to downvote

roadscape:


roadscape:
https://cloud.githubusercontent.com/assets/5168676/24363712/9ee5750e-12de-11e7-9133-03224c712bab.png

mcifani:
This is good. Serves as a good alternative/lowers the perceived "flag" notion for users.

pfunks:
@sneak What other solution could there be to an issue that has caused non-stop user confusion? Some seem to be losing their minds over it: https://steemit.com/flagging/@krnel/i-love-flag-flag-loves-me-we-re-a-big-flagging-family

sneak:
@pfunks changing the icon and word without changing the mechanics isn't likely to help with that, unfortunately. we need a long-term solution, which this PR isn't.  for the reason I just described, I don't think this is a short-term fix, either.

pfunks:
I completely disagree. This will help with it, and it is a long term solution. What other solution is there besides removing downvoting from the UI altogether? People will still get downvoted then, by the way.

finleyexp:
Why not put a downvote icon next to the upvote icon AND make the flag a reporting-only function with a required drop down for flaggers to choose why they're flagging? In this way, the up/down vote system becomes more user-friendly, but the flagging option carries no STEEM / SBD consequence.

sneak:
I'm coming around to this idea, as we are seeing 0.1% flags from huge groups of bots.

I'm thinking maybe we just hide all non-100% flags from the UI, or we disable non-100% downvotes in a future hardfork.  Thoughts?

TimCliff:
Having less than 100% downvoting still seems like it would be useful, especially for large accounts that may be looking to do a measured reduction in rewards without taking the post/comment to zero/negative.

One suggestion that seems to have some support is to change the current upvote slider to have a negative/positive range (from -100% to +100%). For users below the current slider threshold it could be a boolean slider (down 100% or up 100%).

Along with this would be to remove any connotation of 'flag' and present it purely as a 'downvote'. The vote counts could just reflect the number of votes, ignoring positive vs. negative.

TimCliff:
An alternative UI implementation would be to just have two buttons - an upvote and downvote.

sneak:
> Having less than 100% downvoting still seems like it would be useful, especially for large accounts that may be looking to do a measured reduction in rewards without taking the post/comment to zero/negative.

There is no such thing as negative rewards.

If, in the voter's eyes, the post should not have zero rewards - they should not flag the post.  It is not a downvote.  I do not agree that "disagreement with rewards" is a legitimate reason for flagging.  A flag is "this shouldn't be here".

> One suggestion that seems to have some support is to change the current upvote slider to have a negative/positive range (from -100% to +100%). For users below the current slider threshold it could be a boolean slider (down 100% or up 100%).

Nah, the slider is terrible and I want to do away with it entirely.  Sliders are terrible skeuomorphism that are never a good solution to a problem.

> Along with this would be to remove any connotation of 'flag' and present it purely as a 'downvote'.

That's what this PR was, and that's why I rejected it.  Flags are not "simply downvotes", regardless of how the blockchain views them.

> The vote counts could just reflect the number of votes, ignoring positive vs. negative.

Yes, that is obviously the correct behavior.

TimCliff:
Is the plan then to try and solidify the idea of downvoting for abuse only? I.e. flagging? I'm not asking in an argumentative way, but if that is the direction things are headed it will require a cultural shift, as many users see that as a valid use of their stake - regardless of what it is called. There is also the question of how various forms of rewards pool 'abuse' should be handled. Downvoting is pretty much the only way to counter the various forms of siphoning from the rewards pool that were made possible/easier with linear rewards. Maybe 'flagging' still is a valid way to handle this, although it would only be reserved for extreme cases which would actually qualify as abuse.

pfunks:
>I'm thinking maybe we just hide all non-100% flags from the UI, 

First, they're downvotes, not flags. Second, what is the utility of that?

>or we disable non-100% downvotes in a future hardfork. Thoughts?

Not a reasonable change. Voting needs to be symmetrical, including voting scale. Making it unbalanced would skew the usefulness of concentrated SP used for abuse fighting and would necessitate splitting it between accounts which is not good. Less-than-100% downvotes are incredibly useful for fighting abuse.

liondani:
make a downvote button that by default downvotes -5% where the flag button by default downvotes -100%.... so keep both....

thrize:
One of the issues with the intention of trying to perform a down vote only in the current system is that it also effects rep. Conceptually this is a _flagging_ function.

I would propose that they are distinct functions on the blockchain level (so sorry that this is a little pan-topic, but needed), where:

* A down vote acts disagrees with rewards _only_, and in the UI is simple down vote, non threatening
* A flag is a down vote _and_ effects rep, and the UI is similar to now.

So the current rep logic for flagging/down voting will only be activated when is a flag.

Again, a little pan-topic, but for completeness I would simply add a boolean parameter to specify flag to the blockchain vote operation, such as ```is_flag```. Permissively allow the vote operation without this parameter for backwards compatibility where all non-updated code votes as down votes, the lesser evil.

---

On the conversation so far, I agree with @pfunks that voting needs to be symmetric and obvious to users. Down voting and flagging are not uncommon features on other platforms, there needs to be some conceptual alignment with what's out there already.

sneak:
I would do away wity flagging and downvoting *entirely* if it weren't a necessary mechanism for countering abuse.  Disagreement with rewards is best achieved not by downvoting, but by voting for *other things* instead.

Here's an option: hiding flags from the UI entirely until someone has a minimum rep or SP, and hiding the display of flags by anyone without a sufficient minimum rep/SP.

Upvoting is the action we want people to take; flags and downvotes should (ideally) be rare and only in the case of abuse.

goldibex:
@roadscape this PR is blocked on conflicts. Do you want to take a look at those or should I close this PR for now?

roadscape:
Let's table this for now.

roadscape:
Continued at #3307

Issue #1276, opened 2017-03-28
sorting (i.e. new/hot/trending) doesn't work as expected for tags/categories


Gandalf-the-Grey:
some people complain about new behavior on steemit.com  related to tags/categories
previously to find `new` posts in `#health`
one would click on the right on `#health` category
then on top on `new`

currently when clicking `#health` shows `trending` `#health` by default
but then when clicking on `new` we get all `new`, i.e. not `#health` `new`

(way to get `new` from `#health` is to click `new` and then `#health`)

roadscape:
Bug was introduced in https://github.com/steemit/condenser/pull/1267

PR #1280, opened 2017-03-29
fix: preserve tag on sort order switch

roadscape:


roadscape:
Bug reported in #1276

PR #1291, opened 2017-03-30
show claim_reward_balance ops in wallet

roadscape:


roadscape:
not tested. waiting to test against 0.18.2

Issue #1304, opened 2017-04-03
Show "STEEM POWER" instead of "VESTS" in claim transaction history on wallet pages


pfunks:
The STEEM Power portion of claims are currently denominated in VESTS. 

![image](https://cloud.githubusercontent.com/assets/22420130/24596969/78b112c4-1808-11e7-88cb-801f19fa9cba.png)

mcifani:
PR updates in https://github.com/steemit/condenser/pull/1305.

PR #1305, opened 2017-04-03
#1304 show steem power instead of vests in wallet history

mcifani:


mcifani:
- Display STEEM POWER indicator on claims
- Update consistency on SBD indicator

PR #1314, opened 2017-04-06
1070 i18n refactoring

pavvv:


sneak:
wow, this looks great! thank you for sending this in!

PR #1334, opened 2017-04-16
Welcome Page Updates

TimCliff:
- Updated the "Payouts" section for info on 7 day payouts (instead of 24h/30d)
- Removed @dantheman from people to follow
- Clarified live help
- Updated posts in the "Helpful Posts from Steemit Users" section


roadscape:
Thanks!

PR #1356, opened 2017-04-21
change wording on reward claiming alert box

sneak:
"claim" is a bad verb and implies that something happens to them if they're unclaimed.


svk31:
How about differentiating the two balance types, calling one "Your rewards balance", and the other "Your liquid balance"? Liquid may not be the best word, but you probably see what I mean. The 'claim' button could then simply be called 'Transfer to [liquid] balance'.

I think redeem invokes a feeling that the rewards can somehow be lost unless you actually redeem them, which is a problem the old text had as well.

sneak:
@svk31 that would be good too and i support a general rewording/rethinking of the ux. mine was just a quick first pass, feel free to continue to improve it.

PR #1358, opened 2017-04-21
Header dropdowns

svk31:
Adds a new 'Submit story' icon to the header for mobile:

image

Adds icons to the User profile dropdown menu:

image

Organizes the Hamburger dropdown menu in and changes the wording of some of the links.

image


TimCliff:
I like it! Would be good to change all the hamburger links to use consistent capitalization - "Quick Start Guide".

svk31:
Good point, will fix.

mcifani:
Good update with the icons.  The breakup in the hamburger menu is good albeit getting a big long.

svk31:
Yea I agree it's getting long, seems like we could remove some of those entries.

PR #1364, opened 2017-04-24
Add a modal showing QR codes for public/private keys in permissions tab

svk31:
A simple modal showing the public or private keys as a QR code so that mobile apps can scan it for easy migration. Triggered by clicking the QR image on the left of each key.

image


valzav:
It would be nice to implement import/login via QR code. E.g. I want to login on my phone using QR displayed in desktop browser.

svk31:
Oh you're right, I was just thinking  about this in the context of migrating to 3rd party apps but it could be useful for mobile users of Steemit too of course! I'll have a look and see if I can add that too.

Transisto:
Wow fast, hanks for that, What I originally had in mind was to approve an action by scanning a printout of my active key.  I guess that's coming next.

feruzm:
Thanks for adding this... Next version of eSteem will have support for QR reading of keys on password field ;)

PR #1371, opened 2017-04-26
Header fixes

svk31:
The main fixes are the vertical displacement of the header happening when the LoadingIndicator is active, and making the search page header look similar to the regular header. I've also made it so search results open inside the same window.

I've included a minor update to PostsIndex that I spotted while working on this since it doesn't really need its own branch anyway, just adds some translations and makes some links internal so they don't refresh the whole page for no reason.

### DIFF
--- a/app/components/modules/TopRightMenu.jsx
+++ b/app/components/modules/TopRightMenu.jsx
@@ -1,4 +1,3 @@
-import Immutable from 'immutable';
 import React from 'react';
 import { Link } from 'react-router';
 import {connect} from 'react-redux';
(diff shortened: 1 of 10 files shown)


app/components/modules/TopRightMenu.jsx
```diff
@@ -78,7 +77,7 @@ function TopRightMenu({username, showLogin, logout, loggedIn, vertical, navigate
         return (
             <ul className={mcn + mcl}>
                 {!vertical && <li className="Header__search"><a href="/static/search.html" title="Search"><Icon name="search" /></a></li>}
-                <li className={lcn}><LoadingIndicator type="circle" inline /></li>
+                <li className={lcn} style={{paddingTop: 0, paddingBottom: 0}}><LoadingIndicator type="circle" inline /></li>
```
valzav: Please try to avoid inline styling - we are planning to disallow it in CSP headers. If allowed, inline styles can be injection point for malicious code. Unfortunately we can't do it right now - inline styles are used by foundation library, so we need to either switch to new css framework or fix foundation.

sneak: we are using a css library with inline styles?!!? does the latest version of foundation still do that?!

valzav: we are using this library https://github.com/aruberto/react-foundation-components, seems to be it's stale in terms of development, I think we need to look for an alternative..

svk31: Inline styles are extremely common in React so it's gonna be hard to avoid completely.

PR #1393, opened 2017-05-01
misc waiting list fixes

valzav:


db/models/account.js
```diff
@@ -16,7 +16,8 @@ module.exports = function (sequelize, DataTypes) {
         referrer: DataTypes.STRING,
         refcode: DataTypes.STRING,
         remote_ip: DataTypes.STRING,
-        ignored: {type: DataTypes.BOOLEAN}
+        ignored: {type: DataTypes.BOOLEAN},
+        created: {type: DataTypes.BOOLEAN}
```
sneak: ps @valzav a good way of doing these is using a datetime column, and checking for NULL/non-NULL for the boolean... and then when it is created or ignored (for true), set to now().

PR #1398, opened 2017-05-02
Always save login (user's posting key) in local storage (close #1397)

### LINKED ISSUE #1397: Login triggered by upvote doesn't save user's posting key in localStorage

valzav:


valzav:
After reviewing the code I think saving/not saving login (posting key) from content page doesn't affect security - if user uses master password, it will be used to derive all keys and will stay in memory for a little while no matter wether login saved or not. Probably this precaution of not saving login was left from the time when we were saving active and owner keys.

PR #1427, opened 2017-05-17
add request timing logging to condenser

jredbeard:
This will log requests that take longer than 150ms to complete and includes the method and path of the request.

### DIFF
--- a/server/requesttimings.js
+++ b/server/requesttimings.js
@@ -1,0 +1,12 @@
+module.exports = requestTime;
+
+function requestTime() {
+	return function *requestTime(next) {
+    	let start = Date.now();
+    	yield* next;
+    	let delta = Math.ceil(Date.now() - start);
+    	// log all requests that take longer than 150ms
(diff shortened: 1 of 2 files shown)


server/requesttimings.js
```diff
@@ -0,0 +1,12 @@
+module.exports = requestTime;
+
+function requestTime() {
+	return function *requestTime(next) {
+    	let start = Date.now();
+    	yield* next;
```
sneak: ooh he fancy

server/requesttimings.js
```diff
@@ -0,0 +1,12 @@
...
+    	let delta = Math.ceil(Date.now() - start);
+    	// log all requests that take longer than 150ms
+    	if(delta > 150)
+    		console.log('Request took too long! ' + delta + 'ms: ' + this.request.method + ' ' + this.request.path);
```
sneak: please use structured logging when possible

PR #1432, opened 2017-05-18
show green check mark next to Sign Up in the top menu if user is approved to sign up

valzav:
image


valzav:
The problem I'm trying to solve with this is low conversation rate after approvals. I think this is mostly related to sign up approval emails are lost or get into spam folders. Having a visual sign up on steemit.com that Sign Up is approved can help with this issue.

Issue #1337, opened 2017-04-17
Feature request : Show vote power % in dropdowns (Already on golos)


ausbitbank:
On golos.io post views, on the voters dropdown at the bottom of the full post view we're able to see the vote power % that each voter gave on each post. I'd love to see this feature enabled on Steemit too!

![image](https://cloud.githubusercontent.com/assets/1692203/25087727/5ed6b0ea-23b4-11e7-8d46-c0557fc2f48b.png)

I posted about this here and it seems I'm not the only person interested in this feature :
https://steemit.com/steemit/@ausbitbank/a-golos-feature-i-d-love-to-see-on-steemit-vote-in-dropdowns

I wasn't actually able to find the golos repository on github however, I'd be happy to dig through, find the relevant code and submit a patch if someone can point me in the right direction :)

ausbitbank:
The relevant code for this seems to be in /app/components/elements/Voting.jsx of the golos/tolstoy repo
:         
```
for( let v = 0; v < avotes.length; ++v ) {
            const pct = avotes[v].percent
            const cnt = Math.sign(pct)
            if(cnt === 0) continue
            count += 1
            if (showList && voters.length < MAX_VOTES_DISPLAY) {
                const voterPercent= pct / 100 + '%';
                voters.push({value: (cnt > 0 ? '+ ' : '- ') + avotes[v].voter, link: '/@' + avotes[v].voter, data: voterPercent});
```

I don't have a way to test this right now, but hopefully that helps save someone some time :)
The equivalent codeblock in the condensor repo is in the same file, lines 222-230 - basically just adding  "pct" and "voterPercent" constants :)

ausbitbank:
I submitted a pull request to add this feature, and [my post about it](https://steemit.com/steemdev/@ausbitbank/steemit-pull-request-submitted-vote-in-voters-dropdown) got heaps of support .. 
Can somebody please review this ? @sneak

sneak:
I don't think this adds value to most users, at the expense of cluttering the interface with confusing figures that are not explained in any way.

We already truncate the list, requiring someone use a block explorer to view the full list. Vote strengths fall in that same category.

PR #1433, opened 2017-05-19
Add vote strength % to voter dropdown

ausbitbank:
As per issue https://github.com/steemit/condenser/issues/1337 (!), I've made the required changes to display vote % in the voter dropdown menu and verified they work on my own local fork of steemit with no errors.

image


sneak:
Thank you for the PR, but the clutter added by this change is not outweighed by utility to most users of steemit.

PR #1450, opened 2017-05-31
show cents if sign up bonus is less than $1 (close #1449)

### LINKED ISSUE #1449: Shows "Get $0 of STEEM POWER when you sign up today." to new users
img_0141
 
(Reported by @snowflake)

valzav:


valzav:
This was supposed to be a quick fix, I hoped to deploy it yesterday, right now it shows zero. IMO this whole banner needs to be replaced with something more compelling.

mcifani:
Updated without amount, good to go.

torvali:
Here's the new copy I wrote up:

Create, curate, and get rewarded for great content
Be an early adopter in the new attention economy
Blue Button Copy: "Sign me up!"
White Button Copy: "What is Steem?"
Bottom header: 100% peer to peer, 100% censorship resistant, 100% free

PR #1464, opened 2017-06-07
Critical fix for a possibility of making own "password" or "private key" public by exposing it as a memo

noisy:
We want to prevent such things:

selection_999 152

### FIX:

This fix checks memo against public keys of a user, and prevents a mistake:

selection_999 148

selection_999 150

### DIFF
--- a/app/components/all.scss
+++ b/app/components/all.scss
@@ -30,6 +30,7 @@
 @import "./elements/ShareMenu";
 @import "./elements/Author";
 @import "./elements/UserNames";
+@import "./elements/UserKeys";
 @import "./elements/QrKeyView";
 
 // modules
--- a/app/components/elements/UserKeys.jsx
+++ b/app/components/elements/UserKeys.jsx
@@ -91,7 +91,8 @@ class UserKeys extends Component {
             </div>
             <div style={{paddingBottom: 10}} className="column small-12">
                 <Keys account={account} authType="memo" onKey={onKey.Memo} />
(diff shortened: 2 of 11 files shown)


sneak:
Thanks for sending this in!  We immediately undertook the same work as soon as we got the initial report of this happening, but it looks like you're faster. :)

I need to check and make sure that the `.isWif` function will detect a WIF format even if whitespace padded (e.g. `5.... ` or ` 5....`) and the password check is too narrow - it should prevent anything that looks like a password, even if it's not the password for that specific account.

Do you want to make these changes, or should we?  Again, thanks for the quick patch on this!

noisy:
> Do you want to make these changes, or should we? Again, thanks for the quick patch on this!

I think I can try to do that.

> password check is too narrow - it should prevent anything that looks like a password, even if it's not the password for that specific account.

If this should be a matter of checking a pattern of `/^5[HJK].{45,}/i` (I will take a look into source whether this can be more detail check) - than this will be no problem. But technically everything can be a password if someone set a password before last hack, or set own password using console - so I believe, we talking only about checking a pattern?

sneak:
Turns out @mcifani already had this mostly done on his local version and it's merged ( #1467 ) and staged now - should be rolled out to prod within a couple hours. :)

Great minds think alike - thank you for jumping in on this!

sneak:
PS: https://steemitstage.com (not HA, not to be relied upon - but you can see the fix live)

PR #1476, opened 2017-06-12
Add account to the bad actor list

bacchist:
https://steemit.com/steemit/@randowhale/do-not-submit-steem-sbd-to-randomwhale-it-s-randowhale-no-m


noganoo:
I am unable to change the memo for RandomWhale account to shut down voting service.  Unable to login via GUI.  What is the solution to this?

PR #1478, opened 2017-06-13
updated files in condenser to help dev initial setup run immediately …

surfyogi:
…from README instructions. I assume it's easy to change them once a person knows what they are doing.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -1,5 +1,5 @@
 {
-  "database_url": "sqlite:///tmp/sqlite.db",
+  "database_url": "mysql://root:password@127.0.0.1/steemit_dev",
   "disable_signups": false,
   "facebook_app_id": false,
   "grant": {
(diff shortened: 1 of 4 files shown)


mcifani:
@surfyogi Looks good, and updates to the default db. I believe the closer we get any devs running their env locally to production settings is always best.

config/default.json
```diff
@@ -1,5 +1,5 @@
 {
-  "database_url": "sqlite:///tmp/sqlite.db",
+  "database_url": "mysql://root:password@127.0.0.1/steemit_dev",
```
roadscape: @sneak keep sqlite default or no?

sneak: Up to you guys - I am in favor of allowing condenser to continue to work on a local dev instance very simply, but I know with the signup changes we've made as well as the forthcoming dependency on `steemit/hivemind` that that might be unrealistic.  Your call.

PR #1498, opened 2017-06-20
Update README.md

TimCliff:
I'm proposing a few changes to the README.md file to make the setup cleaner.

1. Some of the steps that were recently added were also checked in as changes to the config files in master, so it is no longer necessary for someone going through the install steps to make the edits.
2. I added some `code` formatting to make it clearer when the instructions were commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ mysql -u root
 
 Create a password for the root mysql account:
 
-mysql_secure_installation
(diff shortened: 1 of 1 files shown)


TimCliff:
The config.json and default.json changes may be good to leave in, but with instructions on what to do if the DB password is different than "password". I will revise and submit an updated PR.

PR #1508, opened 2017-06-20
adds basic clustering (multi-core child process) support for condenser

jredbeard:
this gives a decent performance boost when under load on multi-core systems


valzav:
@jredbeard, could you make it optional or only activate in production?
Also I would reduce it to 1x cores - this way it still utilizes all cores but has 2x less memory requirement.

jredbeard:
@valzav updated - now clustering only happens in production mode and number of workers defaults to number of cores (but is configurable by env var, if we desire). I also changed the node `max-old-space-size` from 4GB to 1.75GB - that should be sufficient based on benchmarking I've done and will allow for running an optimal number of these on most instance sizes we would consider.

I also added a bit for restarting dead workers - I tested this by killing a workers PID while benchmarking and it kept serving while restarting the worker.

PR #1511, opened 2017-06-20
Update Vote Slider Threshold

TimCliff:
This pull request updates the vote slider threshold from 1.0 MV to 0.25 MV. This is 1/4 the SP needed to use the slider, since 100% votes are now 4x as strong.

I did tests with a user that had 120 SP (a little less that .25 MV), and they were able to upvote some posts with 1%, while near full voting power. After they have depleted some of their voting power they will start to get a _"Voting weight is too small, please accumulate more voting power or steem power."_ error. They can either increase their voting weight to 2% and continue voting, or buy more SP. They can continue with 2% votes until they get to around 50% voting power and then they will need to up to 3%.

[Edit] Link to Steemit discussion post: https://steemit.com/voting/@timcliff/condenser-pull-request-submitted-minnow-vote-slider-vote-slider-threshold-update-to-125-sp

### DIFF
--- a/app/components/elements/Voting.jsx
+++ b/app/components/elements/Voting.jsx
@@ -23,7 +23,7 @@ const ABOUT_FLAG = <div>
 </div>;
 
 const MAX_VOTES_DISPLAY = 20;
-const VOTE_WEIGHT_DROPDOWN_THRESHOLD = 1.0 * 1000.0 * 1000.0;
+const VOTE_WEIGHT_DROPDOWN_THRESHOLD = 0.25 * 1000.0 * 1000.0;
 
 class Voting extends React.Component {
 


valzav:
@roadscape, do you think the threshold should be decreased?

roadscape:
Does the added complexity provide a net benefit? I'd be curious to know how many active users this would affect.

roadscape:
I consider it more of a power user feature. What if there was just some console cmd or keybd shortcut that enabled it?

TimCliff:
There are quite a few 'minnow' users that have been saying they want it. Users have even been delegating enough SP to others to let them use the slider. I think there is demand for having it.

The current PR is actually neutral as far as the amount of voting power, compared to pre-HF19. Since votes are 4x as strong, the PR makes the threshold 1/4 as much.

A setting to be able to enable and disable the slider, or a keyboard command to pull it up could be a good enhancement. Allowing users with less than 0.25 MV to use it though opens up an additional set of complexity though, because at that point users cannot fully use the slider, since a 1% vote would be below the bandwidth limit.

roadscape:
Of 244,705 accounts:

 - 6,099 have effective voting weight > 1MV and thus have a slider
 - 12,983 are between 0.25 - 1MV (and would get a slider if this PR is merged)

valzav:
I agree with @roadscape - it's more a power user feature, let's do it via the settings

lumian007:
Actually there was none who could delegate some SP, so i need to use the voting slider at 0.25MV.
How could i get it in my account @thirdeye7

TimCliff:
Sorry, GitHub is not the place to ask this question. Try the help channel in Steemit chat.

PR #1559, opened 2017-07-13
adds ability to configure steem-js with http/jsonrpc via env vars

jredbeard:


app/Main.js
```diff
@@ -25,7 +25,10 @@ try {
 function runApp(initial_state) {
     console.log('Initial state', initial_state);
     const config = initial_state.offchain.config
-    steem.config.set('websocket', config.ws_connection_client);
+    if(process.env.USE_JSONRPC)
```
sneak: this is using server env vars but this is the http_connection_client - the URI for the SPA to talk to, no?  Am I confused?

jredbeard: This is duplicating condensers existing functionality with websockets but for http/jsonrpc. As it exists now, condenser makes use of two instances of steem-js's - one for the SPA/client to make (ws_connection_client/http_connection_client) and one for condenser itself. However, these two variables are almost always going to be set to the same URI (as was the case with websockets as well). The USE_JSONRPC env var is used in both places... so either ws_connection_client and ws_connection_server would be set for websockets (as it's always been), or http_connection_client and http_connection_server would be used alternatively if USE_JSONRPC is set. This makes it easy to switch between the two functionalities if necessary (in case a problem is later uncovered with http steem-js for example) and it won't break deployments before the env vars are set/changed.

sneak: yes but the client won't have process.env - maybe that just means this always falls through in the SPA?

jredbeard: I see what you mean @sneak  - I'm not 100% on that, but I'll adjust and have it pull that from the config object instead of process.env - that should be sufficient

jredbeard:
@sneak changes made and tested

jredbeard:
@sneak requested changes were made, please re-review

sneak:
utACK - please test thoroughly

package.json
```diff
@@ -126,7 +126,7 @@
     "sequelize-cli": "^2.3.1",
     "speakingurl": "^9.0.0",
     "sqlite3": "^3.1.8",
-    "steem": "0.5.15",
+    "steem": "github:steemit/steem-js#78d95ec",
```
sneak: 👌

config/default.json
```diff
@@ -95,8 +95,8 @@
     "auth_token": false
   },
   "upload_image": false,
-  "ws_connection_client": "wss://steemd.steemit.com",
-  "ws_connection_server": "wss://steemd.steemit.com",
+  "steemd_connection_client": "wss://steemd.steemit.com",
+  "steemd_connection_server": "wss://steemd.steemit.com",
```
sneak: should these read `api.steemit.com` now that we are moving to http support?

jredbeard: Eventually, yes. I think the default should still be what it is currently for now... we'll still be using websockets to steemd and not jussi/api.steemit until we do a thorough public test on staging. Merging this in though will allow us to function either with websockets or http/jsonrpc, and will ensure that other branches being tested on dev/staging will already be using jsonrpc as well. Then it will just be an env var change in production when we actually switch it over (and we can then update the defaults to api.steemit.com in a new branch).

PR #1563, opened 2017-07-14
Allow condenser to run against different chains

jnordberg:
This patch adds config options for setting a custom chain id and address prefix. I also had to update the browser tests for it to run.

### DIFF
--- a/app/Main.js
+++ b/app/Main.js
@@ -26,6 +26,12 @@ function runApp(initial_state) {
     console.log('Initial state', initial_state);
     const config = initial_state.offchain.config
     steem.config.set('websocket', config.ws_connection_client);
+    if (config.custom_addr_prefix) {
+        steem.config.set('address_prefix', config.custom_addr_prefix);
+    }
+    if (config.custom_chain_id) {
+        steem.config.set('chain_id', config.custom_chain_id);
+    }
     window.$STM_Config = config;
     plugins(config);
     if (initial_state.offchain.serverBusy) {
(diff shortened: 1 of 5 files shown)


app/Main.js
```diff
@@ -26,6 +26,12 @@ function runApp(initial_state) {
     console.log('Initial state', initial_state);
     const config = initial_state.offchain.config
     steem.config.set('websocket', config.ws_connection_client);
+    if (config.custom_addr_prefix) {
+        steem.config.set('address_prefix', config.custom_addr_prefix);
```
sneak: i would recommend setting address_prefix to `STM` in defaults.json, and dropping this logic and the use of the `custom_` prefixes.  Then to use a 'custom' one you just change address_prefix.

jnordberg:
Done

roadscape:
Why remove the memo encryption test?

jnordberg:
@roadscape When the memo test runs a very similar test was run internally in steem-js, with a hardcoded `STM` prefix causing the test case to fail.

As I understood the code, the intention of the test was just to check availability of memo encryption, not throw and stop people from logging in. But since the internal steem-js test runs and throws failing the test case it defeated that purpose, so I figured it was better to remove since it never worked in the first place

roadscape:
@valzav is it safe to remove this test? I'm not sure of the status of memo encryption myself.

valzav:
@roadscape, I think it's safe to remove memo encryption test - memo encryption is not used in condenser anyway

PR #1567, opened 2017-07-17
Move source files into src/ dir and compile them into lib/ for production; Run server with nodejs rather than using babel-node

valzav:


package.json
```diff
@@ -125,7 +126,7 @@
     "sequelize-cli": "^2.3.1",
     "speakingurl": "^9.0.0",
     "sqlite3": "^3.1.8",
-    "steem": "0.5.15",
+    "steem": "0.5.19",
```
sneak: what sort of code review do we have in place on steem-js?

roadscape:
Testing of this is in `rc` branch

PR #1577, opened 2017-07-20
DMCA Takedown Notice received

firepowernaik:
Updated list (rows 2-18)


thecryptofiend:
I'm disappointed that you would respond to a DMCA takedown for my book review.  It is obvious fair use.

PR #1584, opened 2017-07-24
766 - Adding slider for power down

TheIntrnet:
I made the slider its own component, and moved it to a dialog. I also set it up to be translated, should just need translations added for other languages. 

screen shot 2017-07-23 at 11 30 20 pm

screen shot 2017-07-23 at 11 30 29 pm


roadscape:
```
ERROR in ./app/components/modules/Powerdown.jsx
Module not found: Error: Cannot resolve module 'counterpart' in app/components/modules
 @ ./app/components/modules/Powerdown.jsx 57:19-41
```

PR #1589, opened 2017-07-25
766 slider to control withdraw vesting amount

TheIntrnet:
I made the slider its own component, and moved it to a dialog. I also set it up to be translated, should just need translations added for other languages. 

screen shot 2017-07-23 at 11 30 20 pm

screen shot 2017-07-23 at 11 30 29 pm


sneak:
getting warmer. here's the todo:

1) it needs an editable text box with the value so that blind users using screen readers can tab into it and change it using the keyboard to an exact amount.

2) the slider is the wrong color, it should match current steemit blue. it should inherit that from some master component so that when the color is changed in condenser globally it updates all components that use it.

3) the margins around the slider are too small.  give it breathing room. :)

sneak:
If you notice, the slider handle is way too close to the close x as well as way too close to the "power down" default button. Imagine trying to click this on a tiny netbook with a stylus. Make the handles larger and put a much larger (200-300%) inactive whitespace margin around the slider to move every active clickable part apart.  There's plenty of space on the screen that's not being used so feel free to make the modal twice as large to accommodate these safety borders around the click targets.

TheIntrnet:
Updates:

screen shot 2017-08-02 at 9 17 23 pm

screen shot 2017-08-02 at 9 17 33 pm

TheIntrnet:
screen shot 2017-08-04 at 4 17 26 pm

sneak:
I want 50-80 pixels on all sides of the slider.  It's too close to the sides of the dialog.  

The text is too small.  Also, please center the label+input box combo, and include a unit (it should be `Steem Power` or `VESTING_TOKEN` in condenser-speak).

Please use the screen space.

app/locales/en.json
```diff
@@ -809,10 +809,10 @@
   "powerdown_jsx": {
     "power_down": "Power Down",
     "power_down_amount": "Power Down Amount",
-    "select_powerdown_amt": "Please select and amount of %(VESTING_TOKEN) to power down.",
+    "select_powerdown_amt_VESTING_TOKEN": "Please select and amount of %(VESTING_TOKEN)s to power down.",
```
sneak: *an amount

app/locales/en.json
```diff
@@ -809,10 +809,10 @@
...
     "confirm": "Confirm Power Down",
-    "steem_power": "Steem Power",
+    "steem_power": "%(VESTING_TOKEN)s",
     "confirm_power_down": "Are you sure you wish to power down of your Steem Power?",
```
sneak: *power down your Steem Power

app/locales/en.json
sneak: why did the token name change from `select_powerdown_amt` to `select_powerdown_amt_VESTING_TOKEN`?

roadscape:
At some point the UI might need to show how much SP is locked due to delegation.

Currently, if you've delegated any amount you will not be able to power down from steemit.com. The variable slider will help, but it would be nice if it would recognize how much is locked.

jredbeard:
@roadscape at some point yes - but until we support delegation in the interface I think that adding that feature would be over-engineering this and would delay getting it shipped

sneak:
Yeah the slider should only let you power down your undelegated portion. That's not over-engineering, that's the minimum viable UX.

roadscape:
Where are we at with this and can we merge #1656 in the meantime?

jnordberg:
#1669 needs reviewing and possibly localisation. It also fixes #1656

PR #1593, opened 2017-07-28
[WIP] fix, remove dependency on FoundationDropdownMenu from UserProfile Page

gl2748:
This is the guts of the work and is ready for review.

### Known Todos

- [ ] The condition to show/hide the logged-in user's drop down is not working for a split second when you navigate back to steem, as a logged in user, having left. See upper right hand corner in this screen-shot.

error_bug

- [ ] The positioning of the new Overlay does not precisely match the older position, the `FoundationDropdownMenu` has a 'pull-left' type prop that I am working on addressing in the CSS for the updated implementation. For example...
#### Current.

steemfore

#### After changes introduced.
steemfix2

### DIFF
--- a/app/components/modules/TopRightMenu.jsx
+++ b/app/components/modules/TopRightMenu.jsx
@@ -1,40 +1,66 @@
 import React from 'react';
 import { Link } from 'react-router';
 import {connect} from 'react-redux';
+import { findDOMNode } from 'react-dom';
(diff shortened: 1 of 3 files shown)


gl2748:
Closing, no longer relevant. Right now the `FoundationDropdownMenu` component is only used in:
`UserWallet`, and `Post`.

PR #1599, opened 2017-08-01
Added Native Steem Search

KylePiira:
Added the ability to query the steem blockchain directly through the free [AskSteem API](https://steemit.com/asksteem/@thekyle/asksteem-search-api-docs). Allows for making more advanced queries like by author, tag, and others listed [here](https://steemit.com/steemit/@thekyle/introducing-asksteem-a-steem-search-engine) that are currently not supported by the existing Google CSE integration.

### DIFF
--- a/app/assets/static/search.html
+++ b/app/assets/static/search.html
@@ -1,41 +1,58 @@
 <html lang="en">
+
 <head>
     <meta charSet="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/ico" href="/favicon.ico" />
     <title>Search - steemit.com</title>
+    <!-- CSS Frameworks -->
     <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/foundation/6.2.1/foundation-flex.css">
+    <link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
+    <!-- Fonts -->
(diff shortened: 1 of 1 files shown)


sneak:
I appreciate the work you've done here, but we have some blockchain indexing services that I think will be a much better fit for something like this.

We also have a general prohibition on loading 3p resources (despite the cdnjs cloudflare link in this file - it should not have been there in the first place) so calls out to an external API are not something for which we are going to poke a hole in our CSP.

I'm thinking that `sbds` is the place we want to implement search.  Do you hack python?  Would you be interested in putting together the serverside component there?

KylePiira:
Hi @sneak thank you for your consideration, I'm not sure what you mean by "hack python" but I am very well versed in the Python programming language. What did you have in mind specifically for the server-side component?

sneak:
I'm thinking a jsonrpc call in `sbds` (https://github.com/steemit/sbds) that does full text search.

You could do a simple mysql full text search over all of the post/edit rows, and fetch permalinks, then fetch complete edit history for each permalink that hits and apply it locally, and then filter the list of permalink/content pairs with the search terms.

sneak:
https://github.com/steemit/sbds/blob/master/sbds/server/serve.py#L117

https://github.com/steemit/sbds/blob/master/sbds/server/methods.py#L64

If you add a method here that supports full text search (just using the db - the actual text content of the blockchain isn't that big yet) using the db of sbds, then we would happily merge in a condenser PR that queries it.  This is the point of `sbds` and other indexers.

sneak:
(the algorithm i described above would not be perfect, but would be an excellent first pass at getting search running without involving 3p services.)

sneak:
@john-g-g and @roadscape might have better ideas about how to do the requisite post-history-stacking required for an accurate fulltext search call.

Issue #1595, opened 2017-07-29
Layer scrolling bug with the article popover modal on mobile.


gl2748:
### From Slack.
>  "...there is a layer scrolling bug with the article popover modal on mobile when opening an article from a list view...where sometimes it scrolls the mostly hidden list instead of the article ... maddening ... when the article modal is over the list view, sometimes scrolling will scroll the list view and not the article and the article will repaint over it enough times per second that you can see that you're scrolling the list and not the article"

gl2748:
I am working on this over the weekend.

gl2748:
![steem_bug](https://user-images.githubusercontent.com/3374538/28747291-64ff0250-7469-11e7-9fc1-fe5018c9b7c1.gif)

plink01001:
Is this still an issue, or can it be closed?

plink01001:
With the removal of modals, this should be resolved. Please re-open if it is still an issue.

Issue #1386, opened 2017-04-30
Scrolling on mobile has a buggy header nav


jredbeard:
When scrolling in article view on mobile, the header nav is buggy - it constantly repaints.

![image uploaded from ios 1](https://cloud.githubusercontent.com/assets/23157604/25565060/861d9c16-2d8e-11e7-9578-7d70df47af15.jpg)

valzav:
It's a browser's bug/feature with fixed positioning, there is not much we can do about it.

PR #1615, opened 2017-08-03
1386 1595 post scrolling bugs

gl2748:
# Refs
#1386 
#1595

# Mobile - iOS Safari.
steem_scrolling_2

### DIFF
--- a/app/components/cards/PostsList.jsx
+++ b/app/components/cards/PostsList.jsx
@@ -9,7 +9,6 @@ import Icon from 'app/components/elements/Icon';
 import shouldComponentUpdate from 'app/utils/shouldComponentUpdate';
 import {connect} from 'react-redux';
 import { translate } from 'app/Translator.js';
-
 import Modal from 'react-overlays/lib/Modal';
 import Portal from 'react-overlays/lib/Portal';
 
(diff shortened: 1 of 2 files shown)


sneak:
This is exciting.  @valzav pointed out that this is a much better solution (using well-maintained community libraries, vs rolling our own) than we had.  Will be happy to get this merged in.

gl2748:
@sneak @valzav 
Agreed, there is one issue with this implementation however.

When you press 'Back' on mobile, the scroll position of the underlying PostList is not restored. I've been trying to wrap my head around a fix.

1. the react-bootstrap [Modal](https://react-bootstrap.github.io/react-overlays/#modals) component is working as designed in that the focus is restored to the correct element on the PostList. If you use the 'esc' key the modal is closed and the correct position is also restored, I think  this is a function of the browser however rather than the Modal component.

2. The question is how to get a fast scroll to happen to this element ... after a user presses 'Back'? Open to suggestions. This functionality is currently working on the current implementation, if anyone knows how that works, it shouldn't be too hard to re-implement here.

sneak:
FYI this also affects the close X in the top right of the post modal when viewing a post over a list view.

valzav:
@gl2748, I would suggest to comment out useScroll in UniversalRender.jsx and see whether issue #1 goes away or not

Issue #1609, opened 2017-08-03
npm install not working on fresh install - Ubuntu 16.04


TimCliff:
I am running a fresh install on Ubuntu 16.04 using the instructions in the condenser GitHub "Code" section. When I run `npm install`, it gives an error.

```...

.....

Running post install.js

> steem@0.6.1 prepublish /home/condenser10/condenser/node_modules/steem
> npm run build


> steem@0.6.1 build /home/condenser10/condenser/node_modules/steem
> npm run build-browser && npm run build-node


> steem@0.6.1 build-browser /home/condenser10/condenser/node_modules/steem
> rm -rf dist && NODE_ENV=production webpack && gzip -k -f ./dist/*.js && du -h                                                                                         ./dist/*

Hash: a98b1c86d0e44ce98a4d
Version: webpack 1.15.0
Time: 53ms
    + 1 hidden modules

ERROR in Entry module not found: Error: Cannot resolve 'file' or 'directory' /ho                                                                                        me/condenser10/condenser/node_modules/steem/src/browser.js in /home/condenser10/                                                                                        condenser/node_modules/steem

ERROR in ./test/api.test.js
Module parse failed: /home/condenser10/condenser/node_modules/steem/test/api.tes                                                                                        t.js Unexpected token (18:27)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected token (18:27)
    at Parser.pp$4.raise (/home/condenser10/condenser/node_modules/steem/node_mo                                                                                        dules/webpack/node_modules/acorn/dist/acorn.js:2221:15)
    at Parser.pp.unexpected (/home/condenser10/condenser/node_modules/steem/node                                                                                        _modules/webpack/node_modules/acorn/dist/acorn.js:603:10)
    at Parser.pp.expect (/home/condenser10/condenser/node_modules/steem/node_mod                                                                                        ules/webpack/node_modules/acorn/dist/acorn.js:597:28)
    at Parser.pp$3.parseExprList (/home/condenser10/condenser/node_modules/steem                                                                                        /node_modules/webpack/node_modules/acorn/dist/acorn.js:2152:16)
    at Parser.pp$3.parseSubscripts (/home/condenser10/condenser/node_modules/ste                                                                                        em/node_modules/webpack/node_modules/acorn/dist/acorn.js:1741:35)
[...]

PR #1629, opened 2017-08-05
Update README install instructions to use yarn instead of npm

TimCliff:
The current README instructions to install/run a local instance of condenser do not work. As per the info in #1609, condenser now uses yarn instead of npm. This PR updates the README file to work with yarn. It also makes several organization and clarification changes, and updates a few out of date commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # Steemit.com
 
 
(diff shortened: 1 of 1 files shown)


valzav:
@jredbeard has already addressed `yarn` in this pr https://github.com/steemit/condenser/pull/1631

jredbeard:
Thanks for the submission @TimCliff. I think you'll find that the new readme has a lot of useful new information.

TimCliff:
Thanks - I'll check it out.

PR #1631, opened 2017-08-06
Update condenser README

jredbeard:
Condenser's documentation has become heavily updated. This should bring it up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,42 @@
 
-# Steemit.com
+# Condenser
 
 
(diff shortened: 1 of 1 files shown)


sneak:
@jredbeard suggesting people change settings in defaults.json should be replaced by suggesting they change settings in production.json or dev.json in the same dir.

sneak:
for more info see npm docs on 'config' module

jredbeard:
fixed in #1636 instead, freshly branched off master

PR #1634, opened 2017-08-06
add ability to disable cps violation reporting

valzav:


sneak:
LGTM but I would have added a note in the documentation before using a special character `-` to mean "don't report".

Issue #1596, opened 2017-07-30
Flagged posts showing negative and incorect number of votes


GoldenBoar:
I've noticed this on 2 posts now, the number of votes is showing as negative, but when you click the link to read the post it shows a different positive number of votes. Both posts I noticed like this had been flagged.

Here are some screenshots to highlight what I'm talking about:

http://i.imgur.com/qFvHHtn.png

and

http://i.imgur.com/QuvsDLE.png

That particular post can be found [here](https://steemit.com/freedom/@adsactly/a-tale-of-two-futures).

sneak:
It's adding the upvotes and subtracting the flags.  It should show total vote count.

relativityboy:
Fixed by @gl2748

PR #1640, opened 2017-08-08
Fix, display total votes in post summary, refs #1596

gl2748:
Refs #1596

### Before.
screenshot from 2017-08-08 08-14-18

### After.
screenshot from 2017-08-08 08-59-00

This is a naive fix, that uses the existing vote-count from the `Voting` component rather than the `VotesAndComments` component. It might be worth considering consolidating these two components.

Anyway I was wondering if anyone @valzav, @roadscape @TheIntrnet could tell me where [`net_votes`](https://github.com/steemit/condenser/blob/master/src/app/components/elements/VotesAndComments.jsx#L51) is coming from? As setting this up correctly would perhaps be the ideal solution.


gl2748:
@sneak, does the style change with this approach pass muster? See the b4 and after screen caps above ...

sneak:
yeah the vote count in the post list view should not have a dropdown / menu reveal chevron. it should not be clickable at all.

gl2748:
@valzav implemented solution as you suggested.

valzav:
@gl2748, can I ask you to make one more change: please instead of post_obj pass total_votes into the component - it will let shouldComponentUpdate to detect real changes in props and not re-render if something has changed in post obj not related to votes count.

roadscape:
@gl2748 Please rebase, let's test and merge asap.

PR #1642, opened 2017-08-09
Added portuguese tag in Faq

elieltavares:
We came to the conclusion of using the tag 'portugues' to gather all speakers of the Portuguese language

### DIFF
--- a/src/app/help/en/faq.md
+++ b/src/app/help/en/faq.md
@@ -622,7 +622,7 @@ Language-specific groups include:
 - Korean = kr
 - Russian = ru
 - French = fr
-- Portuguese = portuguese
+- Portuguese = portugues
 
 <a href="/faq.html#Table_of_Contents_Posting">^</a>
 <a class="anchor" name="Can_I_delete_something_I_posted"></a>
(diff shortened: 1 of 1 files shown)


originated:
Sounds good to ad, I'm a but curious is it good we have some inconsistencies here and a few have the full wording? 

Perhaps we should use the standard global ISO code for all, example:

Chinese = cn
German = deutsch (should be de)
Spanish = spanish (should be es)
Korean = kr
Russian = ru
French = fr
Portuguese = portugues (should be pp)

bycz6:
Portuguese should be "pt" instead of pp

originated:
@bycz6 Aw yep correct, updated the list.

TimCliff:
For new communities that are just forming, the standard global ISO code seems the obvious way to go. For the existing language communities that are already formed around a tag though (example deutsch) someone from the community will need to work on getting the members of the community to transition to the new tag.

elieltavares:
I agree to use the 'pt' tag, because the Portuguese community is still small, because for now it is possible to change

originated:
One concept we could consider to foster our global growth is have a legend/or easy country code picker. Either directly in the text field as they type with suggestions, more cumbersome, or something more simple under the textbox that they click with the all the codes listed(which could auto populate their tag). This could elimate user error and confusion

PR #1650, opened 2017-08-11
use telesign's phone number cleansing; only check for email and phone…

valzav:
… used after they are confirmed

### DIFF
--- a/src/server/sign_up_pages/enter_confirm_email.jsx
+++ b/src/server/sign_up_pages/enter_confirm_email.jsx
@@ -65,8 +65,11 @@ function* confirmEmailHandler() {
         return;
     }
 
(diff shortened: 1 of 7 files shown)


src/server/sign_up_pages/enter_confirm_email.jsx
```diff
@@ -64,6 +64,21 @@ function* confirmEmailHandler() {
         this.body = '<!DOCTYPE html>Confirmation code expired.  Please <a href="/enter_email">re-submit</a> your email for verification.';
         return;
     }
+
+    const number_of_created_accounts = yield models.sequelize.query(`select count(*) as result from identities i join accounts a on a.user_id=i.user_id where i.provider='email' and i.email='${eid.email}' and a.created=1 and a.ignored<>1`);
```
sneak: is this vuln to sql injection?

sneak: you should be using parameterized queries, never string interpolation.

valzav:
@sneak, please review this PR

Issue #1651, opened 2017-08-11
cover image support


roadscape:
This issue tracks the user profile cover image feature, as per @sneak

PR #1652, opened 2017-08-11
add cover image support #1651

roadscape:
image


valzav:
I've tested it - works great for me!

PR #1656, opened 2017-08-12
Fix powering down when having delagated SP

HappyMaster:
Because it [broadcasts the total amount of vests](https://github.com/steemit/condenser/blob/107f116029b45bf651f46296586cf30d9a3bb785/src/app/components/modules/UserWallet.jsx#L98), when having delegated SP, [it throws](https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L895).

Here I wrote a detailed post about this issue: https://steemit.com/steemit/@happy.man/fixing-powering-down-and-how-to-do-partial-powering-down-on-steemit

### DIFF
--- a/src/app/components/modules/UserWallet.jsx
+++ b/src/app/components/modules/UserWallet.jsx
@@ -95,7 +95,8 @@ class UserWallet extends React.Component {
         const powerDown = (cancel, e) => {
             e.preventDefault()
             const name = account.get('name');
-            const vesting_shares = cancel ? '0.000000 VESTS' : account.get('vesting_shares');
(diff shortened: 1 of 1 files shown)


TimCliff:
Overlap with https://github.com/steemit/condenser/pull/1589

HappyMaster:
Oh didn't notice that fixes this issue as well. Then this pr can be closed. Or if there's more work to be done there, probably it could be a fast patch to merge.

TimCliff:
Might as well leave it open and let the dev team decide what they want to do with it. Thanks for helping out!!

jnordberg:
This was fixed in #1669

Issue #1641, opened 2017-08-09
Flag button disappears if it is on top of an hidden post


skzap:
Example:
![https://i.imgur.com/mn7gs6d.png](https://i.imgur.com/mn7gs6d.png)

Hovering the mouse cursor on top of the "FLAG" button on this setup makes the Flag Popup disappear and impossible to click the button. To reproduce, just try to downvote a content a bit on top of an already downvoted and hidden post.

Chromium Version 59.0.3071.115 (Developer Build) (64-bit)

gl2748:
Issues with flag overlay :
1. SVG elements, because they are rendered after the overlay, appear on-top of it. We want overlays to be separate from the underlying DOM hierarchy, i.e. 'last in' and 'first out'.
![11bforoor](https://user-images.githubusercontent.com/3374538/29271085-ddec22fa-80c7-11e7-8951-b47ba3f8b049.png)

2. The post-image on mobile displays has `transform: translateZ(0);` causing it to dislocate and reappear at the top of the DOM hierarchy. Basically same problem as 2. Overlay shouldn't be susceptible to this.
![screenshot from 2017-08-14 08-06-12](https://user-images.githubusercontent.com/3374538/29271067-acbb71a4-80c7-11e7-9990-4caa3e55f5b2.png)

gl2748:
https://github.com/gl2748/condenser/commit/6f3b86564e9d21a02ea353f30ce046fe51bab9b7

Resolves issues described above.
## Mobile Images:
![sss](https://user-images.githubusercontent.com/3374538/29271138-0fa02e9a-80c8-11e7-98b9-ef9f6a4b47f8.png)

## SVGs
![svgs](https://user-images.githubusercontent.com/3374538/29271172-3adfc926-80c8-11e7-8527-2064cfdedf9c.png)

plink01001:
Is this still an issue?

relativityboy:
Recent theming updates in both desktop:
<img width="984" alt="screen shot 2017-11-27 at 3 13 51 pm" src="https://user-images.githubusercontent.com/599528/33289615-c00b5bf0-d385-11e7-87d5-2c1a01504186.png">
And mobile: 
<img width="574" alt="screen shot 2017-11-27 at 3 14 29 pm" src="https://user-images.githubusercontent.com/599528/33289647-d28d4734-d385-11e7-8cdc-a064bd2a94c4.png">

Seem appear to have resolved this issue.
[...]

PR #1663, opened 2017-08-16
Fix, flag overlay issues, refs #1641

gl2748:
Screen-caps of improvements in discussion at:  #1641


relativityboy:
UI updates have cleared up the issue. This PR & Branch are no longer needed.

PR #1665, opened 2017-08-16
i18n small refactoring: adds some missing translations; removes pluralize; no concat in translation strings anymore

valzav:


src/app/components/elements/CheckLoginOwner.jsx
```diff
@@ -69,14 +69,14 @@ class CheckLoginOwner extends React.Component {
                     <span className="warning uppercase">{tt('g.warning')}:</span> {tt('postfull_jsx.your_password_permissions_were_reduced')} <TimeAgoWrapper date={last_valid_time} />. {tt('postfull_jsx.if_you_did_not_make_this_change') + ' '} <a onClick={this.recover}>{tt('g.recover_your_account')}</a>.
                 </p>
                 <p>
-                    {tt('postfull_jsx.owhership_changed_on')}: {last_valid_date.toString()}
+                    {tt('postfull_jsx.owhership_changed_on', {date: last_valid_date.toString()})}
```
sneak: this toString... does that localize?  should the date be passed into the tt() function as a date and be allowed to be localized by the translation?

src/app/components/elements/DateJoinWrapper.jsx
```diff
@@ -3,14 +3,14 @@ import tt from 'counterpart';
 
 export default class DateJoinWrapper extends React.Component {
     render() {
-        let date = new Date(this.props.date);
-        let monthNames = ["January", "February", "March", "April", "May", "June",
+        const date = new Date(this.props.date);
+        const monthNames = ["January", "February", "March", "April", "May", "June",
```
sneak: these are english month names, the join date should be passed into the translation string for date formatting in the appropriate language.

valzav: Please check out my new commit, now it uses FormattedDate from react-intl lib - our preferred way to localize dates

src/app/components/elements/CheckLoginOwner.jsx
valzav: toLocaleString() should be used instead or it might be a better idea to use FormattedDate component, I'll fix it..

valzav: done, see my last commit

PR #1667, opened 2017-08-19
FAQ Updates

TimCliff:
This pull request makes the following changes to the FAQ page:

Adds Questions:
- How do I set the thumbnail image for my post?
- What is the recommend aspect ratio for thumbnail images?
- What is a MVEST?
- What is the recommend size for the cover image?

Updates:
- Can I delete something I posted?
Clarifies that content cannot be edited after 7 days.
- What is shown in my profile?
Adds info for the cover image
- How do I change my avatar image and other profile information?
Adds info for the cover image


roadscape:
Ready to be merged?

PR #1668, opened 2017-08-19
Allow users to enter domain names without protocol

### LINKED ISSUE #1661: Unable to update profile page.
Hi, I'm unable to update my profile page.  Whenever I type in a change the update button does not highlight allowing me to change my settings.

Here is what it looks like before I changing anything 
steem profile issue 1
Here's what it looks like when I change something.

steem profile issue 2

I am using Google Chrome.  I also noticed this is an issue on my Kindle tablet as well, the same issue of not being able to update.

rayadav:
Don't require the user to prepend protocol to the domain name. Fix #1661.
screen shot 2017-08-19 at 08 02 28

### DIFF
--- a/src/app/components/modules/Settings.jsx
+++ b/src/app/components/modules/Settings.jsx
@@ -27,6 +27,12 @@ class Settings extends React.Component {
     }
 
     initForm(props) {
+        let websitePattern = new RegExp(
+        '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.?)+[a-z]{2,}|'+ // domain name
+        '((\\d{1,3}\\.){3}\\d{1,3}))'+ // OR ip (v4) address
+        '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*'+ // port and path
+        '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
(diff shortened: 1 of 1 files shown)


src/app/components/modules/Settings.jsx
```diff
@@ -187,7 +193,7 @@ class Settings extends React.Component {
 
                     <label>
                         {tt('settings_jsx.profile_website')}
-                        <input type="url" {...website.props} maxLength="100" autoComplete="off" />
+                        <input type="text" {...website.props} maxLength="100" autoComplete="off" />
```
roadscape: Why remove `url` type? It should help w/ accessibility and on mobile.

roadscape:
It's better to save the protocol rather than to push ambiguous URLs to the public account data.

One solution is to simply prepend `https://` upon saving if the URL does not match `/^https?:\/\/`. Otherwise it's up to every client to guess what it is. 

If everyone follows a more strict format then (a) UIs can ignore any entry for this field not matching `/^https?:\/\/` and (b) there is no hidden logic leaving the user guessing why it works in some places and not others.

bnchdrff:
i'll do a quick PR to add an validation error popup explaining to the user that a protocol is necessary

PR #1669, opened 2017-08-19
766 slider to control withdraw vesting amount 

jnordberg:
Fixed version of #1589


sneak:
Can I get a screenshot of what it looks/works like in the PR?

PR #1677, opened 2017-08-24
Update verbiage action signup

originated:


originated:
Additional dialog had old style wording, needs to reflect earning.

Issue #1647, opened 2017-08-10
Auto converting the tags that a user enters to lower case, rather than giving a validation error


TimCliff:
This is an enhancement issue to track a community developer task.

TAMustafa:
It looks like the code change need to happen in the following file:
src/app/components/cards/`CategorySelector.jsx`

At Line 84:
`cats.find(c => /[A-Z]/.test(c))      ? tt('category_selector_jsx.use_only_lowercase_letters') :`

Can the result of the input be passed to a function that uses the **toLowerCase()** method?

*//Function to convert tags into LowerCase*
```
function toLower(tag){
 return tag.toLowerCase();
}
```

*//And use .map method to apply array to toLower function*
`var tagToLower = c.map(toLower);`

I try to get condenser running to test this approach.

voorash:
I think the code at line 84 is only testing that the tag is all lowercase. It needs to be converted to lower case before the validation is run.

voorash:
So, changing this
```
const impProps = {...this.props}
```
to
```
const impProps = {...this.props, value: this.props.value.toLowerCase()}
```
fixes the issue but the validation runs before this code so one additional change handles that but I need a better way to do this without modifying that const.  Too late tonight so I will look at that tomorrow.
```
const ns = nextState
ns.category.value = ns.category.value.toLowerCase()
```

voorash:
Making these two changes fixes the issue. My only issue was I wasn't sure if it was ok to make that change directly to the event or if I needed to make a copy and then change that and then pass it to onChange.

```
this.categoryInputOnChange = (e) => {
          e.preventDefault()
          e.target.value = e.target.value.toLowerCase()
          this.props.onChange(e)
        }

const categoryInput =
          <input type="text" {...cleanReduxInput(impProps)} ref="categoryRef" tabIndex={tabIndex} disabled={disabled} onChange={this.categoryInputOnChange}/>
```

voorash:
Oh and that first function is added to the CategorySelector constructor right after categorySelectOnChange

stephanmullerNL:
Ha that's exactly the solution I came up with, but then the "no uppercase" validation warning still remains even though it's converted? 

I don't think modifying the event should be a problem by the way.

voorash:
I don't get the error message anymore with that change

voorash:
hmm, the extra space removal is a good idea

stephanmullerNL:
Alright, will do!

[...]

PR #1678, opened 2017-08-24
auto lower case for tags instead of error

voorash:
Not sure if it is ok to change the event object. If not let me know and I will add code to make a copy and change the copy.


voorash:
#1647

roadscape:
Perhaps fixing input `onBlur` instead of hijacking `onChange` would be better?

voorash:
There were two issues I ran into that led me to take this approach.  It has been a few weeks so I can't remember which was which but what eventually led me to this was one way fixed the problem but the validation error message still appeared and the other if you typed a category such as hellO then when it corrected it to hello the browser moved the cursor back to the beginning of the line after fixing it. There was also a case where if the last letter you typed was capital then it would go through.

relativityboy:
I've played with a couple solutions. The lightest weight would be to do the transform to all lowercase on post. The second-lightest weight would be to put a preview box just below the tag input. @pkattera ?

relativityboy:
We have a new issue ( #2061 ) that addresses this in a more user friendly way.

@voorash - thanks for working on this code, though we're not merging, it resulted in the creation of 2061, which we'll be working on in short order!

PR #1681, opened 2017-08-25
Update webpack configuration files to use version 3

valzav:


src/app/redux/MarketReducer.js
```diff
@@ -1,40 +1,29 @@
...
+
+    if (action.type === 'market/RECEIVE_ORDERBOOK') {
+        return state.set('orderbook', payload);
+    }
+
+    if (action.type === 'market/RECEIVE_TICKER') {
+        return state.set('ticker', payload);
+    }
+
+    if (action.type === 'market/RECEIVE_OPEN_ORDERS') {
```
mboperator: I recommend creating a few `constants` files that contains these strings. Without action creators or reducer tests these are _very_ susceptible to typos.

src/app/redux/TransactionReducer.js
```diff
@@ -0,0 +1,123 @@
...
+            confirm,
+            warning
+        });
+    }
+
+    if (action.type === 'transaction/HIDE_CONFIRM') {
+        return state.merge({show_confirm_modal: false, confirmBroadcastOperation: undefined, confirm: undefined});
+    }
+
+    if (action.type === 'transaction/BROADCAST_OPERATION') {
```
mboperator: These aren't necessary (but are kind of nice if you want to be explicit about the actions being thrown, even though you don't handle all of em)

valzav: Yes, unfortunately we don't have any meaningful tests coverage, this is going to be changed in soon - now we have more resources now compared to a time when we were starting. I can understand you concerns about quite a lot of different changes going into this single branch. We have thoroughly tested this branch so it should be good to go for now. In the future we are planning to put more efforts into automated testing and any help on this is very welcome!

mboperator: Ahh I totally understand. Under a time crunch tests are often the first things to get cut!
If I can find an hour or two this weekend I'll cut a branch and implement Jest :)

src/app/RootRoute.js
```diff
@@ -1,5 +1,5 @@
 // polyfill webpack require.ensure
-//if (typeof require.ensure !== 'function') require.ensure = (d, c) => c(require);
+if (typeof require.ensure !== 'function') require.ensure = (d, c) => c(require);
```
goldibex: Could you walk me through what's happening in this file with `require.ensure`?

Issue #1455, opened 2017-06-01
Market - depth chart y-axis bug


roadscape:
![image](https://cloud.githubusercontent.com/assets/5168676/26681764/3b73c8c0-46ac-11e7-8bdf-7f3cc6ddcb13.png)

The `$1.50k` makes no sense here, it should be `15k` if anything. Highcharts bug?

relativityboy:
@roadscape not Highcharts bug. A config setting... 
~~~javscript
depth_chart_config.yAxis.labels.formatter 
~~~
Turns out (10e3 != 10^3)

PR #1682, opened 2017-08-25
1455 market depth chart y axis

relativityboy:
https://github.com/steemit/condenser/issues/1455

fix bad exponents in highcharts yaxis label formatter. Keep numeric precision at increments of 1000 above 10k.

Before:
screen shot 2017-08-25 at 10 57 23 am

After:
screen shot 2017-08-25 at 10 56 32 am


roadscape:
One small nit: more legible to use `1e6` and `1e3` over `10e5` and `10e2`.

Issue #446, opened 2016-10-08
command-click (or middle click) does not open new tab when clicking on usernames


sneak:
it opens a popup with "profile", and follow and mute buttons. middle-clicking should open the link target (profile page) in another tab.

sneak:
works on /@username/feed

not on /tag/@username/postname

relativityboy:
An example of the popup behavior on /tag/
<img width="502" alt="screen shot 2017-08-25 at 2 43 31 pm" src="https://user-images.githubusercontent.com/599528/29729990-0abb7bd4-89a4-11e7-9eed-65b6915af252.png">

I'm thinking the 'down carrot' should open the popup, but the username should be a classic link to the user.

relativityboy:
After a conversation with the UI team, the decision was made to move the popup to the user logo, and have the username be a link (as it is in most, if not all other locations in the site)

relativityboy:
After addl review, the popup will be on click of the down-carrot only. It'll be considered an 'expert level' feature. We'll revisit pending user feedback as we do design upgrades.

plink01001:
Has this been addressed / can this be closed?

PR #1686, opened 2017-08-29
446 linkify username in comment

relativityboy:
https://github.com/steemit/condenser/issues/446

Author name is now always a link. Default browser behavior on the links is canceled *unless* <meta> or <ctrl> keys are pressed while clicking. A toggle event for the popup is fired if those keys are not pressed.


sneak:
There is a history bug.

Steps to reproduce:

1) Open link to article page

2) Click username to show popup

3) click username in popup to go to blog page

4) hit back

expected:

return to post page

actual:

nothing happens


(this is not in relation to the issue i filed with middle-clicking to make a new tab; just on the single, original tab.)

relativityboy:
@sneak there was a bug with addEventListener and removeEventListener not always being present. Added checks; that's fixed most of the "back" navigation problem.

relativityboy:
@roadscape - there's variation in links. Some are underlined, some are not. I've added a bit of styling to remove the underline for these links.

PR #1689, opened 2017-08-29
Add support for syntax highlighting

jnordberg:
Only enabled for gfm language tagged codeblocks, e.g. **```javascript**

Using https://highlightjs.org


goldibex:
@roadscape please review per @sneak above

trashhalo:
Is the pr dead on the vine? Would be lovely to get syntax highlighting merged as I'd like to do programming tutorials on steem.

bnchdrff:
highlightjs seems to be pretty solid, and has no huge runtime deps.. any other comments?

i, also, would love to have this feature.

jnordberg:
Ping @sneak @roadscape

roadscape:
Anything in the rendering path requires in-depth security review. A similar request we had was MathJax integration, which was deemed impractical to implement safely (https://github.com/steemit/condenser/issues/7). A cursory check of highlight.js suggests it's safe to use but we need a thorough review.

sneak:
I’m for this, if someone qualified reviews the version of highlight.js we are using and @goldibex signs off on the bundle size increase.

src/app/components/cards/MarkdownViewer.jsx
```diff
@@ -13,7 +14,14 @@ const remarkable = new Remarkable({
     breaks: true,
     linkify: false, // linkify is done locally
     typographer: false, // https://github.com/jonschlinkert/remarkable/issues/142#issuecomment-221546793
-    quotes: '“”‘’'
+    quotes: '“”‘’',
+    highlight: (src, lang) => {
+        if (lang && hljs.getLanguage(lang)) {
+            try {
+                return hljs.highlight(lang, src).value
+            } catch (e) {}
```
goldibex: error needs caught and handled properly

relativityboy:
With the bundle size increase of hjs and changes coming to condenser I'm closing this and opening issue #2095 for syntax highlighting in its place.

bnchdrff:
what is the bundle size increased by?

src/app/components/cards/MarkdownViewer.jsx
jnordberg: What do you consider properly in this case? We fall back to not syntax highlighting if the method throws an error.

jnordberg:
I would like to know that as well... I think it would not be that a large of an increase, and if it is to big highlight.js can be built with fewer supported languages. @relativityboy I'm reopening this ;)

bnchdrff:

jnordberg:
😢

bnchdrff:
the day will come @jnordberg !!!

PR #1701, opened 2017-09-01
1153 kill nested steemit proxy urls

relativityboy:
Add SanitizeConfig.js - it takes a url and a bool 'stripAllProxies' argument.

It tries to be safe about how it strips the proxies. Might be overkill; might save some bacon.


roadscape:
Awesome, I'll add some test cases as part of review. 

Could we rename the method? Sanitize has a specific meaning when it comes to the rendering process. This is perhaps some form of strip, collapse, or cleaning of a proprietary proxy URL.

sneak:
This is precisely what we need; great work @relativityboy.  Excited to get this shipped soon.

relativityboy:
@roadscape - I had similar thoughts about the name, but since this is used inside a rendering+sanitization component I followed that. 

It could be a great place to perform other url operations.

Off the cuff: 'cleanUrl', 'transformUrl', 'processUrl', 'shiftUrl'.

roadscape:
Let's be explicit for now. This method only applies to image URLs and will attempt to cleanly proxify them. How about `proxifyImageUrl`?

roadscape:
I've added some basic tests, let's get these passing and I'll throw some more in.

To run just this test: `npm run mocha -- src/app/utils/SanitizeUrl.test.js`

TODO: audit usages of `img_proxy_prefix`, they could all go through this method.

relativityboy:
I've updated the names, and expanded the tests to cache some potential issues if devs go mucking with the regex down the road.

src/app/utils/ProxifyUrl.js
```diff
@@ -0,0 +1,24 @@
+/*global $STM_Config:false*/
+
+//this regular expression should capture all possible proxy domains
+const rProxyDomains = /http(s)?:\/\/steemit(dev|stage)?images.com\/([0-9]+x[0-9]+)?\//g
+
+/**
+ * Strips all but final 'steemit' image proxy from the beginning of the url.
```
sneak: i think it should strip all of them, then add just the one that is specified in the condenser config

roadscape:
Added a few more tests and they all pass. Nice catch w/ the steemit hosted URLs, I thought for sure I'd break your stuff :p

With regards to usage of `img_proxy_prefix`, I see:
```
grep -r '$STM_Config.img_proxy_prefix' src/ | grep -v ProxifyUrl
src//app/components/elements/Userpic.jsx:            if($STM_Config.img_proxy_prefix) {
src//app/components/elements/Userpic.jsx:                url = $STM_Config.img_proxy_prefix + size + '/' + url;
src//app/components/pages/UserProfile.jsx:            if($STM_Config.img_proxy_prefix) {
src//app/components/pages/UserProfile.jsx:                cover_image_prefix = $STM_Config.img_proxy_prefix + '2048x512/';
src//app/utils/SlateEditor/Image.js:        const prefix = $STM_Config.img_proxy_prefix ? ($STM_Config.img_proxy_prefix + '0x0/') : ''
src//shared/HtmlReady.js:    if (!$STM_Config.img_proxy_prefix) return
src//shared/HtmlReady.js:            node.setAttribute('src', $STM_Config.img_proxy_prefix + '0x0/' + url)
```

The most important one to get is `HtmlReady.js`, this contains most of the rendering meat and cleans all images in the body. Please update `Userpic.jsx` and `UserProfile.jsx` as well. Then we are good to go.

src/app/utils/ProxifyUrl.js
roadscape: This comment is (now) incorrect, what you describe is what happens.

relativityboy:
@roadscape - updated in all requested locations. Please take a close look @ Userpic.jsx the code seemed a little clunky, so I cleaned it up. Think I caught all possible cases, but I'd like another set of eyes.

roadscape:
Looks good. Waiting to test in dev then let's merge

roadscape:
Looks good on stage. There's just one last potential concern (or at least something to be aware of): if a user was using the proxy URL to resize the image to a specific size, we will no longer honor those requested dimensions.

roadscape:
We could implement some logic for smart handling of nested dimensions, e.g.:

 - 0x0..240x320 would use 240x320
 - 100x10..0x0..10x10 would use 100x10?

But it's an undocumented feature, so nobody should be relying on it. Let's merge.

PR #1702, opened 2017-09-02
Add comment rendering limit

jnordberg:
Case in point: https://steemit.com/mangosteem-channel/@ms-general/mangosteem-message-archive


src/app/components/pages/Post.jsx
```diff
@@ -89,13 +89,21 @@ class Post extends React.Component {
...
+        let replies = dis.get('replies').toJS();
 
         let sort_order = 'trending';
         if( this.props.location && this.props.location.query.sort )
            sort_order = this.props.location.query.sort;
 
         sortComments( content, replies, sort_order );
+
+        // Don't render too many comments
+        const commentLimit = (global['process'] !== undefined) ? 500 : 1000
```
roadscape: Why different limits here?

jnordberg: Reasoning was to avoid spending too much time server-side while still letting the browsers render more comments

goldibex: ``` (global['process'] !== undefined) ```
is there possibly a more expressive way of saying "on the server side?" even if you just move this out into a helper function somewhere, many less expert js devs don't know about the environmental differences between node and the browser.

bnchdrff:
tested this locally; works for a few posts. request time for a page w/ many comments predictably goes from ~65 seconds average to ~15 seconds

PR #1705, opened 2017-09-04
Improve verbiage of delegation on steem power, include username in ve…

originated:
…rbiage for clearer understanding

PR #1708, opened 2017-09-05
Improve tooltip about delegation in UserWallet.jsx

dittaeva:
Current version is misleading, as delegation can go both ways, not just "to".


goldibex:
does this conflict with #1705?

dittaeva:
> does this conflict with #1705?
That issue does not touch the tooltip, but perhaps it should. The wording is not in conflict.

bnchdrff:
this text should probably be translated.

bnchdrff:
@dittaeva thanks for submitting this PR! i'm going to cherry-pick your commit into the work in #1705

PR #1711, opened 2017-09-05
Store user preferences on the server, add language selector to settings

valzav:


sneak:
@valzav is this still WIP?

src/server/app_render.jsx
```diff
@@ -16,7 +16,10 @@ async function appRender(ctx) {
     const store = {};
     try {
         let login_challenge = ctx.session.login_challenge;
-        let user_preferences = null;
+        let locale = ctx.getLocaleFromHeader();
+        if (locale) locale = locale.substring(0, 2);
+        if (!locale.match(/(en|es|ru)/)) locale = 'en';
```
sneak: i think the list of supported locales should be autogenerated on server startup from the locale files available in the directory and not hardcoded into a regex buried in the middle of app_render.jsx. at the very least they should be a list in `$STM_Config`.

valzav:
@sneak, it's WIP because it's not well tested yet, also it requires one more db table that we need to make sure is created before it's going to production. I'll make it to fetch the list of locales from files.

valzav:
@sneak please check out my last commit

goldibex:
@valzav what's the status of this PR?

valzav:
@goldibex: it's ready to go, let's test it once again after we merge in wepback3

src/app/Translator.js
```diff
@@ -3,14 +3,21 @@ import {connect} from 'react-redux'
...
+import ru from 'react-intl/locale-data/ru';
 import {DEFAULT_LANGUAGE} from 'app/client_config';
 import tt from 'counterpart';
 
-addLocaleData([...en, ...es]);
+addLocaleData([...en, ...es, ...ru]);
 
+tt.registerTranslations('en', require('counterpart/locales/en'));
 tt.registerTranslations('en', require('app/locales/en.json'));
 tt.registerTranslations('es', require('app/locales/es.json'));
```
goldibex: why isn't there a `counterpart/locales/es` call?

src/app/components/modules/Settings.jsx
```diff
@@ -244,7 +232,7 @@ export default connect(
         let metaData = account ? o2j.ifStringParseJSON(account.json_metadata) : {}
         if (typeof metaData === 'string') metaData = o2j.ifStringParseJSON(metaData); // issue #1237
         const profile = metaData && metaData.profile ? metaData.profile : {};
-        const locale = state.user.get('locale');
+        const user_preferences = state.app.get('user_preferences').toJS();
```
goldibex: camelcase please

src/server/api/general.js
```diff
@@ -435,6 +435,36 @@ export default function useGeneralApi(app) {
         }
         this.body = JSON.stringify({status: 'ok'});
     });
+
+    router.post('/setUserPreferences', koaBody, function *() {
+        const params = this.request.body;
+        const {csrf, payload} = typeof(params) === 'string' ? JSON.parse(params) : params;
+        if (!checkCSRF(this, csrf)) return;
```
goldibex: send an error code and message here (403 maybe?)

src/server/app_render.jsx
```diff
@@ -83,8 +101,18 @@ async function appRender(ctx) {
                 offchain.recover_account = account_recovery_record.account_name;
             }
         }
+        if (ctx.session.a) {
+            const user_preferences_record = await models.UserPreferences.findOne({
```
goldibex: camelcase

src/shared/UniversalRender.jsx
```diff
@@ -59,7 +59,7 @@ const onRouterError = (error) => {
     console.error('onRouterError', error);
 };
 
-async function universalRender({ location, initial_state, offchain, ErrorPage, tarantool }) {
+async function universalRender({ location, initial_state, offchain, ErrorPage, tarantool, user_preferences }) {
```
goldibex: camelcase

src/server/api/general.js
valzav: checkCSRF does exactly that - it sets 403 status if check fails

originated:
@valzav what do you think of merging latest master here? So we can handle any merge conflicts

valzav:
@goldibex, please review the latest changes

src/server/api/general.js
goldibex: Sorry, I meant the side effects should be here in this method as opposed to in `checkCSRF`.

goldibex:
@valzav Please do not dismiss my reviews; ask me to recheck instead.

src/server/api/general.js
valzav: I prefer pure functions too, but we are using checkCSRF in many places, so moving 403 out of checkCSRF will require a lot work and changes, I think this should be done separately from this PR

goldibex: :+1:

src/app/utils/ServerApiClient.js
```diff
@@ -89,6 +89,12 @@ export function saveCords(x, y) {
     fetch('/api/v1/save_cords', request);
 }
 
+export function setUserPreferences(payload) {
+    if (!process.env.BROWSER || window.$STM_ServerBusy) return Promise.resolve();
```
goldibex: does this not mean this function silently short-circuits if `window.$STM_ServerBusy` is set? Shouldn't we reject immediately in that case, or wait around until the server's not busy anymore?

valzav: Server Busy usually means that mysql is not available so there is no point in saving user preferences at this point.
window.$STM_ServerBusy can only change on page reload.

src/app/Translator.js
valzav: es is not fully supported yet, I'll will add some more language specific stuff in a separate brunch after this one is merged

src/app/components/modules/Settings.jsx
goldibex: camelCase

src/app/components/modules/Settings.jsx
```diff
@@ -136,22 +125,23 @@ class Settings extends React.Component {
 
         const {profile_image, cover_image, name, about, location, website} = this.state
 
-        const {follow, account, isOwnAccount, locale} = this.props
+        const {follow, account, isOwnAccount, user_preferences} = this.props
```
goldibex: camelCase

src/app/Main.js
```diff
@@ -98,9 +98,10 @@ function runApp(initial_state) {
 
 if (!window.Intl) {
     require.ensure(['intl/dist/Intl'], (require) => {
-        window.IntlPolyfill = window.Intl = require('intl/dist/Intl')
-        require('intl/locale-data/jsonp/en-US.js')
-        require('intl/locale-data/jsonp/es.js')
+        window.IntlPolyfill = window.Intl = require('intl/dist/Intl');
```
goldibex: these changes should be removed from this PR if we're no longer addressing internationalization here

src/app/components/modules/Settings.jsx
valzav: camelCase for component props too?

PR #1712, opened 2017-09-05
move npm commands to yarn in Dockerfile

sneak:


Dockerfile
```diff
@@ -19,9 +19,7 @@ COPY . /var/app
 #  ./node_modules/.bin/eslint . && \
 #  npm run build
 
-RUN mkdir tmp && \
-  npm test && \
-  npm run-script build
+RUN yarn run test && yarn run build
```
goldibex: this can just be `yarn test && yarn build`

PR #1715, opened 2017-09-06
konami-code

relativityboy:
added k0nami monitoring to enable network logging. depends on npm linked steem-js branch api-injected-logging-support

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "sequelize-cli": "^2.3.1",
     "speakingurl": "^9.0.0",
     "sqlite3": "^3.1.8",
-    "steem": "github:steemit/steem-js#2454f5e",
+    "steem": "github:steemit/steem-js#5f50ad5",
     "store": "^1.3.20",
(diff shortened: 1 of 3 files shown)


roadscape:
lgtm. Will test after 1701.

sneak:
So... this is a good example of what we need to be careful with in code review.

steem-js obviously is a critical piece, and we could easily accidentally ship a serious issue into condenser with a change like this.  Any PR that updates the commit hash id for steemjs should be a red flag and get extra scrutiny.

Including a link such as the following in any PRs is basically required to pass code review:

https://github.com/steemit/steem-js/compare/2454f5e406f6b88f3153185dd7a83f134317f972...5f50ad593412c8a12ab01433bc3009f702994908?expand=1

This is the comparison link to view the differences between the expanded commit IDs of the two steemjs versions updated in package.json.

@roadscape @relativityboy @valzav @originated @jnordberg - be advised.  Any review of condenser code that changes security-sensitive package versions MUST also review the changes to the dependent packages.

PR #1718, opened 2017-09-07
add check translations script

valzav:


scripts/check_translations.js
```diff
@@ -41,12 +41,12 @@ function loadTranslationFiles(path) {
...
     const lines = fs.readFileSync(path, 'utf8').split(/\r?\n/);
     for (const l of lines) {
-        const tts = l.match(/(tt\([\"\'\.\-\_\w]+)/g);
+        const tts = l.match(/(tt\([\"\'\.\-\_\w]+)/g) || l.match(/(FormattedHTMLMessage.+id\=[\"\'\.\-\_\w]+)/g);
         if (tts) {
             // if(tts.length > 1) console.log('-- tt -->', path, l, tts.length, JSON.stringify(tts, null, 4));
             for (const t of tts) {
                 if (t === 'tt(id') continue; // this is exception
-                const m = t.match(/tt\([\'\"]([\.\-\_\w]+)/);
+                const m = t.match(/tt\([\'\"]([\.\-\_\w]+)/) || t.match(/id\=[\'\"]([\.\-\_\w]+)[\'\"]/);
```
sneak: I'm approving this but a comment or better named variables to explain what this code is doing would be ideal here; it's not clear from looking at it unless one a) knows what is being searched for and b) parses the regex visually by reading.  `m` is not a good variable name.

remember: programs are written for humans to read, and only incidentally for computers to execute.

PR #1719, opened 2017-09-07
Improve request logging

jnordberg:
This adds a little better request logging (including account names), it's pretty ugly but since we are working on moving out the api endpoints to their own services I think it's an ok temporary fix.


sneak:
@Gandalf-the-Grey for now the codeclimate issues aren't a hard requirement and they don't fail the merge, I just want people to be aware of them.  we'll start enforcing linting soon.

I like how you handled this @jnordberg, thanks for the PR.   Approved.

sneak:
Oh, val already approved it. @jnordberg if you've tested this, please merge it to master and deploy it to staging, test it there, and then if it doesn't break stuff, deploy it to prod.

jnordberg:
It's tested in dev so should be good to go. I don't have write permissions on condenser yet so you'll have to merge it for me

jnordberg:
We still need to get this merged

PR #1724, opened 2017-09-10
Add links to "The Steemit Shop" and the "Steem Bluepaper" in the main menu

TimCliff:
This PR adds a link to "The Steemit Shop" and the "Steem Bluepaper" in the main menu. Both of the links are external sites and open in a new window. The link to the Whitepaper was also updated to open in a new window, which is consistent with the other external links.


goldibex:
will merge and stage

PR #1726, opened 2017-09-11
Fix some of the i18n issues

valzav:


roadscape:
I've reviewed the code and tested on dev. Looks good, approved.

Two minor things stood out --

Awkward sentence:
```
SAVINGS
Balance subject to 3 day withdraw waiting period,STEEM DOLLARS currently collecting 0.5% APR.
```

Should be STEEM since it's referring to the token:
```
Estimated Account Value
The estimated value is based on an average value of Steem in US dollars.
```

PR #1727, opened 2017-09-11
Depend on @steemit/render for html sanitisation

jnordberg:
This is the first step of moving out the post body rendering to a separate module

https://github.com/steemit/steemit-render


jnordberg:
I think we should invest some time into setting up a npm package deployment pipeline. npm modules are not meant to be depended on from git references.

We could solve the build problem like steem-js with a install hook that builds the library on every install but aside from being slow that has the problem of negating the devDependencies/dependencies split packages have.

And it is against [best practices](https://docs.npmjs.com/misc/scripts):

> You should almost never have to explicitly set a preinstall or install script. If you are doing this, please consider if there is another option. The only valid use of install or preinstall scripts is for compilation which must be done on the target architecture.

Re yarn.lock, it has a package-lock.json. Not sure if yarn will use that or not.

sneak:
Yarn will not.  It needs a yarn.lock.

We are explicitly deciding to use git URLs plus cryptographic hashes.  1) npm's a clusterfuck and 2) we don't want to have to make releases.  these are internal packages and we're basically using this as a type of submodule.  alternately if you're that against doing it this way, put the sanitizer module back into the condenser repo - but keep it as a separate module in its own directory.  the breaking it out into its own git repo is simply a matter of convention.

i prefer it this way, and referenced by git url plus commit hash. we benefit nothing from an npm package deployment pipeline.

sneak:
> has the problem of negating the devDependencies/dependencies split packages have

practically I don't see this as a problem at all. the npm ecosystem is stupid and expects a compiler and headers available at all times even for prod dependencies so who cares if it builds or downloads an extra few megabytes of crap sitting in node_modules that isn't used at runtime? it doesn't hurt anything.

sneak:
hey @jredbeard - what was the thing we were doing for `steem-js` to do the appropriate build steps at install time?  (I'm withholding my review until I get an update on this.)

jredbeard:
@sneak @jnordberg it uses the `postinstall-build` package and it's called from the postinstall script in package.json, see: https://github.com/steemit/steem-js/commit/84d709fc689b449e07f70919f4bf5d0a0d730370

jnordberg:
> @sneak @jnordberg it uses the postinstall-build package and it's called from the postinstall script in package.json, see: steemit/steem-js@84d709f

Yeah I've seen that and also experienced problems with it more than once. FYI that relies on a globally installed npm to build causing it to fail if you have the wrong version and it's pulling in dependencies via npm not respecting your yarn.lock

> We are explicitly deciding to use git URLs plus cryptographic hashes. 1) npm's a clusterfuck and 2) we don't want to have to make releases. these are internal packages and we're basically using this as a type of submodule. alternately if you're that against doing it this way, put the sanitizer module back into the condenser repo - but keep it as a separate module in its own directory. the breaking it out into its own git repo is simply a matter of convention.

Having a package ecosystem with every package pulling in the dev dependencies and building on install... that would be a clusterfuck if anything

Keeping it in its own repo or in the same does not matter, we would still need to have the install hooks unless we want to version control the compiled modules as well.

> practically I don't see this as a problem at all. the npm ecosystem is stupid and expects a compiler and headers available at all times even for prod dependencies so who cares if it builds or downloads an extra few megabytes of crap sitting in node_modules that isn't used at runtime? it doesn't hurt anything.

That's not true, only a small subset of modules do that, and many of the popular ones that have binary dependencies offer pre-built variants when installing with `npm install --production`. What I'm worried about is that the build process is fragile and slow, a failure in a build script will cause your entire install to fail. Further when we start having more modules with inter-dependenices keeping track of what depends on what via git refs will be a nightmare. With published builds we can follow semver and depend on our own packages with less strict versioning since we know that everything we publish has been vetted.

jnordberg:
And I don't think setting up a npm publish pipeline in watt would be a big project, it's basically this:

```
git pull origin master
npm install
npm test
npm publish
```

sneak:
> many of the popular ones that have binary dependencies offer pre-built variants

Downloading and installing pre-built binaries as part of our dependency install is yet another thing that terrifies me about the entire process.  Any tool or package that does this is suspect.

sneak:
Well, the issue we run into there is several-fold:

1) we are depending on the security of npm for integrity of our specific packages (yes, I know we are doing this for other packages like react, but presumably those have more people paying attention to their integrity)

2) watt needs credentials for npm publish, expanding attack surface and risk if compromised

3) watt needs a build environment (probably dockerized, but still)

I would prefer to not have to deal with any of these; so far none of the cited problems with using a commit-hash-versioned dependency reference outweigh any of these.

sneak:
If steemit-render is switched to yarn/yarn.lock, will this postinstall-build process still work if the consuming package (condenser) is installed with yarn?

jredbeard:
@jnordberg suggested making a custom yarn compatible version of the
postinstall-build package, I think this is the way to go of it's ignoring
the yarn.lock - @jnordberg can you test and confirm?

jnordberg:
I had a look at implementing something similar to `postinstall-build` with yarn but it is not as easy as I imagined... I can probably make it work if we really want to go down this path but it's going to take some time to get right

I still think that the headaches this creates outweighs the security concerns... We are depending on over 1000 packages from the npm registry now, adding a couple of more that we strictly control isn't going to change the threat.

Btw, right now on master, yarn install is broken due to steem-js throwing an error during postinstall. And the error comes from npm:

```
npm WARN The package bluebird is included as both a dev and production dependency.

npm ERR! May not delete: ./condenser/node_modules/steem/node_modules/.bin
```

```
$ node -v
v8.5.0
$ yarn -v                                                          
1.0.2
$ git rev-parse HEAD
2b4814baa2c949fd1f3f3944c52fb6f713a86e4e
```

jnordberg:
Ok, it wasn't that hard 😊 https://github.com/jnordberg/ypib

(still don't think this is a good idea though)

sneak:
The fact of the matter is that we can do very little about the packages we don't strictly control; but we absolutely can do something about potentially *silently* including changes from packages we do, such as our most critical crypto ones.

Steemit is not a normal app, and condenser is not a normal codebase.  We have special requirements, and while this is not a perfect solution, it's absolutely better than *not having cryptographically versioned dependencies*.

Yes, it's additional hassle.  Yes, it's swimming upstream.  But sometimes you have to swim upstream when the stream is filled with a cascade of attention deficit teenagers pumping out shit like `leftpad` (and `npm`, which was NONDETERMINISTIC recently).

The javascript ecosystem sucks *badly*, and we are going to bloody ourselves repeatedly doing what we can to avoid common pitfalls, and it won't even be that satisfying because, as you pointed out, it's still not a perfect solution.  But it's still better than the alternative, which is doing nothing like everyone else.

I apologize in advance for the hassle, but whenever we have the option of locking down a set of code, *even a subset*, to that which is validated by a merkle root like a git commit hash or an ipfs multihash or a package checksum—then we will.  Anywhere we can make the network, storage, or database trustless instead of trusted—then we will.  Yes, it's more work.  I am deeply and truly sorry for you and I and everyone else who has ever known Eich's bastard hellspawn utterance of `preventDefault`.  This is the life we chose.

jnordberg:
I understand your concerns and agree, but how is a git hash is more cryptographically versioned than a sha512sum of the package tarball we have in the lockfile?

jnordberg:
Btw lockfiles are only considered for the the root package and not submodules so @steemit/render not having one does not matter. I bet that running a separate `yarn install` or `npm install` as a postinstall hook messes that up though... It's asking for trouble running a second package install that modifies the state of the ongoing install.

https://docs.npmjs.com/files/package-lock.json

jnordberg:
We now have a build pipeline for js packages. This can be merged after addressing steemit/steemit-render#4, steemit/steemit-render#5 and steemit/steemit-render#6

bnchdrff:

PR #1735, opened 2017-09-13
Add "#" to "Steemit" when using Twitter share

andrarchy:
When a post is shared on twitter through the twitter share button, the title to the post is included and "Steemit" is appended, adding a hashtag will improve analytics and social media performance. This change is intended to add a hashtag so that every time a steemit post is shared on twitter it features "#steemit."


sneak:
Looks good @andrarchy - has this been tested?  It looks like you created the branch on a fork in your own user account instead of a branch in the steemit repo, which means you can't do a test deploy to dev...

valzav:
@andrarchy, I agree with Sneak - it's more practical to use local branch instead of fork to submit PR, I'm merging it for now without deploying to dev because this change only affects one string, if it's more changes we wouldn't be able to merge it.

PR #1748, opened 2017-09-21
Debugging instuctions

codewithcheese:
Hi guys, there doesn't seem to be much written about your local dev process. I wrote this Steemit post to make it easier for others to start debugging your front end. Figured it would be useful in the read me.

Would you mind shedding some light on your dev process if you have a alternative method for debugging?

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,13 @@ docker build -t="myname/condenser:mybranch" .
 docker run -it -p 8080:8080 myname/condenser:mybranch
 ```
 
+#### Debugging
+
+To starting debugging your local install refer to the following instructions.
+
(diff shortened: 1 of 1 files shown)


valzav:
@codewithcheese, thanks for the debugging instructions. The only issue I have with this PR is that having a link to post can make it more difficult to keep the information up to date with current state or the code, I would suggest instead of the link just add a plain text instructions to the Readme, make it short, no screenshots are necessary.

bnchdrff:
Also, we will hopefully be making some toolchain changes very soon which will make debugging a simpler -- stay tuned ;)

codewithcheese:
Closed in lieu of debugging updates.

PR #1750, opened 2017-09-22
Linted elements up to PageViewsCounter.jsx

NateBrune:


bnchdrff:
this will be addressed by #2081

PR #1761, opened 2017-09-26
Remove conditional check for sign up actions, we should always displa…

originated:
…y for less friction & more signups

### DIFF
--- a/src/app/components/modules/LoginForm.jsx
+++ b/src/app/components/modules/LoginForm.jsx
@@ -211,12 +211,11 @@ class LoginForm extends Component {
                         {tt('g.cancel')}
                     </button>}
                 </div>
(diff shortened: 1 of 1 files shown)


bnchdrff:
looks good to me; tested locally & it works as expected!

src/app/components/modules/LoginForm.jsx
```diff
@@ -211,12 +211,11 @@ class LoginForm extends Component {
                         {tt('g.cancel')}
                     </button>}
                 </div>
-                {authType == 'Posting' &&
```
bnchdrff: this is great because authType is defined as the result of a call to tt, which i'm pretty sure will return a translated string -- there's no way of knowing this will be `Posting` in the first place, depending on the locale.

PR #1763, opened 2017-09-27
Add a confirm to broadcast containing private key wifs

jnordberg:
This adds an extra layer of protection to users while still allowing you to post a private key if needed (for examples etc.)

I put it at the broadcast level so you will have to confirm any transaction containing keys, should catch all cases, comments/transaction memos/profile description etc.

### DIFF
--- a/src/app/locales/en.json
+++ b/src/app/locales/en.json
@@ -203,6 +203,10 @@
       "zero": "No Responses",
       "one": "%(count)s Response",
       "other": "%(count)s Responses"
+    },
+    "post_key_warning": {
+        "confirm": "It looks like you are about to post a STEEM private key or master password to the public, are you sure you want to do this?",
+        "warning": "You can loose control of your account and funds by doing this."
     }
   },
   "navigation": {
--- a/src/app/locales/en.json
+++ b/src/app/locales/en.json
@@ -205,8 +205,8 @@
       "other": "%(count)s Responses"
     },
     "post_key_warning": {
-        "confirm": "It looks like you are about to post a STEEM private key or master password to the public, are you sure you want to do this?",
-        "warning": "You can loose control of your account and funds by doing this."
+        "confirm": "You are about to publish a STEEM private key or master password. You will probably lose control of the associated account and all its funds.",
+        "warning": "Legitimate actors, including employees of Steemit Inc., will never ask you for a private key or master password."
     }
   },
   "navigation": {
--- a/src/app/locales/en.json
+++ b/src/app/locales/en.json
@@ -206,7 +206,7 @@
     },
     "post_key_warning": {
         "confirm": "You are about to publish a STEEM private key or master password. You will probably lose control of the associated account and all its funds.",
-        "warning": "Legitimate actors, including employees of Steemit Inc., will never ask you for a private key or master password."
+        "warning": "Legitimate users, including employees of Steemit Inc., will never ask you for a private key or master password."
     }
   },
   "navigation": {
--- a/src/app/locales/en.json
+++ b/src/app/locales/en.json
@@ -206,7 +206,8 @@
     },
     "post_key_warning": {
         "confirm": "You are about to publish a STEEM private key or master password. You will probably lose control of the associated account and all its funds.",
-        "warning": "Legitimate users, including employees of Steemit Inc., will never ask you for a private key or master password."
+        "warning": "Legitimate users, including employees of Steemit Inc., will never ask you for a private key or master password.",
+        "checkbox": "I understand"
     }
   },
   "navigation": {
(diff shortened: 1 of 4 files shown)


jnordberg:
screen shot 2017-09-27 at 14 15 27

src/app/redux/TransactionSaga.js
```diff
@@ -125,15 +126,23 @@ function* error_account_witness_vote({operation: {account, witness, approve}}) {
...
-    const operationParam = {type, operation, keys, username, password, successCallback, errorCallback}
+    const operationParam = {type, operation, keys, username, password, successCallback, errorCallback, postUnsafe}
+
     const conf = typeof confirm === 'function' ? confirm() : confirm
     if(conf) {
         yield put(tr.actions.confirmOperation({confirm, warning, operation: operationParam, errorCallback}))
         return
     }
     const payload = {operations: [[type, operation]], keys, username, successCallback, errorCallback}
+    if (!postUnsafe && hasPrivateKeys(payload)) {
```
relativityboy: `!postUnsafe && ...`? if the post is unsafe I would expect postUnsafe to be falsy, rather than truthy. What's happening here?

jnordberg: If you read it as `allowPostUnsafe` maybe it makes sense? I'm no fan of how redux works and that's the best way I could figure out how to show that modal, happy to hear an alternative

src/app/locales/en.json
```diff
@@ -203,6 +203,10 @@
       "zero": "No Responses",
       "one": "%(count)s Response",
       "other": "%(count)s Responses"
+    },
+    "post_key_warning": {
+        "confirm": "It looks like you are about to post a STEEM private key or master password to the public, are you sure you want to do this?",
+        "warning": "You can loose control of your account and funds by doing this."
```
sneak: *lose

src/app/redux/TransactionSaga.js
relativityboy: `allowPostUnsafe` is a fantastic name for this purpose.

src/app/locales/en.json
```diff
@@ -203,6 +203,10 @@
       "zero": "No Responses",
       "one": "%(count)s Response",
       "other": "%(count)s Responses"
+    },
+    "post_key_warning": {
```
goldibex: language-wise we should make this warning a lot louder. maybe:

```json
"post_key_warning": {
  "confirm": "You are about to publish a STEEM private key or master password. You will probably lose control of the associated account and all its funds.",
  "warning": "Legitimate actors, including employees of Steemit Inc., will never ask you for a private key or master password."
}
```

src/app/redux/TransactionSaga.js
```diff
@@ -159,6 +168,10 @@ function* broadcastOperation({payload:
     }
 }
 
+function hasPrivateKeys(payload) {
```
goldibex: because of the optionality of the P, this will trigger for literally any 50-character alphanumeric string that starts with a decimal 5. is there really no other way to validate a private key?

jnordberg: Not literally any alphanumeric string because the keys are base58 :) but yeah, there will be false positives. The keys have a checksum, I can add a check for that as well

src/app/locales/en.json
Gandalf-the-Grey: Agree, proposed "control of the associated account" is much better and more accurate than "control of your account and funds"

src/app/redux/TransactionSaga.js
Gandalf-the-Grey: Strictly speaking, Base58 compliant string so unlikely for false positives on such long scripts but we can also perform [WIF checksum checking](https://en.bitcoin.it/wiki/Wallet_import_format)

sneak: I dont think checksum checking is necessary. What if they mistranscribe it? We still want to block secret keys and passwords that are slightly off. I am fine with this solution.

jnordberg: It's not too uncommon for our imagehoster hashes to create false positives, I think we should add the check, and I already wrote it so... :)

jnordberg: Done! (@goldibex please try to put review comments on a line that would be changed so they get invalidated)

jnordberg:
screen shot 2017-09-29 at 11 42 10

originated:
It's a very good warning but what about changing the "actors" for "users" part, this lingo is most likely not understood by an average user.

originated:
It's a very good warning but what about changing the "actors" for "users" part, this lingo is most likely not understood by an average user.  I've pushed a commit for verbiage change suggestion.

jnordberg:
screen shot 2017-10-03 at 23 04 51

jnordberg:
Can someone merge?

Jolly-Pirate:
Best idea ever.

PR #1765, opened 2017-09-27
recommend node/npm/yarn versions

bnchdrff:
this resolved all of the phantomy mysterious build issues, and should do the trick for others as well.


goldibex:
I'll stage this.

PR #1766, opened 2017-09-27
Assume development environment unless explicitly set to production

jnordberg:
This allows for multiple custom development environments. Very useful to have a set of non-versioned `config/something.json` configs and start with `NODE_ENV=something yarn run start` to test.

Also standardizes the production check, now we check everywhere for `==/!= 'production'`

### DIFF
--- a/src/app/Main.js
+++ b/src/app/Main.js
@@ -20,7 +20,7 @@ const CMD_LOG_TOGGLE = 'log-toggle';
 const CMD_LOG_O = 'log-on';
 
 try {
-    if (process.env.NODE_ENV === 'development') {
+    if(process.env.NODE_ENV !== 'production') {
         // Adds some object refs to the global window object
         ConsoleExports.init(window);
     }
--- a/src/server/app_render.jsx
+++ b/src/server/app_render.jsx
@@ -12,8 +12,9 @@ import { determineViewMode } from '../app/utils/Links';
 
 const path = require('path');
 const ROOT = path.join(__dirname, '../..');
+//HERE IAIN
 const DB_RECONNECT_TIMEOUT =
-    process.env.NODE_ENV === 'development' ? 1000 * 60 * 60 : 1000 * 60 * 10;
+    process.env.NODE_ENV !== 'production' ? 1000 * 60 * 60 : 1000 * 60 * 10;
 
 function getSupportedLocales() {
     const locales = [];
@@ -166,7 +167,7 @@ async function appRender(ctx) {
         const assets = require(assets_filename);
 
         // Don't cache assets name on dev
-        if (process.env.NODE_ENV === 'development') {
+        if (process.env.NODE_ENV !== 'production') {
             delete require.cache[require.resolve(assets_filename)];
         }
 
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "fmt": "prettier --config .prettierrc --write 'src/**/*.js*'",
     "precommit": "lint-staged",
     "production": "NODE_ENV=production node lib/server/index.js",
-    "start": "NODE_ENV=development babel-node ./webpack/dev-server.js",
+    "start": "babel-node ./webpack/dev-server.js",
     "webpush": "node ./scripts/webpush_notify.js",
     "checktranslations": "node scripts/check_translations.js"
   },
--- a/src/app/ResolveRoute.js
+++ b/src/app/ResolveRoute.js
@@ -34,8 +34,8 @@ export default function resolveRoute(path) {
     if (path === '/support.html') {
         return { page: 'Support' };
     }
-    if (path === '/xss/test' && process.env.NODE_ENV === 'development') {
-        return { page: 'XSSTest' };
+    if (path === '/xss/test' && process.env.NODE_ENV !== 'production') {
+        return {page: 'XSSTest'};
     }
     if (path.match(/^\/tags\/?/)) {
         return { page: 'Tags' };
--- a/src/app/RootRoute.js
+++ b/src/app/RootRoute.js
@@ -34,10 +34,7 @@ export default {
             //require.ensure([], (require) => {
             cb(null, [require('app/components/pages/Support')]);
             //});
-        } else if (
-            route.page === 'XSSTest' &&
-            process.env.NODE_ENV === 'development'
-        ) {
+        } else if (route.page === 'XSSTest' && process.env.NODE_ENV !== 'production') {
             //require.ensure([], (require) => {
             cb(null, [require('app/components/pages/XSS')]);
             //});
--- a/src/app/components/pages/XSS.jsx
+++ b/src/app/components/pages/XSS.jsx
@@ -3,7 +3,7 @@ import MarkdownViewer from 'app/components/cards/MarkdownViewer';
 
 class XSS extends React.Component {
     render() {
-        if (!process.env.NODE_ENV === 'development') return <div />;
+        if (!process.env.NODE_ENV !== 'production') return <div />;
         let tests = xss.map((test, i) => (
             <div key={i}>
                 <h2>Test {i}</h2>
--- a/src/db/models/index.js
+++ b/src/db/models/index.js
@@ -31,7 +31,7 @@ Object.keys(db).forEach(function(modelName) {
 db.sequelize = sequelize;
 db.Sequelize = Sequelize;
 
-if (env === 'development') {
+if(env !== 'production') {
     // in dev, sync all table schema automatically for convenience
     sequelize.sync();
 }
--- a/src/server/api/oauth.js
+++ b/src/server/api/oauth.js
@@ -21,7 +21,7 @@ function logErrorAndRedirect(ctx, where, error) {
             ctx.req.headers['user-agent']
         }: ${msg}`
     );
-    if (process.env.NODE_ENV === 'development') console.log(error.stack);
+    if (process.env.NODE_ENV !== 'production') console.log(error.stack);
     ctx.flash = { alert: `${where} error: ${msg}` };
     ctx.redirect('/');
     return null;
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -259,7 +259,7 @@ app.use(
     )
 );
 // Proxy asset folder to webpack development server in development mode
-if (env === 'development') {
+if (env !== 'production') {
     const webpack_dev_port = process.env.PORT
         ? parseInt(process.env.PORT) + 1
         : 8081;
--- a/src/shared/UniversalRender.jsx
+++ b/src/shared/UniversalRender.jsx
@@ -200,7 +200,7 @@ const sagaMiddleware = createSagaMiddleware(
 
 let middleware;
 
-if (process.env.BROWSER && process.env.NODE_ENV === 'development') {
+if (process.env.BROWSER && process.env.NODE_ENV !== 'production') {
     const composeEnhancers =
         window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose; // eslint-disable-line no-underscore-dangle
     middleware = composeEnhancers(applyMiddleware(sagaMiddleware));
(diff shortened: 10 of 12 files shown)


sneak:
I don't know if this is the right way to do this.  The decision to default to prod was a deliberate one.

What is the benefit of doing it this way?  What do you mean by "multiple custom development environments"?

I think we should probably factor those tests into some sort of feature flag singleton, which has an .isProduction() and an .isDevelopment() set of methods.

jnordberg:
Before this the checks where not so deliberate, a mix of `if not development` and `if not production`. So with a node env that is not `development` or `production` the behaviour is undefined. With this we default to `development` and only use production vars in `production`.

What i mean by multiple custom enviroments is that the config module we are using resolves config files with `NODE_ENV` so you can have a config setup e.g. for your local test node and start with `NODE_ENV=mynode`. Another common usecase for a NODE_ENV that is not development or production is setting it to `testing` to setup the testing env during ci-testing.

goldibex:
Code potentially to be run in a browser shouldn't be testing values of environment variables, let alone NODE_ENV. I realize that modern JS build tools polyfill `process.env` in the browser environment, but application configuration should be encapsulated.

If the config module resolves config files with NODE_ENV that's fine, but one should then be testing a value in the config with `config.get('some.value.here')`, not in `process.env`. 

@sneak worth pointing out that the config module currently does not assume production environment in the absence of `NODE_ENV`.

sneak:
> Code potentially to be run in a browser shouldn't be testing values of environment variables, let alone NODE_ENV.

Why?  It's undefined AFAIR.  Shouldn't matter.

> @sneak worth pointing out that the config module currently does not assume production environment in the absence of NODE_ENV.

I think that's why we default to `NODE_ENV=production` in the `yarn run start` command.

goldibex:
> Why? It's undefined AFAIR. Shouldn't matter.

Consistency is key. Either we should just use `process.env` and drop the `config` module, or we should use `config` and drop the use of `process.env`. It reduces developer and ops cognitive burden, slims the JS bundle, and cuts onboarding time for new devs. My vote is to drop `process.env`.

Polyfilling `process.env` on the browser is not a great idea for a whole lot of reasons:
- It's a lie; there are plenty of Node constructs on `process` like `abort`, `argv`, `memoryUsage`, `send`, `nextTick`, `getuid`, etc., that have no meaningful analogues in the browser environment.
- it's a magic global. The developer has no idea where it came from, and build tools have to explicitly be told not to worry about it. The `config` module, on the other hand, is quite clear on that point: `import { config } from 'config'`.
- Just checking `process.env.THINGS` isn't resilient to mistakes like variable misspellings and the like. The `config` module explicitly checks to see if a configuration variable is defined in the schema and blows up when you ask for an unanticipated one. 

One of the unanticipated side effects of ES6 modules is that, because Node still doesn't support them, it's thrown the whole `browserify` model of "just use Node constructs everywhere" out the window. Mixing and matching magic globals and ES6 imports makes it much more difficult to comprehend and maintain code.

 I have a lot of experience playing the debugging game of _Whose Env Is It Anyway?_, where everything's made up and the variables don't matter. It invariably becomes a cause of tons of dev-hours wasted in debugging.

sneak:
BZZZZZZZZZZT.  Sounds great - let's move all this into the config module instead.  50 points for Ryan.

jnordberg:
Can we merge this in as a stepping stone towards that?

gl2748:
Looking at this now.

valzav:
We are using transform-inline-environment-variables babel plugin 
 (https://www.npmjs.com/package/babel-plugin-transform-inline-environment-variables) so all calls to env vars are being replaced with constansts in prod. This will always outperform config.get calls. I agree with @goldibex arguments in general, but we only use and check two env vars: BROWSER and NODE_ENV, so I would stick to process.env for those vars.

jnordberg:
I don’t think the performance of`config.get‘ calls is something that we need to consider

gl2748:
regarding performance, on the client side the config is static, so no get() methods etc... it's just an object.
https://github.com/lorenwest/node-config/wiki/Webpack-Usage

gl2748:
So to extend the usefulness of this work I have:
- Removed all instances of `process.env.NODE_ENV` from webpack, prefering the `NODE_ENV` as explicitly defined by our config.
- exposed `node-config` config options as defined in the `config` dir to the client side
  - Removed `src/app/client_config.js` preferring these options to be defined in the config object.
  - Remove `process.env.NODE_ENV` from the client side, preferring NODE_ENV as defined in the config, same rationale as described above.

What I have not done. - Touched the server side.
-  Therefore `process.env.BROWSER` still appears.
- This chunk looks like it could, one day be removed, and the config attributes it gets accessed directly across the SSR code base. https://github.com/steemit/condenser/blob/master/src/server/index.js#L11-L34.

config/default.json
```diff
@@ -99,5 +101,48 @@
   "steemd_connection_client": "wss://steemd.steemit.com",
   "steemd_connection_server": "wss://steemd.steemit.com",
   "chain_id": "0000000000000000000000000000000000000000000000000000000000000000",
-  "address_prefix": "STM"
+  "address_prefix": "STM",
+  "client": {
+    // sometimes it"s impossible to use html tags to style coin name, hence usage of _UPPERCASE modifier
```
goldibex: this is illegal in JSON, isn't it?

config/development.json
```diff
@@ -0,0 +1,4 @@
+{
+    "BABEL_ENV": "browser",
+    "KOA_BABEL_ENV": "server"
+}
```
goldibex: missing EOL

config/production.json
```diff
@@ -1 +1,3 @@
-{}
+{
+    "NODE_ENV": "production"
+}
```
goldibex: missing EOL

src/app/components/App.jsx
```diff
@@ -17,10 +17,14 @@ import MiniHeader from 'app/components/modules/MiniHeader';
...
 import PageViewsCounter from 'app/components/elements/PageViewsCounter';
 import {serverApiRecordEvent} from 'app/utils/ServerApiClient';
-import { APP_NAME, VESTING_TOKEN, LIQUID_TOKEN } from 'app/client_config';
 import {key_utils} from 'steem/lib/auth/ecc';
 import resolveRoute from 'app/ResolveRoute';
 import {VIEW_MODE_WHISTLE} from 'shared/constants';
+import config from 'config';
+
+const {
+    LIQUID_TOKEN
```
goldibex: what happened to APP_NAME and VESTING_TOKEN?

src/server/app_render.jsx
```diff
@@ -12,7 +12,8 @@ import {determineViewMode} from "../app/utils/Links";
 
 const path = require('path');
 const ROOT = path.join(__dirname, '../..');
-const DB_RECONNECT_TIMEOUT = process.env.NODE_ENV === 'development' ? 1000 * 60 * 60 : 1000 * 60 * 10;
+
+const DB_RECONNECT_TIMEOUT = process.env.NODE_ENV !== 'production' ? 1000 * 60 * 60 : 1000 * 60 * 10;
```
goldibex: shouldn't `DB_RECONNECT_TIMEOUT` just be defined directly in the configuration without the conditional expression?

config/default.json
gl2748: Github certainly thinks so. My guess would be that `node-config` is minifying it.  See first comment here: https://stackoverflow.com/a/4183018

src/app/components/App.jsx
gl2748: Not being used.

webpack/base.config.js
```diff
@@ -2,12 +2,17 @@ import path from 'path';
...
+import config from 'config';
+import fs from 'fs';
 
 const Webpack_isomorphic_tools_plugin = require('webpack-isomorphic-tools/plugin');
 const webpack_isomorphic_tools_plugin =
     new Webpack_isomorphic_tools_plugin(require('./webpack-isotools-config'))
         .development();
 
+// https://github.com/lorenwest/node-config/wiki/Webpack-Usage
+fs.writeFileSync(path.resolve(__dirname, '../tmp/client.json'), JSON.stringify(config))
```
goldibex: I understand the need for this but it seems very strange to me that a configuration file has side effects. Possibly this could be moved to a build step?

src/app/Main.js
```diff
@@ -10,7 +10,7 @@ import ConsoleExports from './utils/ConsoleExports';
 import {serverApiRecordEvent} from 'app/utils/ServerApiClient';
 import * as steem from 'steem';
 import {determineViewMode} from "app/utils/Links";
-
+import config from 'config';
```
goldibex: import { NODE_ENV } from `config`, no? or is `config` an `export default` object?

src/server/app_render.jsx
gl2748: yes. I'll do that as it isn't something that occurred to me.

src/app/components/App.jsx
goldibex: ok

gl2748:
@goldibex regarding your should-be-flat code smell, i would agree, i missed this but the docs state:
"If you have things in the config that you do not want to publish to the front end, you may consider making a specific client section of the config, and only publishing that key to the static json file, e.g. import { client } from 'config';"

src/app/Main.js
gl2748: no, you are correct.

config/default.json
jnordberg: The config module supports [TOML](https://github.com/toml-lang/toml) which is really nice for 
config files, we should switch to that

```toml
address_prefix = "STM"
[client]
APP_NAME = "Steemit"
# sometimes it"s impossible to use html tags to style coin name, hence usage of _UPPERCASE modifier
APP_NAME_UPPERCASE = "STEEMIT"
```

valzav:
@gl2748 :

> The entire config options as defined here: https://github.com/steemit/condenser/blob/master/config/default.json
> are available as a static object on the client side. We should ensure this does not expose any security vulnerabilities, the custom-environment-variables as defined here: https://github.com/steemit/condenser/blob/master/config/custom-environment-variables.json remain secret.

Is there any plan for that?

webpack/base.config.js
gl2748: Were you thinking of something as simple as abstracting the current approach with the `fs.writeFileSync` side effect abstracted out into a plugin: https://gist.github.com/gl2748/9f7a05190ef742c3943df5aae8d9885d ?

Or something that leverages webpacks' virtual filesystem ...
https://github.com/rmarscher/virtual-module-webpack-plugin

Regardless -  

I do think this is alright for now. I understand the concern but in the interest of getting this merged think this is OK ... `node-config` at the time of writing is not compatible with webpack and a hack makes sense... The above approaches could be the basis for a re-visit...

gl2748:
@valzav 
Regarding security, I don't think anything in `config/default.json` would be an issue. There is this line... 
https://github.com/steemit/condenser/blob/master/config/default.json#L80
But if that's an issue it should be purged from our version control ...

webpack/base.config.js
goldibex: let's please get it right. there isn't some massive rush here.

gl2748: Ok.

sneak:
it’s not an issue; it isn’t a secret. it’s a default for dev/tests. it is overridden in prod via env var.

bnchdrff:
https://stackoverflow.com/questions/30030031/passing-environment-dependent-variables-in-webpack#30061249

gl2748:
CI currently breaking, our mocha isn't sharing webpack config, therefore aliased modules are not being resolved in the mocha processes, suggest:
https://github.com/steemit/condenser/issues/2115

This would bring code under test into line with code as it is ultimately compiled, and ideally do away with `mocha.setup.js`

src/server/app_render.jsx
bnchdrff: @gl2748 ping

bnchdrff:

@gl2748 do you think this would be a quick & easy task to bring back up to date?

PR #1768, opened 2017-09-29
Primary navigation

pkattera:
I made some changes to the primary navigation and removed some elements. As I'm new to the condenser code base, please review to make sure I haven't broken anything. @valzav 

screen shot 2017-09-29 at 6 11 19 pm

– Increased primary navigation height on large screens (allows for a larger future logo)
– Removal of subnavigation (integrating it with the primary nav)
– Removing titles within the primary nav for people's names and topics 
   (future designs will place titles more clearly above the feed on the page)
– Renamed 'Submit a story' to 'Post' inside a button to place more weight on this element.
   
By removing the dynamic width elements from the navigation, it becomes less fragile and prone to overlapping elements when there are long topics or names.

### DIFF
--- a/src/app/components/App.scss
+++ b/src/app/components/App.scss
@@ -11,11 +11,11 @@
     }
     /* Medium only */
     @media screen and (min-width: 40em) and (max-width: 63.9375em) {
-        margin-top: 2.75rem;
+        margin-top: 3rem;
     }
     /* Large and above */
     @media screen and (min-width: 64.063em) {
-        margin-top: 2.75rem;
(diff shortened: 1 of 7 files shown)


sneak:
Hey @valzav remember when you were convinced that designers that can also make PRs were mythical? 😏 

Absolutely excellent work, @pkattera!  I'm deploying this to dev right now to test.

valzav:
@sneak, as far as I remember I said they are very rare, not mythical. Glad we found one..

PR #1777, opened 2017-10-03
add sorting to open orders table

### LINKED ISSUE #1757: open orders for the logged in user are not sorted in the market display
ideally the headers would be clickable to sort open orders

bnchdrff:
closes #1757 

will work on adding tests for helper fn & the reducer

### DIFF
--- a/src/app/components/pages/Market.jsx
+++ b/src/app/components/pages/Market.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { connect } from 'react-redux';
 //import Highcharts from 'highcharts';
+import classNames from 'classnames';
+import { Map } from 'immutable';
 
 import * as transactionActions from 'app/redux/TransactionReducer';
 import * as appActions from 'app/redux/AppReducer';
@@ -26,10 +28,12 @@ class Market extends React.Component {
     static propTypes = {
         orderbook: React.PropTypes.object,
         open_orders: React.PropTypes.array,
+        open_orders_sort: React.PropTypes.instanceOf(Map),
         ticker: React.PropTypes.object,
         // redux PropTypes
         placeOrder: React.PropTypes.func.isRequired,
         user: React.PropTypes.string,
+        feed: React.PropTypes.instanceOf(Map),
     };
 
     constructor(props) {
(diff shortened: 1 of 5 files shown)


bnchdrff:
waiting until redux-modules is removed to continue work on this

bnchdrff:
i just refactored this PR after our removal of redux-modules, and it's ready for review finally!

bnchdrff:
also added tests

src/app/components/pages/Market.jsx
```diff
@@ -26,6 +27,7 @@ class Market extends React.Component {
     static propTypes = {
         orderbook: React.PropTypes.object,
         open_orders: React.PropTypes.array,
+        open_orders_sort: React.PropTypes.object,
```
gl2748: Do we want to start describing shape in our proptypes? this can give improved errors if something breaks or changes? Not seen it elsewhere in our codebase so a general q.

bnchdrff: that's a grand idea. i'll add it here :)

src/app/components/pages/Market.jsx
```diff
@@ -295,6 +306,7 @@ class Market extends React.Component {
 
         let account = this.props.account ? this.props.account.toJS() : null;
         let open_orders = this.props.open_orders;
+        let open_orders_sort = this.props.open_orders_sort;
```
gl2748: Suggest destructing `this.props` too.

src/app/components/pages/Market.jsx
```diff
@@ -911,7 +986,8 @@ module.exports = {
                 account: state.global.getIn(['accounts', username]),
                 history: state.market.get('history'),
                 user: username,
-                feed: state.global.get('feed_price').toJS(),
+                feed: state.global.get('feed_price'),
```
bnchdrff: @gl2748 i ended up changing the `open_orders_sort` prop as well as one of the other props `feed` to pass as an immutable rather than a plain object, so shouldComponentUpdate can more efficiently compare them -- unfortunately we can't define a proptypes shape for immutables -- we should eventually use a library like https://github.com/HurricaneJames/react-immutable-proptypes to do this

src/app/components/pages/Market.jsx
bnchdrff: done!

gl2748:
# after
### makes columns for 'open orders' on markets page sort-able.
sortable

src/app/components/pages/Market.jsx
gl2748: ok cool.

PR #1779, opened 2017-10-03
Fix all i18n and linting warnings

valzav:


scripts/check_translations.js
```diff
@@ -76,12 +80,16 @@ function checkKeys(translations, used_keys) {
     for (const lang in translations) {
         const lang_keys = translations[lang];
         for (const key in used_keys) {
-            if (!lang_keys[key]) console.warn('Translation key not found: ', lang, key);
-            errors_counter += 1;
+            if (!lang_keys[key]) {
+                console.warn('Translation key not found: ', lang, key);
+                errors_counter += 1;
```
jnordberg: Logic is changed here

scripts/check_translations.js
```diff
@@ -76,12 +80,16 @@ function checkKeys(translations, used_keys) {
...
+            }
         }
         for (const key in lang_keys) {
-            if (!used_keys[key]) console.warn('Unused translation: ', lang, key);
-            errors_counter += 1;
+            if (!used_keys[key]) {
+                console.warn('Unused translation: ', lang, key);
+                errors_counter += 1;
```
jnordberg: Also here

valzav: right, but check_translations.js is only run in dev env

PR #1782, opened 2017-10-04
Signup button tweak

jnordberg:
Free Money => Free Steem

screen shot 2017-10-04 at 19 58 42

### DIFF
--- a/src/app/components/modules/LoginForm.scss
+++ b/src/app/components/modules/LoginForm.scss
@@ -7,18 +7,18 @@
   form {
     margin-top: 1.5rem;
   }
-  .button.sign-up {
+  .sign-up {
+    .button {
       background-color: #4078C0 !important;
       padding-top: 1rem;
       padding-bottom: 1rem;
       font-size: 1.2rem;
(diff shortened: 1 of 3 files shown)


src/app/components/modules/LoginForm.scss
```diff
@@ -7,18 +7,18 @@
   form {
     margin-top: 1.5rem;
   }
-  .button.sign-up {
+  .sign-up {
+    .button {
```
relativityboy: Nitpick - wondering about the rational for making this a class `.button` & not just an element type `button`, particularly when the class name IS the element name (lots of possibility for collision as css changes over time)

jnordberg: left it as-is because I figured it was inheriting from some global .button class

jnordberg:
We want to be consistent in our message, the signup button under posts currently looks like this:

screen shot 2017-10-04 at 20 19 08

And the banner when landing on steemit.com:

screen shot 2017-10-04 at 20 20 32

cc @nedsteem

PR #1788, opened 2017-10-05
instruct eslint to ignore files ending in test.js

bnchdrff:
test files will throw all sorts of warnings for globals and such. if someone is so motivated we could develop a separate ruleset for tests. i don't think that's necessary.


sneak:
Approving this, but wouldn’t a better fix be to disable those specific warnings for the files it complains about, at the top of those files? I think we should still be getting warnings for tests - they are code that we have to maintain, too.

bnchdrff:
ya know that's probably a better idea, and i think i do that in other projects anyways

Issue #1789, opened 2017-10-05
Use /u/<username>/avatar for user pics


relativityboy:
There's a new feature coming up in imagehoster 

https://github.com/steemit/imagehoster/pull/36

We should use this for user avatars

PR #1790, opened 2017-10-05
1789 - update Userpic to use /u/${username}/avatar

relativityboy:
https://github.com/steemit/condenser/issues/1789

This should be ready to go. Marking as WIP until the avatar code from /imagehoster is in prod


src/app/utils/ProxifyUrl.js
```diff
@@ -8,9 +8,14 @@
  * <proxy>/{int}x{int}/[<proxy>/{int}x{int}/]<proxy>/<file url>
  * @type {RegExp}
  */
-const rProxyDomain = /^http(s)?:\/\/steemit(dev|stage)?images.com\//g;
-const rProxyDomainsDimensions = /http(s)?:\/\/steemit(dev|stage)?images.com\/([0-9]+x[0-9]+)\//g;
-const NATURAL_SIZE = '0x0/';
+const rProxyDomain = /^http(s)?:\/\/steemit(dev|stage)?images.com\//g
```
sneak: Let’s make a URL parsing class that takes a string and has methods like isImageProxyUrl() and isValidURL() (from a better url validation library that isn’t maintained by us) and imageProxyDimensions() and suchlike, so that all URL-parsing regex are in a single place.

relativityboy: That sounds good-ish, but solving it all the way is beyond the bounds of this issue.

The only functional addition to ProxifyUrl is the imageProxy. It was added to reduce direct dependence on window.$STM_Config (a terrible thing) and to begin the process of creating that "URL parsing class".

What we have here is a step in the right direction; further changes are not warranted for this merge.

src/app/components/elements/AuthorDropdown.jsx
```diff
@@ -16,7 +16,7 @@ const AuthorDropdown = (props) => {
             <div className = "Author__container">
                 <div className="Author__dropdown">
                     <Link to={'/@' + props.author}>
-                        <Userpic account={props.author} width="75" height="75" />
```
roadscape: IIRC - main reason for passing size to Userpic is because it needs to know if it should request hi- or lo-res img from the server

roadscape: I see.. this functionality is removed entirely. Hmm. The high resolution image is bound to 320x320, whereas low resolution image is 120x120. Assuming the source image is a square, the high-res image will be ~7x bigger. My concern is that on some pages with 10's or 100's of (small) avatars this could really add up.

src/app/utils/ProxifyUrl.js
```diff
@@ -21,21 +26,21 @@ const NATURAL_SIZE = '0x0/';
  * @returns string
  */
 export default (url, dimensions = false) => {
-    const proxyList = url.match(rProxyDomainsDimensions);
-    let respUrl = url;
+    const proxyList = url.match(rProxyDomainsDimensions)
+    let respUrl = url
```
roadscape: are all these semicolon adjustments for eslint warnings?

relativityboy: To reduce visual clutter, and to (I hope eventually) eliminate semicolons from our codebase. There's a lot of variance, even within a given file. I decided to pic one way, and clean up where I touch files if at all possible. I chose the [standardjs](https://standardjs.com/) way as it's somewhat popular in the React community.

src/app/components/elements/AuthorDropdown.jsx
relativityboy: With Userpic rendering only urls from `/u/${username}/avatar` we don't need to worry about sizing, as [Imagehoster takes care of it](https://github.com/steemit/imagehoster/pull/36/files#diff-ef4e02d1f3af008c9df50485227330ccR18).

src/app/utils/ProxifyUrl.js
valzav: hmm.. actually the code style guide line requires semicolons but it was up to developers to either use them or not, now when we have automated checks I think it's better to tieten things up and stick to the rules whenever it's possible

relativityboy: _Mumble mumble grumble..._

src/app/components/elements/AuthorDropdown.jsx
jnordberg: Imagehoster only serves 128x128 avatars as it is now

src/app/components/elements/Userpic.jsx
```diff
@@ -1,51 +1,43 @@
 import React, {Component, PropTypes} from 'react';
 import {connect} from 'react-redux';
 import shouldComponentUpdate from 'app/utils/shouldComponentUpdate';
-import proxifyImageUrl from 'app/utils/ProxifyUrl';
+import { imageProxy } from 'app/utils/ProxifyUrl';
```
jnordberg: Following the surrounding code style this should not have spaces

src/app/components/elements/Userpic.jsx
```diff
@@ -1,51 +1,43 @@
...
         try {
             const md = JSON.parse(json_metadata);
-            if(/^(https?:)\/\//.test(md.profile.profile_image)) {
-                const size = (width && width > 48)? '320x320' : '120x120';
-                url = proxifyImageUrl(md.profile.profile_image, size);
-            }
-        } catch (e) {}
-        if(!url) {
-            if(hideIfDefault) {
+            if(!/^(https?:)\/\//.test(md.profile.profile_image) && hideIfDefault) {
```
jnordberg: Why is this check still needed?

src/app/utils/ProxifyUrl.js
```diff
@@ -12,6 +12,8 @@ const rProxyDomain = /^http(s)?:\/\/steemit(dev|stage)?images.com\//g;
 const rProxyDomainsDimensions = /http(s)?:\/\/steemit(dev|stage)?images.com\/([0-9]+x[0-9]+)\//g;
 const NATURAL_SIZE = '0x0/';
 
+export const imageProxy = () => $STM_Config.img_proxy_prefix;
```
jnordberg: This should be a constant

src/app/components/elements/Userpic.jsx
```diff
@@ -1,51 +1,43 @@
...
+    account: PropTypes.string.isRequired
+}
+
 export default connect(
     (state, ownProps) => {
-        const {account, width, height, hideIfDefault} = ownProps
+        const {account, hideIfDefault} = ownProps
         return {
+            account,
             json_metadata: state.global.getIn(['accounts', account, 'json_metadata']),
```
jnordberg: This should be removed as well

src/app/components/elements/AuthorDropdown.jsx
relativityboy: It's a nice compromise size. A visual check of the only place the larger image was used looked good.

Also since the /avatar urls may point @ different images, we don't really want different sizes of profile image being cached for indefinitely long periods of time.

src/app/components/elements/Userpic.jsx
relativityboy: The code indicated there are situations where the user does not have a profile image set, and we do not want to show the default image. - There is at least one instance in condenser where that path could have been triggered (I think it's the user's page)

src/app/components/elements/Userpic.jsx
relativityboy: Based on my above comment, you're incorrect.
> The code indicated there are situations where the user does not have a profile image set, and we do not want to show the default image....

src/app/utils/ProxifyUrl.js
relativityboy: That is not an option.

src/app/components/elements/Userpic.jsx
relativityboy: A nice stylistic catch. My understanding is that we're using Airbnb style guidelines, which indicate _everywhere_ should have spaces for module imports of this type. https://github.com/airbnb/javascript#modules
I'll add them for consistency.

src/app/components/elements/Userpic.jsx
jnordberg: Hm, why are we doing this? I thought it was so we didn’t have to make user lookups in condenser for every profile pic. Now we just loose the ability to dynamically size them and double the amount of calls to steemd

src/app/utils/ProxifyUrl.js
jnordberg: Is it injected after the app code is loaded? Could make it into a `getAvatarUrl(account)` that would justify its existence:)

src/app/components/elements/Userpic.jsx
roadscape: This is only used for user page headers -- the default avatar looks pretty bad on the blueish gradient.

@jnordberg - I was under the impression that wherever the avatar image is used does not require any extra calls to steemd. This is because get_state includes author json for all accounts in a discussion. (However, now the image_hoster will be making a lot of extra get_account calls.. has the response time been measured? We may find that it needs to cache results for several minutes...)

src/app/utils/ProxifyUrl.js
roadscape: Is this compatible w/ the way the ProxifyUrl tests inject this global variable?

src/app/components/elements/AuthorDropdown.jsx
roadscape: We may want to bump it up to 150x150 or so. Existing sizes were chosen w/ retina sizes and some room for growth in mind. I think currently the max size we render is 75x75, so to look good on retina should 2x that.

A single size should work for now... but we'll need to revisit this if we ever need anything bigger than small or med-small image. Ideally the imagehoster would accept size params.

src/app/components/elements/Userpic.jsx
roadscape: We could axe `hideIfDefault` if it's ok that default profiles look a bit uglier. Hell, maybe that would encourage more users to customize theirs. Maybe check w/ Pon. But this feature doesn't cause any extra lookups, so the main benefit is simply less LOC.

src/app/components/elements/AuthorDropdown.jsx
relativityboy: Imagehoster accepting size params or a size profile (sm/med/lg/full) for avatars is a great idea.

src/app/components/elements/Userpic.jsx
relativityboy: @roadscape - caching of avatars will be a fantastic feature for imagehoster, and I think it's part of the intended roadmap; [definitely worth logging another issue](https://github.com/steemit/imagehoster/issues/38).

src/app/components/elements/Userpic.jsx
relativityboy: *Scope Creep* The intent of this issue is to get us using the avatar url, not to change the way the ui presents to the user. Let's leave that to a different issue.

src/app/utils/ProxifyUrl.js
relativityboy: @roadscape Tests are passing. :)

src/app/components/elements/AuthorDropdown.jsx
relativityboy: @roadscape - [I've created an issue to track sizing](https://github.com/steemit/imagehoster/issues/39). Please comment there if you think it could use revision. :)

roadscape: Ok, maybe we should keep the width/height param handling intact in Userpic if we want to leverage it later

sneak:
I THINK THE BIKESHED SHOULD BE BLUE

PR #1797, opened 2017-10-07
Upgrade to webpack3

valzav:
based on webpack3 branch but it doesn't include redux-modules refactoring


src/app/RootRoute.js
```diff
@@ -1,10 +1,10 @@
-// polyfill webpack require.ensure
-//if (typeof require.ensure !== 'function') require.ensure = (d, c) => c(require);
-
 import App from 'app/components/App';
 import PostsIndex from 'app/components/pages/PostsIndex';
 import resolveRoute from './ResolveRoute';
 
+// polyfill webpack require.ensure
+if (typeof require.ensure !== 'function') require.ensure = (d, c) => c(require);
```
goldibex: why are we doing this? condenser uses es6 modules. `require` shouldn't even be defined in the browser context.

src/app/RootRoute.js
```diff
@@ -79,19 +79,21 @@ export default {
             cb(null, [require('app/components/pages/Witnesses')]);
```
goldibex: again, please walk me through the use of `require` and `require.ensure` in this context

src/app/components/pages/Welcome.jsx
```diff
@@ -8,11 +8,11 @@ class Welcome extends React.Component {
                 <div className="column large-8 medium-10 small-12">
                     <div className="Welcome__banner">
                         <div className="Welcome__welcome">Welcome to</div>
-                        <img src={require(`app/assets/images/welcome.jpg`)} />
+                        <img src={require('app/assets/images/welcome.jpg')} />
```
goldibex: the name `require` for anything other than the commonjs module importer confuses me. could we think of a better name for this helper?

webpack/dev.config.js
```diff
@@ -1,69 +1,40 @@
...
-                assets: true,
-                colors: true,
-                version: false,
-                hash: false,
-                timings: true,
-                chunks: false,
-                chunkModules: false
-            }
-        }
+    ...baseConfig,
```
goldibex: take that, `Object.assign`!

I really love the object spread operator, it's the best thing in Javascript since `=>`

goldibex:
back to you @valzav, and thanks so much for taking point on this

src/app/RootRoute.js
valzav: This part is being rendered both server and client side isomorphically.
In browser require.ensure is provided by webpack but when it renders on the server we just make it to use regular require.

src/app/RootRoute.js
valzav: When require is wrapped with require.ensure it will make webpack to dynamically load a component and it's deps

src/app/components/pages/Welcome.jsx
valzav: This is webpack's magic :)
To make it work server side we use webpack-isomorphic-tools

src/app/RootRoute.js
goldibex: Are we very sure this is a dynamic usage? commonjs `require` is synchronous, meaning that the dependency has to be available within the current tick of the event loop. Why not pull that `require` out as an ES6 `import` at the top, thus above we would have

```ES6
import WitnessComponent from 'app/components/pages/Witnesses';
```

and at line 79, we'd have
```ES6
cb(null, [WitnessComponent]);
```

My point is that this isn't working like a dynamic route loader that checks whether it's on a route like `/@someUser` and then loads the code at runtime, saving those bytes from the initial bundle and making time to first content shorter. That `require` is still being parsed out at compile time and the associated code added to the bundle. As a result, it isn't clear to me why we want the added complexity of a `require` as opposed to another standard ES6 `import`.

The tell is that CommonJS `require` is synchronous, whereas the ES6 module loader, `System`, returns a promise with the module once it's loaded.

If I'm wrong or misunderstanding any of these points, please let me know!

src/app/components/pages/Welcome.jsx
goldibex: Sorry, I'm confused. Can you point me to the definition of this `require` function and explain the magic a bit? It seems to me as though we just need a path to where the image is being hosted, right? Or does this inject like a Base64-encoded version of the image inline in the HTML?

I'm pretty familiar with how JS building and bundling works, I've contributed to both AMD and SystemJS.  I've never seen a `require` used from within a JSX template.

valzav: It can do both - inline and hosted images.
Webpack loads resources via the loaders, take a look at webpack/base.config.js:58 - it defines the rule for images: {test: /\.(jpe?g|png)/, use: 'url-loader?limit=4096'}, so if image size is less than 4kb it will use inline version if larger it will add src with corresponding path. Images embedded with links are added to server's static assets.

src/app/RootRoute.js
valzav: It's working like a dynamic route loader (code splitting in webpack's terminology). See https://webpack.js.org/api/module-methods/#require-ensure for details
It really decreases the bundle's size, for example loading react rte editor only when user need  to compose a post allows us to save hundreds of kilobytes in the main bundle file.
To see how it works watch the Network tab and navigate to /submit.html

Issue #1800, opened 2017-10-09
Universal presentation of '/' route on server 


relativityboy:
right now the server redirects `/` to `/@username/feed` on logged in users we want the same page for `/` to be served to everyone and the client detects if it's logged in or not and redirects itself to the `/@username/feed` route

changes: 1) server doesn't redirect on /, 2) client handles redirecting itself when logged in and / is loaded

*this may mean that the server needs a route to let the client know if it's logged in or not via session cookie*

relativityboy:
This branch.
https://github.com/steemit/condenser/tree/1800-kill-slash-redir-on-login

plink01001:
@relativityboy - two PRs have been merged for this. Can the issue be closed?

PR #1801, opened 2017-10-09
1800 eliminate redirect

relativityboy:
https://github.com/steemit/condenser/issues/1800

This addresses eliminating the server redirect portion of 1800. *1) server doesn't redirect on /,*

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -81,15 +81,7 @@ app.use(function* (next) {
         this.body = {status: 'ok'};
         return;
     }
-
-    /*
-    // redirect to home page/feed if known account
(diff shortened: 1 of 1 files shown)


src/server/server.js
```diff
@@ -82,12 +82,14 @@ app.use(function* (next) {
         return;
     }
 
+    /*
```
goldibex: why didn't you just remove the code outright?

relativityboy: I thought I'd be doing more before pulling the wip tag.

sneak:
This should only be done when the clientside fix for not having the redirect is integrated.  Until then, this redirect should stay in master.

PR #1808, opened 2017-10-10
make claim button inactive forever after click

### LINKED ISSUE #1311: Disable claim reward button during broadcast
From @ausbitbank  

> [I] request some sort of visual feedback that the claim button has been pressed. Right now I click it and a few seconds pass before it dissapears, impatient double clicking gives an error message - a simple animation or colour change to recognise that it's been clicked would stop this.

bnchdrff:
fixes #1311

this is a naive and quick solution that does not rely on a success event and requires a reload to reset the state

is it necessary to reset this state? how long will it be before the next time a user would get the option to claim again? is it likely they'll have the same browser window open?


sneak:
I would like it better if it morphed into a spinner when clicked

goldibex:
this has been approved, @bnchdrff, please merge and deploy

PR #1810, opened 2017-10-10
Move offchain state to separate api call

jnordberg:
Hooks in to the (deprecated?) [iso module](https://github.com/goatslacker/iso) we use to serialize state and drops the offchain data, loaded again via api call on the client side.


src/server/api/general.js
```diff
@@ -44,6 +45,27 @@ export default function useGeneralApi(app) {
...
+        }
+        const offchain = {
+            csrf: ctx.csrf,
+            flash: ctx.flash,
+            new_visit: ctx.session.new_visit,
+            account: ctx.session.a,
+            config: $STM_Config,
+            uid: ctx.session.uid,
+            serverBusy: false,
+            signup_bonus: '$1', // TODO: don't hardcode this
```
relativityboy: Don't hardcode this...

relativityboy:
I like this. @roadscape @valzav.

src/server/api/general.js
jnordberg: thx cpt. obvious :D 

@valzav / @roadscape do you know if `signup_bonus` is still used, would be nice if it can be removed

roadscape: The variable is still referenced and even passed to translator, but as far as I can tell the translations are no longer making use of it. The amount of STEEM rewarded to new users is low right now and I don't see it increasing since we rely on delegation now. Giving more directly on signup just encourages more abuse.

jnordberg:
@roadscape @valzav Should be good to go

src/server/server.js
```diff
@@ -49,6 +49,14 @@ const numProcesses = process.env.NUM_PROCESSES || os.cpus().length;
 
 app.use(requestTime(numProcesses));
 
+// drop set-cookie headers for everything but state api
+app.use(function*(next) {
+    yield* next;
+    if (this.request.url !== '/api/v1/state') {
```
valzav: Please add exceptions for user sign up and account recovery routes. Look for them in account_recovery.js and enter_confirm_email/phone.js

jnordberg: Good catch, I enabled set-cookie headers on everything that looks like it could be an entry point now

goldibex:
@valzav anything else before you can sign off?

bnchdrff:
is this PR still relevant & desired @jnordberg @valzav

jnordberg:
Yes & Yes :)

bnchdrff:
sweet... i can take on rebasing it if nobody else is available!

src/app/components/App.scss
```diff
@@ -23,17 +23,30 @@
...
+  }
+}
+
 .welcomeWrapper {
   margin-top: -1rem;
   padding-bottom: 1rem;
 }
 .welcomeBanner {
-  padding: 2.35rem 0 4em;
+  padding: 2.35em 0 4em;
```
bnchdrff: why these css changes?

PR #1817, opened 2017-10-15
remove unused implementation of componentWillReceiveProps

bnchdrff:
whatever code is in there is not even being executed! any need for this setState anymore?

### DIFF
--- a/src/app/components/App.jsx
+++ b/src/app/components/App.jsx
@@ -55,13 +55,6 @@ class App extends React.Component {
         }
     }
 
-    componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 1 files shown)


bnchdrff:
for reference, here's where the first call was added https://github.com/steemit/condenser/commit/0c0e70bd#diff-fbf810c4f5cc4809a93b9272c9cb75a3L41

and the second: https://github.com/steemit/condenser/commit/7e2e8992#diff-fbf810c4f5cc4809a93b9272c9cb75a3R64

goldibex:
please merge and deploy @bnchdrff

PR #1821, opened 2017-10-17
add transform-inline-environment-variables babel plugin

valzav:
In my tests this simple change shows 46% SRR performance boost

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "description": "steemit.com is the koa web server & middleware and react.js in-browser code for the world's first blockchain content + social media monetization platform!",
   "main": "index.js",
   "scripts": {
-    "build": "NODE_ENV=production ./node_modules/babel-cli/bin/babel-node.js ./node_modules/.bin/webpack --config ./webpack/prod.config.js; rm -rf ./lib; babel src --out-dir lib -Dq",
+    "build": "NODE_ENV=production ./node_modules/babel-cli/bin/babel-node.js ./node_modules/.bin/webpack --config ./webpack/prod.config.js; rm -rf ./lib; NODE_ENV=production babel --plugins transform-runtime,transform-inline-environment-variables src --out-dir lib -Dq",
     "mocha": "NODE_ENV=test mocha ./mocha.setup.js",
     "test": "npm run mocha -- src/app/**/*.test.js",
     "test:watch:all": "npm test -- --watch --watch-extensions jsx",
@@ -31,6 +31,7 @@
     "babel-loader": "^7.1.2",
     "babel-plugin-react-intl": "^2.2.0",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
+    "babel-plugin-transform-inline-environment-variables": "^0.2.0",
     "babel-plugin-transform-runtime": "^6.15.0",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -7,7 +7,10 @@ const ROOT = path.join(__dirname, '../..');
 
 // Tell `require` calls to look into `/app` also
 // it will avoid `../../../../../` require strings
+
+// use Object.assign to bypass transform-inline-environment-variables-babel-plugin (process.env.NODE_PATH= will not work)
 Object.assign(process.env, {NODE_PATH: path.resolve(__dirname, '..')});
+
 require('module').Module._initPaths();
 
 // Load Intl polyfill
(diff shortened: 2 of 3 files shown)


src/server/index.js
```diff
@@ -2,14 +2,12 @@ import config from 'config';
...
 
-delete process.env.BROWSER;
-
 const path = require('path');
 const ROOT = path.join(__dirname, '../..');
 
 // Tell `require` calls to look into `/app` also
 // it will avoid `../../../../../` require strings
-process.env.NODE_PATH = path.resolve(__dirname, '..');
+Object.assign(process.env, {NODE_PATH: path.resolve(__dirname, '..')});
```
goldibex: why the Object.assign rather than the direct assignment?

src/server/index.js
```diff
@@ -2,14 +2,12 @@ import config from 'config';
 
 import * as steem from 'steem';
 
-delete process.env.BROWSER;
```
goldibex: :+1: i wish they'd never put `delete` in Javascript

package.json
```diff
@@ -8,7 +8,7 @@
   "description": "steemit.com is the koa web server & middleware and react.js in-browser code for the world's first blockchain content + social media monetization platform!",
   "main": "index.js",
   "scripts": {
-    "build": "NODE_ENV=production ./node_modules/babel-cli/bin/babel-node.js ./node_modules/.bin/webpack --config ./webpack/prod.config.js; rm -rf ./lib; babel src --out-dir lib -Dq",
+    "build": "NODE_ENV=production ./node_modules/babel-cli/bin/babel-node.js ./node_modules/.bin/webpack --config ./webpack/prod.config.js; rm -rf ./lib; NODE_ENV=production babel --plugins transform-runtime,transform-inline-environment-variables src --out-dir lib -Dq",
```
goldibex: Not your code I know, but some questions:

- this could just be `babel-node` without the long path. and why do we need to invoke the webpack binary via `babel-node`? using `babel-node` in production is an antipattern.

src/server/index.js
valzav: Because the plugin replaces both process.env.NODE_PATH and process.env['NODE_PATH'], Object.assign is the only one way I found that is unaffected

relativityboy: It would be worth adding a comment about that above this change. You could save someone time next year.

relativityboy: @valzav ?

package.json
valzav: webpack config uses some fancy est6 syntax, probably we can refactor it to use plain js

valzav: What babel-node without long path would it use - global babel-node or the one that is in node_modules?

src/server/index.js
valzav: @relativityboy good point, I've added a comment

relativityboy: Thanks.

package.json
goldibex:
```
➜ dev mkdir foo
➜ dev cd foo
➜  foo npm install -g babel-cli
[snip]
+ babel-cli@6.26.0
added 233 packages in 49.679s
➜  foo which babel-node
/Users/goldibex/Library/Homebrew/bin/babel-node
➜  foo echo 'echo "BOOM" && exit 1' > /Users/goldibex/Library/Homebrew/bin/babel-node
➜  foo babel-node
BOOM
➜  foo npm init
[snip]
➜  foo npm install --save babel-cli
[snip]
➜  foo cat package.json | jq '.scripts.boom = "babel-node"' | tee package.json
[snip]
➜  foo npm run boom

> foo@1.0.0 boom /Users/goldibex/Development/steemit/foo
> babel-node

> 
```

goldibex: Node >= 8 supports object splats, so the only problem is the import/export syntax. 

Just so you know I'm not blowing smoke, check the output of `time` when babel-node is involved vs. when it isn't:

```
yarn build  47.55s user 1.89s system 112% cpu 43.801 total
yarn build  45.33s user 1.51s system 116% cpu 40.179 total
```

Shaves three seconds off the production build, well worth the effort.

goldibex:
@valzav any questions about the changes needed?

valzav:
@goldibex, the scope of this PR is to add transform-inline-environment-variables-babel-plugin, are there any changes needed that are not beyond this scope?

package.json
valzav: It doesn't use babel-node in production, `build` is pre-production step

goldibex: It's in the production build process, which is already pretty slow. It's an easy win for CI.

goldibex:
You're right about scope, so I'll sign off now.

package.json
valzav: @goldibex, I've created a separate github issue for this https://github.com/steemit/condenser/issues/1824

Issue #807, opened 2016-12-08
meta-discussion over change to URL routes


sneak:
I propose the following changes to the main URL routes:

(variables are prefixed with `$` sigil to distinguish from byte literals)

`/@$username` becomes `/$username`

`/$category/@$author/$permaslug` becomes `/$author/$permaslug`

(old `/$category/@$author/$permaslug` links should redirect to the new scheme)

current `/@$username/(.*)` routes get moved to `/$username/meta/$1` to distinguish from permalinks to articles

current `/$sortmode/$tagname` becomes `/headline/$sortmode/$tagname` (`/` being implicit `/headline/trending`).

`/s/(.*)` reserved for steemit.com specific pages such as help and documentation, the tag list (currently at the bizarre `/tags.html/created` route), static assets, et c.

I want our permalinks to be beautiful and concise.  The `@` is noise and the top tag doesn't need to be in the URL.

Thoughts?

sneak:
@valzav @roadscape @bytemaster @mcifani Your feedback is requested.

mcifani:
Definately in favor of the simplified routes suggested. Removing all the .html was talked about prior and good.  The reserved /s/ seems also clean for internal pages.  @sneak

roadscape:
`/@name` is an elegant (albeit non-compliant) way of namespacing.. better than `/~name`or `/u/name`, short of not using a ns at all, though this creates other conflicts.

Here's a potential layout without using `@`s:


```
Post:         /<community>/<user>/<permlink>
User:         /<user>
User meta:    /<user>/<tab>
Tag:          /tags/<tag>
Tag, sorted:  /tags/<tag>?sort=<sort>
Docs:         /docs/<page>
Users:        /users/<action>          (create, recover, etc)
```

Basically, a URL with 3 segments is always a post. A URL with 1 or 2 segments always starts with a username (some of which have been reserved for internal use).

Important to figure out what to do with communities vs. tags.. IMO communities would almost entirely replace them anyway (reddit), and tags could still be accessed via search (twitter). Tags will be a mess as they are not moderated, so they'd need to be combined with a relevance search anyway. Moderated (or even just *branded*) spaces will be the source for high SNR. (IMO category-based communities could work quite well even without blockchain-level consensus support)

sneak:
Settled on C from https://github.com/steemit/steempunks/wiki/URL-Routes

clayop:
[...]

PR #1826, opened 2017-10-19
migrate to new url scheme

valzav:
https://github.com/steemit/condenser/issues/807 <- contributes to this issue.


.eslintrc
```diff
@@ -62,6 +62,7 @@ find node_modules/eslint-config-airbnb -name '*.js'|xargs sed -i "s/': 2/': 1/"
     "jsx-a11y/anchor-is-valid": ["warn", { "aspects": ["invalidHref"] }],
     "import/no-extraneous-dependencies": 0,
     "import/no-unresolved": 0,
-    "import/extensions": 0
+    "import/extensions": 0,
+    "global-require": 0
```
sneak: what does disabling global-require do?

valzav: it allows to "require" a module inside a function or an expression, not only globally in the top of a source file

src/app/RootRoute.js
```diff
@@ -98,8 +98,6 @@ export default {
             //require.ensure([], (require) => {
             cb(null, [require('app/components/pages/PostPage')]);
```
bnchdrff: is there an advantage to doing the inline require here, rather than importing the component at the top like we're doing with `PostsIndex`?

relativityboy:
@valzav - jnordberg's got a good point there. Simple code is much better over the long haul. What are your thoughts?

src/app/Main.js
```diff
@@ -93,7 +94,7 @@ function runApp(initial_state) {
     initial_state.app.viewMode = determineViewMode(window.location.search);
 
     const location = `${window.location.pathname}${window.location.search}${window.location.hash}`;
-    universalRender({history, location, initial_state})
+    universalRender({history, location, initial_state, setStore })
```
bnchdrff: why do we pass setStore off to universalRender in this scope?

src/app/components/modules/Header.jsx
```diff
@@ -39,7 +37,7 @@ class Header extends React.Component {
         if (nextProps.location.pathname !== this.props.location.pathname) {
             const route = resolveRoute(nextProps.location.pathname);
             if (route && route.page === 'PostsIndex' && route.params && route.params.length > 0) {
-                const sort_order = route.params[0] !== 'home' ? route.params[0] : null;
+                const sort_order = route.params[1] !== 'home' ? route.params[1] : null;
                 if (sort_order) window.last_sort_order = this.last_sort_order = sort_order;
```
bnchdrff: no need to multi-assign

also... why are we using window here? why not use redux state? window.last_sort_order is used in TagList.jsx

src/shared/UniversalRender.jsx
```diff
@@ -156,6 +153,7 @@ async function universalRender({location, initial_state, offchain, ErrorPage, ta
 
     if (process.env.BROWSER) {
         const store = createStore(rootReducer, initial_state, middleware);
+        // setStore(store);
```
bnchdrff: setStore is not being used by universalRender; what's going on here?

src/app/components/elements/ChangePassword.jsx
```diff
@@ -72,7 +73,7 @@ class ChangePassword extends React.Component {
             if(onClose) onClose()
             if(resetForm) resetForm()
             notify('Password Updated')
-            window.location = `/login.html#account=${accountName}&msg=passwordupdated`;
+            window.location = pathTo.login() + `#account=${accountName}&msg=passwordupdated`;
```
bnchdrff: instead of `window.location` should we use [react-router-redux's `push` action](https://github.com/reactjs/react-router-redux#pushlocation-replacelocation-gonumber-goback-goforward)?

src/app/RootRoute.js
valzav: Yes, it gives us an option to use webpack's dynamic loading, as it's already done for the market page - all it requires is to uncomment require.ensure block.

src/app/components/elements/ChangePassword.jsx
valzav: Yes good point, I'll change it..

src/app/components/modules/Header.jsx
valzav: Hmm.. don't remember exactly why I used window here, probably it was even before I've started using redux.

src/shared/UniversalRender.jsx
valzav: I will enable setStore later

valzav:
@jnordberg: I was thinking about sharing routes data between routeTo and resolveRoute too, but let's do this later as refactoring. I want to have this branch out asap.

.eslintrc
sneak: what in particular requires (no pun intended) requiring a module not at the top of a file?

roadscape:
It seems like this branch re-introduces the following files, which IIRC were removed on master. Could be a sign of a bad merge at some point.

```
src/app/components/cards/CardView.js
src/app/components/elements/MarkNotificationRead.jsx
src/app/redux/PollDataSaga.js
src/server/api/notifications.js
src/server/service-worker.js
```

valzav:
@roadscape, good catch! I'll take a look what could go wrong.

.eslintrc
bnchdrff: code splitting

sneak:
@valzav please begin your refactoring as discussed above, on a branch off of this branch.

roadscape:
In terms of conflicts it doesn't look too bad but may want to document and prevent URL collisions, e.g.:

 - inaccessible permlinks: `blog|posts|comments|recommended|wallet|curation-rewards|author-rewards|permissions|created|recent-replies|feed|password|followed|followers|settings`
 - inaccessible account names: `<anything>.json`, `hot|votes|responses|trending|trending30|promoted|cashout|payout|payout_comments|created|active`

(I might have missed some)

 - Should condenser prevent saving permlinks that match other user routes?
 - Should the faucet prevent registration of any names ending in `.json`?

I don't like the idea of some posts on chain being 100% inaccessible. Nor to have invisible accounts. Is it feasible to have an alternate route or something?

sneak:
> Should condenser prevent saving permlinks that match other user routes?

Yes, or just silently append `_` or something when someone tries.

> Should the faucet prevent registration of any names ending in .json?

I think dots need to have exactly three characters after them per the blockchain.

> I don't like the idea of some posts on chain being 100% inaccessible. Nor to have invisible accounts. Is it feasible to have an alternate route or something?

Me neither.  `/:username/permlink/:permlink` would be okay for me (with the literal string `permlink` in the second part).  If `:permlink` isn't route-reserved, it should issue a permanent redirect to `/:username/:permlink`, otherwise serve it.

TimCliff:
> I think dots need to have exactly three characters after them per the blockchain.

It needs to be at least three, but more than three is accepted.

valzav:
> inaccessible permlinks: blog|posts|comments|recommended|wallet|curation-rewards|author-rewards|permissions|created|recent-replies|feed|password|followed|followers|settings

@roadscape, rather than making them inaccessible and add checks to the client and makes more sense to me to put those urls under another namespace under account, e.g.
/user/c/blog
/user/c/wallet

roadscape:
Going to re-evaluate our approach; will use this PR as reference but closing for now.

PR #1831, opened 2017-10-20
Restructure homepage

valzav:


src/app/components/cards/PostSummary.jsx
```diff
@@ -110,20 +116,53 @@ class PostSummary extends React.Component {
         const content_body = (<div className="PostSummary__body entry-content">
             <a href={title_link_url} onClick={e => navigate(e, onClick, post, title_link_url)}>{desc}</a>
         </div>);
-        const content_title = (<h3 className="entry-title">
+        const content_title = (<h2 className="articles__h2 entry-title">
```
bnchdrff: this class is unused

bnchdrff: actually many classes are unused (in css) -- this is not a big deal though!

roadscape:
Currently Userpic loading beyond the 1st page is broken because `get_state` includes relevant account objects but the pagination methods only return post (not author/profile) data. So we need to add lookups for new author references beyond the 1st page.

One solution to this is to use the new imagehoster endpoint which performs the lookups itself. @relativityboy has a relevant change to Userpic in this PR: https://github.com/steemit/condenser/pull/1790

This new endpoint has not been thoroughly tested, and the caching (which is important because it will result in dozens more calls to steemd per page load) has not yet been deployed to staging or production.

The main problem with PR https://github.com/steemit/condenser/pull/1790 is that it removes the wiring for retrieving optimal image sizes. Currently it loads either 320x320 or 120x120 images depending on the size requested. The difference between these is up to 7x pixels which can be significant in terms of bandwidth.

There is an issue logged to add size param to avatar URL in imagehoster (https://github.com/steemit/imagehoster/issues/39) but this is pointless if #1790 removes the functionality.

----

To sum up:

1. imagehoster latest needs to be deployed to dev/staging, caching/performance tested, then deployed to prod
2. #1790 needs to be merged before this one, but revert the removal of size params
3. https://github.com/steemit/imagehoster/issues/39 should be implemented sooner than later because we care about optimizing bandwidth usage AND image quality. Until then, we are degrading the UX. These were priorities for the original implementation.

PR #1838, opened 2017-10-22
Remove reputation protection from downvote post hide

TimCliff:
There is logic in condenser that prevents a users post from being hidden if they have reputation > 65. This prevents a high reputation user's posts from being hidden if their account is compromised. This PR removes the high reputation protection, so that any users posts can be hidden if they receive sufficient downvotes.


sneak:
@TimCliff has this been tested?

TimCliff:
@sneak Yes. I verified that if one of my posts is downvoted to negative it is hidden just like lower resolution users. Same thing for comments.

roadscape:
Lgtm. I implemented this change originally.

Issue #1835, opened 2017-10-22
Not allow to have URL text that is different than actual URL


Transisto:
not allow to have URL text that is different than actual URL 

Like 
`[http://steemit.com/123](http://SCAMIT.com/123)` 

A lot of fishing attempt are being made this way.

PR #1839, opened 2017-10-22
if a link smells phishy, change it to plain text

bnchdrff:
refs #1835 

i'm not sure which phishing links are out in the wild currently, but we can update this with other matchers

### DIFF
--- a/src/app/assets/stylesheets/app.scss
+++ b/src/app/assets/stylesheets/app.scss
@@ -140,10 +140,10 @@ h1, h2, h3, h4, h5, h6 {
 }
 
 .phishy {
-    color:red;
-}
-
-.phishy:hover {
-    background:red;
-    color:yellow;
+  display: inline;
+  color: red;
+  &:hover {
+    background: red;
+    color: yellow;
+  }
 }
--- a/src/app/assets/stylesheets/app.scss
+++ b/src/app/assets/stylesheets/app.scss
@@ -138,3 +138,12 @@ a.ptc {
 h1, h2, h3, h4, h5, h6 {
   line-height: 1.2 !important;
 }
+
+.phishy {
+    color:red;
+}
+
+.phishy:hover {
+    background:red;
+    color:yellow;
+}
--- a/src/app/assets/stylesheets/app.scss
+++ b/src/app/assets/stylesheets/app.scss
@@ -142,8 +142,4 @@ h1, h2, h3, h4, h5, h6 {
 .phishy {
   display: inline;
   color: red;
-  &:hover {
-    background: red;
-    color: yellow;
-  }
 }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "description": "steemit.com is the koa web server & middleware and react.js in-browser code for the world's first blockchain content + social media monetization platform!",
   "main": "index.js",
   "scripts": {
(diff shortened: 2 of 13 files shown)


bnchdrff:
another thing: do we want to provide some visual indication that the link is questionable?

valzav:
We can also paint outgoing links in red and add "outgoing" icon next to it, when user hover the icon it can say "please be careful and don't enter your password anywhere except steemit/steemconnect"

bnchdrff:
cool, i can do that. are you good w/ the feature to remove the link when the text contains a steemit url but the link target is different, in addition to the style for all outgoing links?

relativityboy:
@bnchdrff @valzav I updated styles to make text red, and yellow+red on hover.

bnchdrff:
@relativityboy `span.phishy` will not end up being links; they get the `<a>` removed

i think @valzav is recommending that we style all outgoing links (i.e. any url that doesn't match `/https?:\/\/(.*@)?(www\.)?steemit\.com/` or `/\//`) in an obvious but not super-alarm-sound-danger way. URLs that are definitely phishy, i.e. when a user submits a link which says steemit.com but goes somewhere else, should not be links and also potentialy have an additional style to discourage users from cutting and pasting them, even.

relativityboy:
@bnchdrff 
1. what @valzav said - totally. We can do that as part of this branch, or another one.
    1. additionally, it would be good to show the url in the title attribute of any links.
1. Right now the code only catches steemit.com, phishyness. It doesn't capture something like `Click [https://www.google.com/search?q=pears](https://www.google.com/search?q=my+evil+malware+site) to find only the very best information about pears!` I think that's something we should do either in this branch, or another one.
1. Url display text may mention steemit.com but not link to the actual steemit.com site, and may still be non-malicous. `[ ]` may have human readable non-link text. That's why we should append, rather than replace. the content of `child` (ex `phishySpan.textContent = child.textContent + ' / ' + url`) this gives the users of our site the ability to accurately assess.


I think it would be very good to do both the first two items in this branch if there's time. If no-time, then at a minimum do the 3rd and create issues for the items we don't do.

bnchdrff:
for point 2, are there live examples of this that are clearly malicious & something we should block? i think i'd be content with these just falling into the case of external links & styled with an icon.

re: 3: i'm catching cases when the description has a [whole `https://steemit.com` kinda thing going on](https://github.com/steemit/condenser/pull/1839/files#diff-642b3b38c9e0d9e61fabf73e7949ecd4R139) -- i'm ok with this idea but i have a feeling it's pretty uncommon for any non-malicious link to match this pattern

bnchdrff:
just addressed 3 in 87fe5bb

bnchdrff:
alright, this PR is all cleaned up and ready to go. just want to clear the design stuff w/ people who might have an opinion, though. here's a screenshot of a phishy link as well as an outbound link:

screenshot from 2017-10-23 17-01-25

@pkattera @sneak others?

src/app/assets/stylesheets/app.scss
```diff
@@ -138,3 +138,22 @@ a.ptc {
 h1, h2, h3, h4, h5, h6 {
   line-height: 1.2 !important;
 }
+
+.phishy {
+  display: inline;
+  color: red;
+}
+
+a[rel="noopener"]:after {
```
valzav: I would rather insert real image after a link - this way you can add hover message saying "External link", also I think we can make it a little bit smaller and position better if use separate img tag.

bnchdrff: @relativityboy why did these css changes disappear when you did a master merge? strange!

Issue #1771, opened 2017-10-01
Tags in non-latin languages.


zinovi:
I'd like to be able to tag my posts in the language that I wrote them in.

My use-case - Bulgarian use Cyrillic alphabet, but the functionality should be useful for posts in other non-latin languages too.

I haven't check if the first category on the steem blockchain is in ASCII or UTF, ideally it should be UTF.

The filter is done here : 
https://github.com/steemit/condenser/blob/8b38caa3fbdb6b481bf583fe567567f3e36ece85/src/app/components/cards/CategorySelector.jsx#L85

For me, the issue will be solved when I could use "България" as a first tag on my post.

sneak:
Great catch! We’ll look into this, but I think the way the tags plugin works (with categories being part of the URL) it may be limited to ASCII at the moment. 

Categories (the notion of a primary tag) are going away soon (and won’t be in the url) and the indexing of tags is getting moved out of steemd, so even if we can’t fix this bug easily with a change to condenser, it will eventually be squashed. :)

sneak:
@roadscape be advised

skenan:
I have supported the Chinese tags in my cloned condenser. The commit is here https://github.com/skenan/condenser/commit/e52cb6615d14030ef0210601a71eaa2c31a39c2d

The example: https://cnsteem.com/trending/编程
I am not sure how you guys would support other languages. If you have plans, I can help.

sneak:
@skenan could you please make a Pull Request to our condenser? :) This is great!

zinovi:
Are you open to using https://github.com/slevithan/xregexp as an regex compiler ?
Then we could use '^\\pL+$' as a character filter.

skenan:
Personally, I prefer subranges. If we choose xregexp, for the consistence, a lot of places should be changed. Also, since ^\pL+$ matches all the character which may make the tags mess up. 

There are lots of languages in the world, so could we support some majors first?

gl2748:
Blocked by TBD back-end issue / decision regarding how we will handle categories / tagging and UTF-8.

shawnpringle:
There are multiple ways of representing characters from various languages in Unicode apart from the encodings.  https://en.wikipedia.org/wiki/Unicode_equivalence

PR #1840, opened 2017-10-22
#1771 support non-latin tags

skenan:


bnchdrff:
@gl2748 this is a good one to start with -- want to add test coverage for these changes & verify?

gl2748:
Looking at this now.

gl2748:
@skenan   thanks for contributing!

bnchdrff:
@gl2748 where did this leave off?

gl2748:
@bnchdrff See:
https://github.com/steemit/condenser/compare/1771-non-latin-support?expand=1

I am going to extract testing work on `resolveRoute.js` and `TransactionSaga.js` into a separate PR as that is salvageable.
Basically until: the blockchain supports UTF-8 tags this work is blocked - furthermore there is work slated to remove categories altogether which will impact this approach.

PR #1843, opened 2017-10-23
bump steem-js version to 72a36c8

valzav:


sneak:
Read this to see why this was reverted:

https://github.com/steemit/condenser/pull/1715

PR #1845, opened 2017-10-24
bump steem-js version to 72a36c8

valzav:
https://github.com/steemit/steem-js/compare/ec4439c...72a36c8

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "sequelize-cli": "^2.3.1",
     "speakingurl": "^9.0.0",
     "sqlite3": "^3.1.8",
-    "steem": "github:steemit/steem-js#72a36c8",
+    "steem": "github:steemit/steem-js#72a36c835a74f446da47e36feea5d5f1f4dde55b",
     "store": "^1.3.20",
     "style-loader": "^0.18.2",
     "svg-inline-loader": "^0.8.0",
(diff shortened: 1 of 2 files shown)


goldibex:
My changes to `steem-js` were made to introduce appbase support on the client. Appbase itself is behind an option, but I did make changes that affect non-appbase Websocket clients, mostly rationalization of portions of the codebase badly affected by technical debt. Tests (including new ones I introduced specifically _for_ the appbase support) passed on those changes which is why I felt comfortable releasing them.

package.json
```diff
@@ -126,7 +126,7 @@
     "sequelize-cli": "^2.3.1",
     "speakingurl": "^9.0.0",
     "sqlite3": "^3.1.8",
-    "steem": "github:steemit/steem-js#ec4439c",
+    "steem": "github:steemit/steem-js#72a36c8",
```
sneak: we should be using the full sha of the commit, not the short one.

goldibex:
Can you describe in the PR issue the reason for the version bump?

roadscape:
Looking at the diff, the biggest change was a refactor with the addition of appbase switcher. 
Lgtm but good to get more eyes on it: https://github.com/steemit/steem-js/commit/1ebe788987d395af5fd8c80c1bbf06e24614a86c

Everything else is straightforward; tx expiration changes from 60 to 600s and tapos ref block is changed to use last_irreversible. I've left a note to revert both of these changes in the future (https://github.com/steemit/steem-js/issues/265). 
These changes will help in 2 conditions: (1) tx's not propagating within 60s, and (2) nodes being behind by up to 30 secs in some cases, when currently steem-js code requires they are at most 9 seconds (3 blocks) behind.

PR #1846, opened 2017-10-24
denote outbound links to ward off spam/phishing

bnchdrff:
refs #1835

and the discussion at
https://github.com/steemit/condenser/pull/1839/files/a10f879f10ffc1973090870aa18d1a1d8ac653fd#r146426090


bnchdrff:
this is the external link, and it looks like this:

imageedit_2_2956705030

bnchdrff:
@pkattera this is a moldy ol' pull request but i think it's important -- any thoughts on the outbound link icon?

bnchdrff:
feedback from @pkattera :

change colors:

use `colorTextGray` for icon color if in-post day mode & night mode

keep color as is for comments

roadscape:
Hey @bnchdrff, where are we at with this? Just styling adjustments left?

bnchdrff:
@roadscape yes -- i never did the changes; will do 'em quick right now

roadscape:
Minor change to use em units:

Before
image

After
image

Also I checked on how image links are rendered and it doesn't look bad although we should consider if there's any other edge cases:
image

roadscape:
Also @bnchdrff can we remove the 2nd `background-image` here: https://github.com/steemit/condenser/pull/1846/files#diff-8f54cac7fc81ba07061b0d97d6ad3a68R212

It looks like the only difference is the 2nd one has a fill.

bnchdrff:
the second image is to fulfil pon's request above for the in-post image to have a different fill color. em unit change is a good idea!

sneak:
@pkattera do we really believe that cluttering up ALL outbound links on the site is going to do anything meaningful against phishing?

i don’t think the cost/benefit here is worth it. we should be better at labeling suspicious and malicious links and maybe even keeping a live dangerous link registry in conveyor that is consulted in real-time
to pop up a warning interstitial, but doing this to all links regardless of suspicion seems like a net loss to me. people who are suseceptible to phishing are not going to be observant enough to notice a little icon, especially when it appears on almost every link; it becomes meaningless and invisible.

PR #1851, opened 2017-10-26
Update BadActorList.js

TimCliff:


TimCliff:
@KINGdotNET - randomwhale is already in there.

PR #1855, opened 2017-10-26
Change "People I Follow" to "My Feeds"

Jolly-Pirate:
"People I follow" is confusing, and with the sans font it looks like "People l follow". I suggest "My Feeds" instead; the users were already familiar with that more intuitive notion.


Jolly-Pirate:
closing this one and creating a proper pull.

PR #1863, opened 2017-10-26
validate account name server side before trying to create account - s…

### LINKED ISSUE #1859: enforce 16 characters account name length both client and server side

valzav:
…hould protect from hackers trying to bypass the forms (fix #1859)


src/server/api/general.js
```diff
@@ -92,6 +93,13 @@ export default function useGeneralApi(app) {
             return;
         }
 
+        const name_error = validate_account_name(account.name)
+        if (name_error) {
+            this.body = JSON.stringify({error: name_error});
+            this.status = 500;
```
jnordberg: Status code should be 400 since it is a client error

bnchdrff:
probably not worth working on this anymore, but the faucet team should ensure this is handled in that codebase! @jnordberg @bonustrack

PR #1869, opened 2017-10-27
store user preferences in cookies instead of db

valzav:


src/server/app_render.jsx
```diff
@@ -15,8 +15,15 @@ const DB_RECONNECT_TIMEOUT = process.env.NODE_ENV === 'development' ? 1000 * 60
 async function appRender(ctx) {
     const store = {};
     try {
-        let login_challenge = ctx.session.login_challenge;
         let userPreferences = {};
+        if (ctx.session.user_prefs) {
```
goldibex: `ctx.session` is signed against a secret, right?

valzav: cookie session data is encrypted via libsodium

PR #1876, opened 2017-10-28
allow user to scroll to flag dropdown popup on mobile

### LINKED ISSUE #1875: flag dropdown is partially hidden on mobile
if you try to flag a comment at the very bottom of a post on a mobile device, the dropdown dialog is partially hidden and you can't scroll to it

this is part of the suite of problems with the foundation dropdown/modal components

bnchdrff:
fixes #1875


bnchdrff:
ideally we'd use a dropdown component that would check if it's too close to the viewport edge.

until then we can apply this quick css fix.

bottom-of-post

after-dropdown-click

scroll-to-see-it

gl2748:
refs #2258 
refs #2303

PR #1883, opened 2017-10-28
Fix for "Unhandled rejection Error"

Jolly-Pirate:
steem-js fix for issue steemit#1853, as suggested by @bnchdrff.


bnchdrff:
the fix for this is to use the version of steemjs as spec'ed in master's package.json, and then use https rather than websockets

PR #1792, opened 2017-10-05
Update the signup page to include periods in the allowed account characters

Jolly-Pirate:
Update the signup page to include periods in the allowed account characters, when a user types illegal characters. For example: `user.name` is allowed.

PR #1884, opened 2017-10-28
Change "People I Follow" to "My feed"

Jolly-Pirate:
"People I follow" is confusing, and with the sans font it looks like "People l follow". I suggest "My Feeds" instead; the users were already familiar with that more intuitive notion.


imwatsi:
I don't think that this change will be necessary because from what I can tell from the new design I saw on the @steemitblog latest post, there will be a drop-down menu there where it says "People I Follow" and I presume that there will be "communities" listed there. So the phrase will be necessary to differentiate people from communities. Just my two cents. :)

Jolly-Pirate:
Still doen't make sense to me, because it's listing the feeds, not the followers. It's an awkward notion.

bnchdrff:
idk this language change doesn't make a ton of sense to me, and while #1792 was indeed approved i feel like this might've just snuck past a discussion

@sneak or others, any input?

pkattera:
@Jolly-Pirate I agree that the uppercase 'i' looks like a vertical bar divider "|". 

Let's use 'My feed' for this title (with lowercase 'f' in the jsx) to give us more future styling control in the css. Note, we'll soon be updating text across the site to be in sentence case.

@imwatsi When Communities go live, it's likely that we'll use the dropdown you mentioned as Post categories (e.g. Communities vs Blogs). 'My feed' will be a level below and will display a filtered subset of posts within those categories.

Jolly-Pirate:
@ pkattera new commit, changed to 'My feed'. I think it looks good :)

PR #1893, opened 2017-10-31
Autocomplete=On For Wallet Transfer "To:" Field

LexiconicalGap:
Toggle autocomplete on for the Steemit Wallet Transfer "To" Field. This should reduce account impersonation (Randowhale/Randomwhale/Randowale) and unnecessary traffic on the Steemit block-chain slightly from mistaken transfers/returns. On exchanges, this feature also cuts down on lost funds from transfer mistakes/typos.

For additional information:

https://steemit.com/utopian-io/@lexiconical/feature-request-reduce-transfer-fraud-by-quickly-adding-form-autocomplete-support-for-the-steemit-wallet-transfer-to-field

### DIFF
--- a/src/app/components/modules/Transfer.jsx
+++ b/src/app/components/modules/Transfer.jsx
@@ -182,7 +182,7 @@ class TransferForm extends Component {
                                 type="text"
                                 placeholder={tt('transfer_jsx.send_to_account')}
                                 onChange={this.onChangeTo}
-                                autoComplete="off"
+                                autoComplete="on"
                                 autoCorrect="off"
                                 autoCapitalize="off"
                                 spellCheck="false"


LexiconicalGap:
> I believe this was disabled on purpose to prevent certain devices from filling in the wrong value when it thinks it knows what the user is trying to type.

To be clear, some users (presumably on mobile?) have gotten incorrect auto-complete entries?

Is this still a valid reason for disabling this feature, given the existence of Esteem, for example? I have not heard of this issue.

I notice this function at all of the exchanges for cryptocurrency addresses. It certainly gives me peace of mind to see the auto-complete box confirming a successful previous transfer target. Adding this feature would make the transfer interface a little less prone to errors for new users.

sneak:
We will not be doing this, unless the autocomplete popup is provided by the server for previous transfer targets—never the browser. 

However, we will eventually be adding a confirmation screen where you can review the payee and amount before signing/broadcast.

bnchdrff:
superceded by #1964

Issue #1917, opened 2017-11-02
Open Post as main view from post list


relativityboy:
**When opening a post from a post list view, should open as new page & preserve history**

Opening a post in a modal by default is a bad user experience. Clicking on a post link should result in a 'post page' view for a better user experience.

If possible, clicking on the back button will reload the previous' page & page state (the users place in the post list would be preserved)

pkattera:
@relativityboy I agree – I'm all for this change!

bnchdrff:
an additional expectation that should be met if we use react-router to link to posts instead of using the overlays:

expected behavior:

When I go to the homepage, click on a post, and click on my browser's "back" button and return to the homepage, and then click on my browser's "forward" button, I expect to return to the post I'd just been reading.

acutal behavior: forward history is lost, and the forward button isn't clickable.

relativityboy:
@bnchdrff - 
'actual behavior' sounds like you're reporting a bug. Do you mean "current modal-type behavior doesn't put anything in history" ?

bnchdrff:
yes, it's a bug as you just described, and it would hopefully get fixed by not using a modal

plink01001:
Has this been addressed / can this be closed?

bnchdrff:
a partial fix was deployed in #1919 but clicking on a post's image still opens a modal

relativityboy:
@bnchdrff - thanks. That image thing is a bug. Thought I got all the <a> tags.

PR #1919, opened 2017-11-03
1917 - from post list articles launch in main view. 

relativityboy:
#1917

### DIFF
--- a/src/shared/UniversalRender.jsx
+++ b/src/shared/UniversalRender.jsx
@@ -82,6 +82,7 @@ async function universalRender({ location, initial_state, offchain, ErrorPage, t
 
     if (process.env.BROWSER) {
         const store = createStore(rootReducer, initial_state, middleware);
+        window.store = store; //todo: remove
         sagaMiddleware.run(PollDataSaga).done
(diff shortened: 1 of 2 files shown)


relativityboy:
@valzav would you take a close look at the updates to useScroll in `UniversalRender.jsx`?

You were the last one to touch it, but the old logic looked to be out of date.

src/shared/UniversalRender.jsx
```diff
@@ -82,15 +82,16 @@ async function universalRender({ location, initial_state, offchain, ErrorPage, t
 
     if (process.env.BROWSER) {
         const store = createStore(rootReducer, initial_state, middleware);
+        window.store = store; //todo: remove
```
goldibex: what's up with this here?

relativityboy: Justifying the review...

bnchdrff:
works for me

but... should we change all the other instances of `e => navigate`? the glaring one for me is at https://github.com/steemit/condenser/pull/1919/files#diff-8a794293a156dbd272160f7e96c0b313R224 -- clicking on the lead image for a post will still open it popup style

PR #1916, opened 2017-11-02
add Japanese language

inoue-a:
Please add Japanese to locales.

PR #1947, opened 2017-11-08
add ja help

inoue-a:
Only help contents files.
To use these files, it is necessary to develop a dynamic switching function of help locales.

Add Japanese locale request is here: PR #1916


originated:
@inoue-a Great work here.  We currently don't have robust enough switching in the help file, we will re-open this ticket when it can be addressed as a whole with a picker.

PR #1953, opened 2017-11-09
match more mentions

### LINKED ISSUE #1951: auto-linkifier for @usernames doesn't break on correct set of characters
https://steemit.com/life/@michelle.gent/you-gotta-have-friends

> `Back in the 90s Trev (@s0u1, who I met because of CB Radio) started up a` 

both `(` and `,` should be treated as whitespace for purposes of the auto-username-linking code.

bnchdrff:
fixes #1951

### DIFF
--- a/src/shared/HtmlReady.test.js
+++ b/src/shared/HtmlReady.test.js
@@ -89,4 +89,11 @@ describe('htmlready', () => {
         const res = HtmlReady(textwithmentions).html;
         expect(res).to.equal(htmlified);
     });
+
+    it('should detect only valid mentions', () => {
+        const textwithmentions = '@abc @xx (@aaa1) @_x @eee, @fff!';
+        const res = HtmlReady(textwithmentions, {mutate: false});
+        const usertags = Array.from(res.usertags).join(',');
+        expect(usertags).to.equal('abc,aaa1,eee,fff');
+    });
 });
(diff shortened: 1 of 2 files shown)


src/shared/HtmlReady.test.js
```diff
@@ -82,4 +82,11 @@ describe('htmlready', () => {
         const res = HtmlReady(somanylinks).html;
         expect(res).to.equal(htmlified);
     });
+
+    it('should link usernames', () => {
+        const textwithmentions = '<xml xmlns="http://www.w3.org/1999/xhtml">@username (@a1b2, whatever</xml>';
+        const htmlified = '<xml xmlns="http://www.w3.org/1999/xhtml"><span><a href="/@username">@username</a> (<a href="/@a1b2">@a1b2</a>, whatever</span></xml>';
+        const res = HtmlReady(textwithmentions).html;
```
roadscape: Is there a reason it tests the html output and not simply the output state of `usertags`?

bnchdrff: i wrote the tests this way because it gives a clear end-user picture of what htmlready does -- we should have tests covering `linkifyNode` and `linkify` specifically, also

roadscape:
Reverted; found a bug: https://github.com/steemit/condenser/issues/2128

Issue #1933, opened 2017-11-06
Bug: Condenser website kills my browser when loading a post with huge number of comments


ianpark:
## Problem:
Condenser website cannot open a post with huge number of comments. An obvious example is:
https://steemit.com/poctest2/@philipnbrown/genesis-poctest2

![Screen Shot 2017-11-06 at 16.46.16.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1509987134/qyixsnbqamv6lxxtjsu1.png)

On Chrome:
Chrome opens the post and hangs for long time, and then it fails to recover from the infinite loop in the script, then crash.
On Firefox:
Forefox opens the post and hangs for long time, and shows the dialog to choose to stop the script or close the window.

## How to reproduce:
1. Open the URL below using Chrome or Firefox
https://steemit.com/poctest2/@philipnbrown/genesis-poctest2
2. Your browser will hang and you will encounter the situation that were explained above.

<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@asbear/bug-condenser-website-kills-my-browser-when-loading-a-post-with-huge-number-of-comments-1932">Utopian.io</a></em><hr/>

bnchdrff:
leaving this open til it's completely solved

relativityboy:
I suggest creating new issues for improvements to initial fixes. Re-using issues is a lot like re-using facial tissues.

roadscape:
Would be solved by https://github.com/steemit/condenser/issues/3194

PR #1962, opened 2017-11-10
improve client-side performance for posts with 1000s of top level comments

bnchdrff:
only load 500 root level comments by default, and press button to load more

refs #1933 

only rendering onscreen comments is going to be difficult; i think this approach makes more sense.

ideas:
- add backend capability to serve a limited number of comments per sort
- don't render images for many coments
- consider total # of comments including children in this functionality

### DIFF
--- a/src/app/components/pages/Post.jsx
+++ b/src/app/components/pages/Post.jsx
@@ -23,10 +23,12 @@ class Post extends React.Component {
         location: React.PropTypes.object,
         signup_bonus: React.PropTypes.string,
     };
+
     constructor() {
         super();
         this.state = {
             showNegativeComments: false,
+            commentLimit: process.env.BROWSER ? 500 : 5,
         };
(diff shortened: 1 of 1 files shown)


src/app/components/pages/Post.jsx
```diff
@@ -23,11 +23,13 @@ class Post extends React.Component {
         location: React.PropTypes.object,
         signup_bonus: React.PropTypes.string,
         current_user: React.PropTypes.object,
-    };
+    }
```
relativityboy: Semicolons are part of our code-style. We should be adding them, not removing.

relativityboy:
@bnchdrff - how about adding a server-side path that reduces the number of ssr comments to 100 or eliminates rendering of comments completely?

bnchdrff:
i don't think this should be merged -- it will cause problems with deep links to comments; see #2135

PR #1963, opened 2017-11-10
Allow toggle nightmode to be used even when the user is not logged in…

netuoso:
# Issue
Toggle Nightmode button is found in the "users" profile and will not be visible unless a user is logged in.

# Solution
Put a "Toggle Nightmode" function inside the main sidebar menu that is visible to all users. Allow users to be able to toggle nightmode.

# Description
This pull request contains the code required to implement the solution to the issue detailed above. When a user is not logged in and they toggle nightmode, the nightmode theme will be active until the user hard refreshes the page. This means nightmode will persist as long as the user continues navigating thru the normal UI links.

# Screenshots
screen shot 2017-11-10 at 1 54 15 pm

### DIFF
--- a/src/app/components/App.jsx
+++ b/src/app/components/App.jsx
@@ -229,9 +229,14 @@ class App extends React.Component {
                             {tt('navigation.change_account_password')}
                         </a>
                     </li>
-                    <li className="last">
+                    <li>
                         <a href="/~witnesses" onClick={this.navigate}>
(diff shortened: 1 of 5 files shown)


sneak:
Night mode is one of many features that we want to reserve for users that sign up.  I'd be ok with the "toggle night mode" button for not-logged-in users to go to the signup page. 😉

PR #1964, opened 2017-11-11
Enable auto suggestion for transfer input based off of users own transfer history

netuoso:
# Issue
Due to the inability to turn on basic autocomplete in the transfer form, users were encountering issues and scams by sending transfers to users with the wrong name. This is largely due to the fact that users had to constantly retype the name of the users they were sending transfers to.

# Solution
As a solution to this issue, I have enabled the ability for condenser to check the users transfer history and suggest accounts based off this information.

# Additional Config Options
This implementation of the autosuggest transfer feature activates only when the first letter of the name is typed and matched with available results. On top of this, the results are limited to 10 matches at a time in an effort to reduce screen disruption.

# Screenshots
screen shot 2017-11-10 at 9 27 23 pm

screen shot 2017-11-10 at 9 33 49 pm

### DIFF
--- a/src/app/components/modules/Transfer.jsx
+++ b/src/app/components/modules/Transfer.jsx
@@ -2,6 +2,9 @@ import React, { PropTypes, Component } from 'react';
 import ReactDOM from 'react-dom';
 import reactForm from 'app/utils/ReactForm';
 import { Map } from 'immutable';
+import Autocomplete from 'react-autocomplete';
+import tt from 'counterpart';
+
 import * as transactionActions from 'app/redux/TransactionReducer';
 import * as userActions from 'app/redux/UserReducer';
 import * as globalActions from 'app/redux/GlobalReducer';
@@ -12,7 +15,6 @@ import {
     validate_memo_field,
 } from 'app/utils/ChainValidation';
 import { countDecimals } from 'app/utils/ParsersAndFormatters';
-import tt from 'counterpart';
 import { APP_NAME, LIQUID_TOKEN, VESTING_TOKEN } from 'app/client_config';
 
 /** Warning .. This is used for Power UP too. */
@@ -22,22 +24,92 @@ class TransferForm extends Component {
         currentUser: PropTypes.object.isRequired,
         toVesting: PropTypes.bool.isRequired,
         currentAccount: PropTypes.object.isRequired,
+        following: PropTypes.object.isRequired,
     };
 
     constructor(props) {
         super();
         const { transferToSelf } = props;
-        this.state = { advanced: !transferToSelf };
+        this.state = {
+            advanced: !transferToSelf,
+            transferTo: false,
+            autocompleteUsers: [],
+        };
         this.initForm(props);
     }
 
     componentDidMount() {
         setTimeout(() => {
             const { advanced } = this.state;
-            if (advanced) ReactDOM.findDOMNode(this.refs.to).focus();
+            if (advanced) this.to.focus();
             else ReactDOM.findDOMNode(this.refs.amount).focus();
         }, 300);
+
         runTests();
+
+        this.buildTransferAutocomplete();
+    }
+
+    buildTransferAutocomplete() {
+        // Get names for the recent account transfers
+        const labelPreviousTransfers = tt(
+            'transfer_jsx.autocomplete_previous_transfers'
+        );
+        const labelFollowingUser = tt(
+            'transfer_jsx.autocomplete_user_following'
+        );
+
+        const transferToLog = this.props.currentAccount
+            .get('transfer_history')
+            .reduce((acc, cur) => {
+                if (cur.getIn([1, 'op', 0]) === 'transfer') {
+                    const username = cur.getIn([1, 'op', 1, 'to']);
+                    const numTransfers = acc.get(username)
+                        ? acc.get(username).numTransfers + 1
+                        : 1;
+                    return acc.set(username, {
+                        username,
+                        label: `${numTransfers} ${labelPreviousTransfers}`,
+                        numTransfers,
+                    });
+                }
+                return acc;
+            }, Map());
+
+        // Build a combined list of users you follow & have previously transferred to,
+        // and sort it by 1. desc the number of previous transfers 2. username asc.
+        this.setState({
+            autocompleteUsers: this.props.following
+                .toOrderedMap()
+                .map(username => ({
+                    username,
+                    label: labelFollowingUser,
+                    numTransfers: 0,
+                }))
+                .merge(transferToLog)
+                .sortBy(null, (a, b) => {
+                    //prioritize sorting by number of transfers
+                    if (a.numTransfers > b.numTransfers) {
+                        return -1;
+                    }
+                    if (b.numTransfers > a.numTransfers) {
+                        return 1;
+                    }
+                    //if transfer number is the same, sort by username
+                    if (a.username > b.username) {
+                        return 1;
+                    }
+                    if (b.username > a.username) {
+                        return -1;
+                    }
+                    return 0;
+                })
+                .toArray(),
+        });
+    }
+
+    matchAutocompleteUser(item, value) {
+        return item.username.toLowerCase().indexOf(value.toLowerCase()) > -1;
     }
 
     onAdvanced = e => {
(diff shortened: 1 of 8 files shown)


netuoso:
I agree. I will make the suggested change.

netuoso:
Ok I was able to remove the need for `uniq` by using `.filter(function(e,i,arr) {return arr.lastIndexOf(e) === i});`

Wasn't able to get reduce to work the way I wanted. Maybe you can modify the filter to use reduce if you prefer. However this does result in the expected result of a sanitized array.

TimCliff:
One use case I can think of where this would cause a problem is if a user had already transferred to a bad/wrong account, it would keep showing up in their list going forward. Could it at least check the list against the blacklisted accounts? Hopefully that will at least cover the worst scenarios.

bnchdrff:
what if we weighted a user's followed accounts to the top of the autocomplete list & denoted them as such?

TimCliff:
That would be good to do too.

netuoso:
Great suggestions. Thanks for actually taking this one into consideration. I have been getting an immense benefit from the alpha version.

bnchdrff:
alright i've got this nice & cleaned up; switched to a more accessible autocomplete component, autocomplete includes followed users as well as previous transfers along with the # of times you transferred to them

todo:

- [ ] on change, query steemd & see if the user has a negative rep and if so, display a warning
- [ ] make sure it works with many many autocomplete & truncate if necessary
- [ ] get ux / style feedback from @pkattera 
- [ ] generalize the autocomplete into a wrapper component maybe

bnchdrff:
to be clear, i think the only two immediate & critical issues are truncating the list & style feedback; the other two things can happen in later PRs

src/app/components/modules/Transfer.jsx
```diff
@@ -128,11 +132,36 @@ class TransferForm extends Component {
...
         const isMemoPrivate = false;
+
+        // Get names for the recent account transfers
+        const labelPreviousTransfers = tt('transfer_jsx.autocomplete_previous_transfers');
+        const labelFollowingUser = tt('transfer_jsx.autocomplete_user_following');
+
+        const transferToLog = currentAccount.get('transfer_history').reduce((acc, cur) => {
+            if (cur.getIn([1, 'op', 0]) === 'transfer') {
+                const username = cur.getIn([1, 'op', 1, 'to']);
+                const numTransfers = acc.get(username) ? acc.get(username).numTransfers + 1 : 1
```
bnchdrff: semi pls

bnchdrff:
i've rebased this PR; i'm happy with where it's at. @relativityboy take a peek.

src/app/components/modules/Transfer.jsx
```diff
@@ -175,6 +182,42 @@ class TransferForm extends Component {
...
+                return acc;
+            }, Map());
+
+        // Build a combined list of users you follow & have previously transferred to,
+        // and sort it by the number of previous transfers.
+        const autocompleteUsers = this.props.following
+            .toMap()
+            .map(username => ({ username, label: labelFollowingUser }))
+            .merge(transferToLog)
+            .sortBy(u => u.numTransfers)
```
relativityboy: @bnchdrff @netuoso - this sort doesn't work.

src/app/components/modules/Transfer.jsx
```diff
@@ -175,6 +182,42 @@ class TransferForm extends Component {
...
+            .reverse()
+            .toArray();
```
relativityboy: We need to add a limit to 10 users (as suggested by @bnchdrff )

src/app/components/modules/Transfer.jsx
bnchdrff: are you saying we shouldn't sort by the number of previous transfers, or that it actually doesn't work? it's working for me:

screenshot from 2017-12-17 13-37-36

bnchdrff:
ok, made some changes: limited to 10 entries and refactored to a method that only runs once (no need for a wrapper quite yet, we can do that if this has to be reused somewhere else)

src/app/components/modules/Transfer.jsx
bnchdrff: ah found a condition where it doesn't work, fixing it now...

bnchdrff:
i'm not sure how to do the negative rep thing quickly/easily ... i'd rather put that part into another PR since it's another boatload of work.

the getComputedStyle error is tricky; it's in `react-autocomplete` and i'm investigating it now.

bnchdrff:
the `getComputedStyle` is actually caused by our webpack config; we are overwriting `global` in `dev.config.js` https://github.com/steemit/condenser/blob/master/webpack/dev.config.js#L29 & `prod.config.js` https://github.com/steemit/condenser/blob/master/webpack/prod.config.js#L14

these were both defined in `Initial commit` 😠 https://github.com/steemit/condenser/commit/06c90aa8260f09c4ae061e652d884f68b8a6a409#diff-bb7d85ed7641f411d08dfc8a989a1dabR40 https://github.com/steemit/condenser/commit/06c90aa8260f09c4ae061e652d884f68b8a6a409#diff-5f694022cd10cae14c85a89025b45c45R64

things seem to run fine if i remove them, and there are no refs to this variable in our sourcetree; @valzav @roadscape @relativityboy any idea what the purpose of `global.TYPED_ARRAY_SUPPORT` might be?

relativityboy:
No idea

bnchdrff:
just checked if `TYPED_ARRAY_SUPPORT` is used anywhere else in the code back at `Initial commit` -- it's not!

bnchdrff:
```
bc@zebrina:~/steemit/condenser$ git log -GTYPED_ARRAY_SUPPORT 
commit 01669549774bba961a72d4d64ae8a2332c034f6c
Author: Valentine Zavgorodnev <i@valzav.com>
Date:   Thu Oct 12 16:46:41 2017 -0400

    Upgrade to webpack3 (#1797)
    
    * only webpack related changes from webpack3 branch
    
    * update steem-js in order to use http json rpc
    
    * move imports to the top (eliminate codeclimate warning)
    
    * bump npm version to 5.4.2

commit 06c90aa8260f09c4ae061e652d884f68b8a6a409
Author: James Calfee <james@jcalfee.info>
Date:   Thu Jul 28 09:58:41 2016 -0400

    Initial commit
```

roadscape:
Grepping `node_modules`, it seems this flag is used by `buffer`. @valzav added these flags originally, let's check with him.

src/app/components/modules/Transfer.jsx
relativityboy: It didn't sort at all on live data. `.sortBy` return(s/ed) a list that has the same order as `this.props.following`. 

There were 2 transfers in the list generated by `.merge(transferToLog)` with the rest being undefined. Neither of those had their order changed. The comparator fn will prob need to move to an `a > b? 1 : (a < b? : -1 : 0)` operation.

relativityboy:
This sorting on `numTransfers` now. That's really good. But the slice is happening too soon. This could result in significantly sub-optimal results for users with lots of transfers & followings.

screen shot 2017-12-18 at 10 16 41 am

relativityboy:
I've updated the sortBy call to order by `numTransfers` desc, then `username` asc, and truncate and removed the trucate/`slice` because the single render means a user won't be prompted for their end-of-alphabet followings. @bnchdrff requesting 👍 / 👎 feedback
screen shot 2017-12-18 at 12 04 40 pm

bnchdrff:
@relativityboy looks fine to me; you aren't using that const yet though!

bnchdrff:
if you want to use a const for that one truncation, also pls use it for the other one as well

roadscape:
Implementation looks good to me, this will be a useful feature. Reliance on transfers list is fine for now, we can leverage better APIs in the future to populate the autocomplete.

Before merging we need clarification from @valzav why `TYPED_ARRAY_SUPPORT` was added; it's referenced by buffer which touches potentially critical functions such as signing.

valzav:
@roadscape, I think you are right - TYPED_ARRAY_SUPPORT was needed for the buffer lib, it must be James who added it (I most probably just helped him with webpack's config). Not sure if it's steel needed. I would suggest to try to remove and check if it breaks anything.

roadscape:
Thanks Val, James' commits around that one have to do with adding a QR reader which has since been disabled. I also found a [reference](https://github.com/feross/buffer/issues/136) to this flag being related to browser compatibility issues with typed arrays (iOS/Safari in that case).

Browser support for typed arrays looks better currently: https://caniuse.com/#feat=typedarrays

In condenser, buffer seems to be mainly involved in these operations:
 - login
 - posting
 - image upload

bnchdrff:
i'm testing those ops now, and also looking through the code for other `Buffer` usage

bnchdrff:
tested login, image upload, posting, and account creation -- no errors found!

PR #1969, opened 2017-11-11
centralize & standardize hash link scroll behavior

### LINKED ISSUE #1946: Clicking on replies does not jump to the reply
For most of the replies in my reply tab, when I click on the reply, it just takes me to the top of the post and I have to scroll down to see the reply. I have reproduced on Firefox, Chrome, and Safari. It does not seem to happen with every reply though.

This example seems to happen every time I try it:
https://steemit.com/witness-category/@timcliff/the-reports-from-the-witnesses-2017-11-07#@justncase/re-timcliff-the-reports-from-the-witnesses-2017-11-07-20171107t184836970z

relativityboy:
Closes #1946 
Page scrolling on hash navigation and on 'cold landing' has been, spread across multiple files, difficult to maintain and not meeting user expectations.

This pr moves all scrolling behavior into UniversalRender, within a class dedicated to the purpose.
Header-height is determined in one location, and scrolling is done by a scrolling middleware that's called at the correct times by Router, rather than ad-hock native access as was previously done in Comment and Post .jsx files.
   
Elimination of some ui-jumping in the post page was a part of this process.

### DIFF
--- a/src/app/utils/UserUtil.js
+++ b/src/app/utils/UserUtil.js
@@ -1,0 +1,60 @@
+/**
+ * This file replicates and extends the contents of User.js in branches new-url-scheme && notifications-ui.
+ * Additions here need to be added to ./User.js when those are merged. Imports should pull that file, and this one should be deleted.
+ */
+
+
+let store = false;
+
+export const setStore = (theStore) => {
+    if(!store) {
+        store = theStore;
+    } else {
+        throw Error("Routes.setStore - store has already been set.");
+    }
+}
+
+/**
+ * returns the current user's username *if* there is one.
+ *
+ * later can be modified to return full user object if passed *true* for 1st argument
+ * @returns {boolean}
+ */
+export const currentUser = () => {
+    const state = store.getState();
+    if(state.user) {
+        const current = state.user.getIn(['current']);
+        if(current) {
+            return current;
+        }
+    }
+    return false;
+}
+/**
+ * returns the current user's username *if* there is one.
+ *
+ * later can be modified to return full user object if passed *true* for 1st argument
+ * @returns {boolean}
+ */
+export const currentUsername = () => {
+    const state = store.getState();
+    if(state.user) {
+        const uName = state.user.getIn(['current', 'username']);
+        if(uName) {
+            return uName;
+        }
+    }
+    return false;
+}
+
+export const isOwnAccount = (username) => {
+    return (username === currentUsername());
+}
+
+/**
+ *
+ * @returns {boolean}
+ */
+export const isLoggedIn = () => {
+    return localStorage.getItem('autopost2')? true : false;
+}
--- a/src/shared/UniversalRender.jsx
+++ b/src/shared/UniversalRender.jsx
@@ -88,8 +88,15 @@ async function universalRender({ location, initial_state, offchain, ErrorPage, t
 
         const history = syncHistoryWithStore(browserHistory, store);
 
-        const scroll = useScroll();
-        
+        const scroll = useScroll( (previous, { location }) => {
+            if(location.hash) { //in situations where we have a hash url, and we are navigating forward, we want to abdicate scrolling to Post.jsx (we don't want to scroll here)
+                if((previous === null && location.action === 'POP') || (previous !== null && location.action === 'PUSH') ) {
+                    return false;
+                }
+            }
+            return true;
+        } );
+
         if (process.env.NODE_ENV === 'production') {
             console.log('%c%s', 'color: red; background: yellow; font-size: 24px;', 'WARNING!');
             console.log('%c%s', 'color: black; font-size: 16px;', 'This is a developer console, you must read and understand anything you paste or type here or you could compromise your account and your private keys.');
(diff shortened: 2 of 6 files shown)


src/app/utils/UserUtil.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * This file replicates and extends the contents of User.js in branches new-url-scheme && notifications-ui.
+ * Additions here need to be added to ./User.js when those are merged. Imports should pull that file, and this one should be deleted.
+ */
+
+
+let store = false;
+
+export const setStore = (theStore) => {
```
roadscape: It looks like this and the next 2 functions aren't used anywhere. Better to save them for a PR which introduce them?

src/shared/UniversalRender.jsx
```diff
@@ -26,10 +26,63 @@ import extractMeta from 'app/utils/ExtractMeta';
...
+    while(el) {
+        offset += el.offsetTop;
+        el = el.offsetParent;
+    }
+    return offset;
+}
+
+//BEGIN: SCROLL CODE
+const SCROLL_TOP_TRIES = 20;
+const SCROLL_TOP_DELAY_MS = 50;
```
roadscape: If I understand this right, the code will try 20 times over a period of 1 second to scroll to the desired element. I'm concerned this is not enough time.

src/shared/UniversalRender.jsx
```diff
@@ -39,6 +92,8 @@ const sagaMiddleware = createSagaMiddleware(
     ...marketWatches
 );
 
+const TOP_OFFSET = 70; //this should be declared in a sass file. The way the offset 'happens' currently is via a collection of padding and margin values on different elements; pretty bad. Once that's fixed, we can move this declaration to a central location and have it all work in lockstep.
```
roadscape: I'm not seeing where this is used

src/shared/UniversalRender.jsx
relativityboy: I'd like to err on the conservative side. If our pages take a more than a second to render we should be doing something about that.

src/shared/UniversalRender.jsx
relativityboy: That was an accidental leave in. I decided to find < header > and use it instead.

I'll remove it. < header > *should* have a fixed height declared in a scss file, exportable to js so we can skip calculations, and constants like `TOP_OFFSET`.

src/shared/UniversalRender.jsx
roadscape: Our pages take well over 1s to load, and when there are lots of images they may be loading for quite some time as well. We can leave as-is for now but let's see how it works in practice; I have a feeling 5-10s may be necessary. I doubt a user would want to scroll anywhere else that fast if they're jumping to a specific post*. If we could detect that all images are loaded we can stop trying early.

* but does this code also affect e.g. scrolling to `#comments`?

src/shared/UniversalRender.jsx
roadscape: Ok, gotcha.. but depending on viewport width there may be different header heights (at least due to mobile vs. desktop views)

src/app/utils/UserUtil.js
relativityboy: I want to get them into master so they _can be used_. I've seen 'cobbled together' url-gen going into master since this stuff was created. It's time to get ahead of the curve.

roadscape: Understood, but it makes for cleaner reviews. Otherwise, the next PR where you implement these functions it would be easy to assume the already-merged functions are tested where in fact they are untested dead code.

relativityboy: They _are_ tested.

src/shared/UniversalRender.jsx
relativityboy: <3

It affects urls with `#comments` as the hash component. That's actually how the delay requirement was discovered. I was getting partial scrolls before all the child comments had loaded (and other odd behavior)

src/shared/UniversalRender.jsx
relativityboy: > depending on viewport width

Indeed. A fixed header affects the content viewport height at all times. While I'd expect differences at breakpoints, we probably want to guard the user's content viewport height jealously. AKA, the fixed header will always only have one of a discrete set of approved by @pkattera values; all of which are specified in the `.scss` file as described above.

Hadn't figured out how we determine which one to use, but that's a problem for when we actually have them stored in the file.

Jolly-Pirate:
Commit https://github.com/steemit/condenser/pull/1969/commits/503fe20c3eee28cd5d98369aa4a458c7b5291f76 is buggy working. Now when you refresh the page containing the comment link, an internal server error 500 error occurs, before the comment appears.
error

TimCliff:
Probably related to https://github.com/steemit/condenser/issues/2015

Jolly-Pirate:
@TimCliff indeed, I thought it was only happening with hashtags in the links (because that's what I was testing) but it's happening also with regular links.

PR #1977, opened 2017-11-13
Add price per steem to user profile dropdown

netuoso:
## Issue
Currently there is nowhere on the SteemIt website to view the price per steem; a user has to navigate first to the currency market.

## Solution
Add price per steem to user profile dropdown at the top that links to the users wallet.

## Summary
Users want the ability to be able to see the price per steem at a quick glance. Without knowing this conversion, we are forcing users to navigate away from the site to 3p options, or to have to navigate to the currency market. In my opinion, there is no trade off to implementing this solution and it will greatly improve the end user experience.

## Screenshots
- desktop - screen shot 2017-11-13 at 1 23 33 am
- mobile - screen shot 2017-11-13 at 1 23 18 am

### DIFF
--- a/src/app/components/modules/TopRightMenu.jsx
+++ b/src/app/components/modules/TopRightMenu.jsx
@@ -21,7 +21,7 @@ const defaultNavigate = (e) => {
     browserHistory.push(a.pathname + a.search + a.hash);
 };
 
-function TopRightMenu({username, showLogin, logout, loggedIn, vertical, navigate, toggleOffCanvasMenu, probablyLoggedIn, nightmodeEnabled, toggleNightmode}) {
+function TopRightMenu({username, showLogin, logout, loggedIn, vertical, navigate, toggleOffCanvasMenu, probablyLoggedIn, nightmodeEnabled, toggleNightmode, pricePerSteem}) {
     const mcn = 'menu' + (vertical ? ' vertical show-for-small-only' : '');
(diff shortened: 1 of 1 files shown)


pkattera:
For discussion: I think it's better to keep this dropdown exclusively for items related to the person's profile and activity.

To solve @netuoso issue – we could make the price of STEEM more visible by placing it on the homepage with a link to the currency market page. Thoughts?

screen shot 2017-11-13 at 11 11 03 am

netuoso:
@TimCliff, in response to your concerns, the Witness price feed price is used every where throughout the app to apply a conversion to dollars. If we don't care about it anywhere else, why would it be a problem to show what is being used to convert STEEM to $$ on the wallet page?

Maybe a tooltip or something to clarify the difference?

TimCliff:
AFAIK, all of the places it is currently used are estimated values such as potential payouts, and those estimates are affected by the same factors. 

If users see it being shown as the “STEEM Price” the default user behavior is going to be to assume it is the current price. 

A tooltip seems like a good solution to me.

Jolly-Pirate:
A small price ticker for SBD and STEEM prices in the header perhaps? Not too intrusive.

relativityboy:
This is an interesting idea. I think that if we do it, we'll want to find a different location for it. Please make an issue for it. 

Closing.

PR #1979, opened 2017-11-13
Theme fixes

pkattera:
This contains a series of styling improvements and visual bug fixes. Includes:
- Fix for tags in primary navigation
- Applying the layout-list & layout-block views to personal blog pages
- Post Summary author header
- A new hidden compressed view


src/app/components/modules/Header.jsx
```diff
@@ -210,9 +210,7 @@ class Header extends React.Component {
                                     </Link>
                                 </li>
                                 <li className="Header__top-steemit show-for-medium noPrint"><Link to={logo_link}><span className="beta fade-in--10">beta</span></Link></li>
-                                
-                    
-                                {(topic_link || user_name || page_name) && sort_order }
+                                {(topic_link || user_name || page_name) && sort_order && <li className="delim show-for-small-only"></li>}
```
bnchdrff: what exactly is this fixing?

src/app/components/pages/UserProfile.jsx
```diff
@@ -312,10 +314,60 @@ export default class UserProfile extends React.Component {
             tab_content = <div>Unavailable For Legal Reasons.</div>;
         }
 
+        var page_title = "";
+        // Page title
+
+        if (isMyAccount) {
+            if (section === 'blog') {
+                page_title = "My blog";
```
bnchdrff: we should use translated strings here

src/app/components/pages/UserProfile.jsx
```diff
@@ -312,10 +314,60 @@ export default class UserProfile extends React.Component {
...
+                    </div>         
+                </div> 
+                <hr className="articles__hr" /> 
+            </div>
+        );
+
         if (!(section === 'transfers' || section === 'permissions' || section === 'password')) {
             tab_content = <div className="row">
-                <div className="UserProfile__tab_content layout-list column">
+                <div className={'UserProfile__tab_content column' + layoutClass + " " + section}>
```
bnchdrff: this would be a good place to use `classnames()`

src/app/components/modules/Header.jsx
pkattera: Right now, there is a new bug introduced where if you click on a tag on desktop, this appears. Reverting this back to the original fixes it.

screen shot 2017-11-13 at 1 56 54 pm

bnchdrff: looks like it used to have content though; a pipe `|` https://github.com/steemit/condenser/commit/a528595192e76c19fc670735f7c783fbabca03b3

PR #1981, opened 2017-11-13
Use a frozen lockfile when building in docker

jnordberg:


relativityboy:
#2065 took this over.

Issue #1930, opened 2017-11-06
Bug at promotion text box. using Mobile phone.


ruahtheking:
# Bug at promotion text box. using Mobile phone.
While promoting one of my posts. in steemit using my mobile phone I encounter a small but issuable bug.
I think it's need to be fixed as soon as possible for mobile users that always using this features.
![Screenshot_2017-11-05-14-50-40.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1509950504/imizesxon0a0hdffcwot.png)
the text box width is too narrow. I guess the text box needed a specific width and not sizeable. so when using mobile its width is fixed. 

To change the width of the textbox you need to enable your screen rotation and rotate it to make it wider. 
![Screenshot_2017-11-05-14-51-24.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1509950646/zyzgciwca6mjxozklvr1.png)

While confirming the promotion and rotate it to its original. heres the screenshot.
![Screenshot_2017-11-05-14-52-00.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1509950706/z7t4uhpgowqfrcoisyru.png)

To fix this box, you can change its width into fixed state and not auto.
Device : Samsung Note 3
Browser : Chrome for Android. Ver. 56.0.2924.87
<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@ruah/bug-at-promotion-text-box-using-mobile-phone">Utopian.io</a></em><hr/>

TimCliff:
Duplicate issue: https://github.com/steemit/condenser/issues/1520

plink01001:
Closing duplicate

Issue #1520, opened 2017-06-23
Mobile - Promote Page Amount entry is too small


TimCliff:
When viewing the promote page, the text box to enter the amount is too small. Any numbers that the user types are hidden.
https://steemit.com/devs/@justinashby/promote-post-bug-help-needed

TimCliff:
The user who reported this is saying that it will not let them promote for any amount, and it is giving them a validation error that they do not have enough funds, even if they have more than enough SBD in their account. User is on chrome, zmax pro.

originated:
Closing ticket, the SBD input was updated. https://github.com/steemit/condenser/issues/1520  If there is issues is still actual operation issues/with promotions please open another ticket

PR #1982, opened 2017-11-13
Fixes tiny text field bug on small screens

sirrius-a-b:
Fix for issue: #1520 ( and it's duplicate #1930 )

Changes I made:
- modified the grid CSS classes for promotion popup from `small-4` to `small-6 large-4`

So that the amount section has it's width set to 50% on small screens, but on large screens the width will remain as it is now.

Here is how it looks with this change on different screen sizes:

### 360px
promote-360px

### 480px
promote-480px

### 800px
promote-800px

### 1024px
promote-1024px

### DIFF
--- a/src/app/components/modules/PromotePost.jsx
+++ b/src/app/components/modules/PromotePost.jsx
@@ -75,7 +75,7 @@ class PromotePost extends Component {
                        <p>{tt('promote_post_jsx.spend_your_DEBT_TOKEN_to_advertise_this_post', {DEBT_TOKEN})}.</p>
                        <hr />
                        <div className="row">
-                           <div className="column small-6 large-4">
+                           <div className="column small-7 medium-5 large-4">
                                <label>{tt('g.amount')}</label>
                                <div className="input-group">
                                    <input className="input-group-field" type="text" placeholder={tt('g.amount')} value={amount} ref="amount" autoComplete="off" disabled={loading} onChange={this.amountChange} />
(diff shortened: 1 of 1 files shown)


pkattera:
Thanks @sirrius-a-b – this is certainly an improvement. How about also setting a 'min-width' as well to cover it for all the smaller screens? 

.PromotePost .input-group-field { 
  min-width: 120px;
}

sirrius-a-b:
Hey @pkattera thanks for looking over my PR.

I agree that on screens smaller than 360px, namely 320px, the input field gets smaller. For this I modified the classes to be `small-7 medium-5 large-4`. This should cover all screens larger than 320px :)

I do not think that we should force a `min-width` there because on small screens this will overflow out of the responsive CSS grid classes. Should we not try to use the CSS grid classes everywhere we can and only force widths in the most extreme cases?

Here is how this looks with the new commit:

### 320px
promote2-320px

### 360px
promote2-360px

### 480px
promote2-480px

### 800px
promote2-800px

### 1024px
promote2-1024px

pkattera:
Thanks @sirrius-a-b – this looks fine to me. 

ps. In terms of 'min-width', you're right about it potentially causing overflows and scrollbars etc. However, in general, if the min-width of the whole item is less than '200px', I'm not too concerned.

gl2748:
Merged:
https://github.com/steemit/condenser/pull/2130/commits/eca8a9340dd1fcd997c2e4ae735aeeea654c75ad

PR #1984, opened 2017-11-14
Allow users the ability to upload a secondary mobile cover image

netuoso:
## Issue
Due to the fact that cover images are recommended to be sized and set for 2048px, they end up looking horrible on mobile (if even visible).

## Solution
Instead of some magic with trying to resize the image and position it depending on screen size, allow users the ability to set a mobile cover image that will be displayed automatically when the screen size is mobile size.

## Summary
Users get a lot of pride about differentiating their profiles from others. Unfortunately, they are running into issues trying to size an image that has a happy medium between to drastically different screen resolutions. This pull request is very unobtrusive to the UI and allows users to have more freedom over their own profile page customization. This is an easy win for an greatly improved end user experience on condenser.

## Screenshots
screen shot 2017-11-14 at 1 12 42 am
screen shot 2017-11-14 at 1 12 34 am
screen shot 2017-11-14 at 1 12 20 am

### DIFF
--- a/src/app/components/pages/UserProfile.jsx
+++ b/src/app/components/pages/UserProfile.jsx
@@ -42,6 +42,16 @@ export default class UserProfile extends React.Component {
         this.loadMore = this.loadMore.bind(this);
     }
 
+    componentDidMount() {
+        window.addEventListener("resize", this.resize.bind(this));
+        this.resize();
+    }
+
+    resize() {
+        this.setState({showMobileCover: window.innerWidth <= 639});
+        this.forceUpdate();
+    }
+
     shouldComponentUpdate(np) {
         const {follow} = this.props;
         const {follow_count} = this.props;
@@ -434,7 +444,7 @@ export default class UserProfile extends React.Component {
             </div>
          </div>;
 
-        const {name, location, about, website, cover_image} = normalizeProfile(account);
+        const {name, location, about, website, cover_image, cover_image_mobile} = normalizeProfile(account);
         const website_label = website ? website.replace(/^https?:\/\/(www\.)?/, '').replace(/\/$/, '') : null
 
         let cover_image_style = {}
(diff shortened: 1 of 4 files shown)


gl2748:
@sneak @pkattera thoughts?

pkattera:
I'm okay with this change.

Note: In the next couple of months, I'd like to improve this further by:
 – Setting new hero dimensions for desktop (it feels too short right now) 
 – Giving explicit guidelines to users on the site about our recommended ideal dimensions 
 – Moving profile information into a card like structure, possibly not covering the hero image on desktop
 – Introducing an image cropper that helps people crop images to our recommended dimensions.

src/app/components/pages/UserProfile.jsx
```diff
@@ -434,20 +444,27 @@ export default class UserProfile extends React.Component {
...
+
+        const showMobileCover = this.state.showMobileCover;
+
         return (
             <div className="UserProfile">
 
                 <div className="UserProfile__banner row expanded">
 
-                    <div className="column" style={cover_image_style}>
+                    <div className="column" style={showMobileCover ? cover_image_mobile_style : cover_image_style}>
```
gl2748: What styles are applied here if `showMobileCover` is true but there is no `cover_image_mobile` ... won't `cover_image_mobile_style` be an empty object?

src/app/components/pages/UserProfile.jsx
```diff
@@ -42,6 +42,16 @@ export default class UserProfile extends React.Component {
...
+    componentDidMount() {
+        window.addEventListener("resize", this.resize.bind(this));
+        this.resize();
+    }
+
+    resize() {
+        this.setState({showMobileCover: window.innerWidth <= 639});
+        this.forceUpdate();
+    }
+
```
gl2748: @netuoso thanks for this. I would prefer to not have to manage the responsive behavior in state like this. We have foundation styles available and this behavior could be  implemented by applying the classname 'show-for-small' to the new mobile image and 'hide-for-small' to the existing one.
https://foundation.zurb.com/sites/docs/v/3.2.5/media-queries.php#screenCode

Example:
https://github.com/steemit/condenser/blob/master/src/app/components/modules/TopRightMenu.jsx#L53

bnchdrff:
Hi @netuoso, we're cleaning up our pull request queue and need to close this PR for now; this isn't to say that we don't want to merge it, though -- rather, feel free to re-open it when you have a chance to make the requested changes. Thanks for your contributions!

Issue #1985, opened 2017-11-14
condenser sets HSTS, while it should be a TLS termination job 


Gandalf-the-Grey:
condenser is setting STS header like this:
`strict-transport-security:max-age=15552000; includeSubDomains`
it uses `helmet` which uses `hsts` by default

Setting STS header at app level might be unexpected and undesired behavior and it should be set consciously by admin at TLS termination.

If TLS termination is configured properly, then it already has such header which makes one added by condenser redundant.
If TLS termination is not configured then such header doesn't make any sense because [the header is ignored in insecure HTTP](https://tools.ietf.org/html/rfc6797#section-8.1)

PR #1986, opened 2017-11-14
Remove redundant Strict-Transport-Security response header

Gandalf-the-Grey:
Fixes: #1985

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -161,7 +161,9 @@ if (env === 'production') {
     app.use(koa_logger());
 }
 
-app.use(helmet());
+app.use(helmet({
+    hsts: false
+}));
 
 app.use(
     mount(


sneak:
or rather the frontend proxy should only add it if it does not exist. this way someone running condenser even without a reverse proxy is still protected.

Gandalf-the-Grey:
unfortunately wouldn't be protected because condenser by default listen on 8080 without TLS and [in such case this header is ignored anyway](https://tools.ietf.org/html/rfc6797#section-8.1)
but actually because of that we can leave it as it is on condenser if we assume that frontend proxy would override it instead of duplicating

sneak:
i'm sold.  do it on the TLS terminator.

bnchdrff:
@Gandalf-the-Grey to confirm, this branch does not require any changes to `sdc-frontend` & can be deployed now?

Gandalf-the-Grey:
yes, changes to `sdc-frontend` [was already made](https://github.com/steemit/sdc-frontend/pull/21)

PR #1988, opened 2017-11-14
Depend on steem-js via the @steemit/steem-js pkg

jnordberg:
Published steem-js version is same as previously used (except some doc changes, diff here https://github.com/steemit/steem-js/compare/72a36c835a74f446da47e36feea5d5f1f4dde55b...02d396915d187fdf01ca65fdce8934359d69936d)


sneak:
do not approve until we get appropriate security around that npm namespace in place today.

goldibex:
@jnordberg @sneak status here?

package.json
```diff
@@ -128,7 +128,7 @@
     "sequelize-cli": "^2.3.1",
     "speakingurl": "^9.0.0",
     "sqlite3": "^3.1.8",
-    "steem": "github:steemit/steem-js#72a36c835a74f446da47e36feea5d5f1f4dde55b",
+    "@steemit/steem-js": "^0.6.5",
```
goldibex: peg the exact version number. don't caret this

jnordberg: My intention was to have the lockfile be the source of truth for exact versions but we could do both

goldibex: I want it to be a conscious decision for us to change the version number. could be excessively paranoid however. @sneak your thoughts?

sneak: exact version AND lockfile checksum.

jnordberg:
Changes in steem-js: https://github.com/steemit/steem-js/compare/72a36c835a74f446da47e36feea5d5f1f4dde55b...aeeca39e49bb6fff2cf680b70ebe7aba25897177

sneak:
@jredbeard is the package namespace and permissions sufficiently locked down on npm?

PR #1993, opened 2017-11-14
Add more comment payout options so users have more control over payout

netuoso:
## Issue
Currently only 3 payout options display on posts limiting users ability to control their reinvestment

## Solution
Add two more payout options that should help address 99% of use cases. These payout options are 75/25 and 25/75.

## Summary
Allowing the users finer control over their payout options might promote them to choose 75% instead of 50% promoting the stability of the entire Steem economy. It would possibly be a good idea in the future to add a slider component so users could more effectively fine tune the payout options on posts.

## Screenshots
screen shot 2017-11-14 at 4 02 01 pm


Jolly-Pirate:
The new percentages may be confusing. Perhaps indicate what each % corresponds to.

pkattera:
@Jolly-Pirate @netuoso How about this:

screen shot 2017-11-14 at 5 38 33 pm

I don't think we need to use the word 'Default' since it is already the default choice and will have a selection mark beside it in most browsers.

netuoso:
I think that looks clean. I try my best to be the least obtrusive as possible on the UI.

👍

TimCliff:
I thought the 50/50 option was 50% SP, 50% SBD, and that the blockchain would not allow more than 50% of the payout to be liquid.

mvandeberg:
The most SBD a post can be rewarded with is 50%. The blockchain supports 0-50% SBD payouts in 0.01% increments.

netuoso:
screen shot 2017-11-14 at 9 53 02 pm

Took the suggestions into account and made the changes. Removed the 25% SBD option.

mvandeberg:
The tooltip should read STEEM Power and not STEEM. It is misleading as the user will not receive STEEM directly.

netuoso:
@mvandeberg https://github.com/steemit/condenser/pull/1993/commits/c7cd63fe1dcdecfb068d3424bcb0dd7394bed16c

It was in a commit before your comment. Though it is SP and not STEEM Power. Is that suitable?

gl2748:
Currently [this branch](https://github.com/steemit/condenser/tree/more-options-for-payout-2), has this looking like:
### Before: 
current

### After.
new_dropdown

pkattera:
@TimCliff @gl2748 Should we say 'SBD' instead of 'STEEM Dollars'?

TimCliff:
@pkattera - I'm neutral on that. I think either are OK.

gl2748:
According to @sneak we want 'fewer payout options, not more' so marking this as WIP and on hold for the time being.

PR #2002, opened 2017-11-16
Add SteemPays and variations to BadActorList.

netuoso:
## Issue
User `steem-pays` is a service that users can use to exchange STEEM. There are users `steempay,steempays,steemitpay` that are typo squatting.

## Solution
Add the accounts typo squatting to the BadActorList for warning.

## Summary
Steem-Pays is a legit service that users are encouraged to send money to. It seems some users are making typo mistakes and sending money to a scam or inactive account. This change adds the violating users to the BadActorList so the sending user is warned before burning their SBD.


relativityboy:
Conflicts. Re-addressing in #2149 (quicker)

PR #2005, opened 2017-11-16
Add target blank to User Profile website link to prevent the window location from changing

netuoso:
## Issue
Currently, there is no external target added to the link on the User Profile website. This causes the window.location to change, taking users away from SteemIt.

## Solution
Simply add the external target option so the website will open in a new tab and will not take the user away from the SteemIt website.

## Summary
This ridiculously easy, unobtrusive change will improve the end user experience when they end up clicking on a link the user sets in the User Profile (Settings). Due to the fact the external target is not being set, following the user's website link will change the window.location forcing the website to have to be completely reloaded to continue browsing.

## Screenshots
screen shot 2017-11-15 at 11 09 04 pm


Jolly-Pirate:
I suggest you add `rel="noopener"` while you're at it. Check this reference for its benefits: https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/

PR #2009, opened 2017-11-16
1946 handle even more cases of delayed document loading (scrollHeight…

### LINKED ISSUE #1946: Clicking on replies does not jump to the reply

relativityboy:
Closes #1946 ...even more
… change) in a generic way

This update handles instances where the browser incorrectly sticks the viewport to the bottom of the page, or does not hold the scroll position at the correct element when images take longer to load.

### DIFF
--- a/src/app/utils/UserUtil.js
+++ b/src/app/utils/UserUtil.js
@@ -21,11 +21,13 @@ export const setStore = (theStore) => {
  * @returns {boolean}
  */
 export const currentUser = () => {
-    const state = store.getState();
-    if(state.user) {
-        const current = state.user.getIn(['current']);
-        if(current) {
-            return current;
+    if (store) {
+        const state = store.getState();
+        if (state.user) {
+            const current = state.user.getIn(['current']);
+            if (current) {
+                return current;
+            }
         }
     }
     return false;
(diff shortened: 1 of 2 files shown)


src/app/utils/UserUtil.js
```diff
@@ -21,11 +21,13 @@ export const setStore = (theStore) => {
  * @returns {boolean}
  */
 export const currentUser = () => {
-    const state = store.getState();
-    if(state.user) {
-        const current = state.user.getIn(['current']);
-        if(current) {
-            return current;
+    if (store) {
```
roadscape: Shouldn't it be a hard requirement to call setStore before these methods? ie should it not be strict and throw an exception?

relativityboy: That was my initial thought. 

@bnchdrff changed my mind about that. Maybe a console statement would be appropriate.

bnchdrff: did you look into delaying scroll until content & images are done loading?

bnchdrff: did you look into delaying scroll until content & images are done loading?

PR #2020, opened 2017-11-17
remove redux-modules

### LINKED ISSUE #1148: Consider to get rid of redux-modules in order to decrease the size of the bundle file
redux-modules has ramda dependency and it adds up 273 Kb alone to the output bundle

bnchdrff:
fixes #1148

also, switch to fully immutable typed redux store

this single commit squashes all the work from @valzav and me in the remove-redux-modules and webpack3 branches


goldibex:
I approve of immutability but why is it necessary to pull in a library with massive throw weight to accomplish it?

src/app/components/cards/PostSummary.jsx
```diff
@@ -281,13 +280,13 @@ export default connect(
...
-            blogmode: state.app.getIn(['user_preferences', 'blogmode']),
+            username: state.getIn(['user', 'current', 'username']) || state.getIn(['offchain', 'account']),
+            blogmode: state.getIn(['app', 'user_preferences', 'blogmode']),
         };
     },
 
     (dispatch) => ({
-        dispatchSubmit: data => { dispatch(user.actions.usernamePasswordLogin({...data})) },
-        clearError: () => { dispatch(user.actions.loginError({error: null})) }
+        dispatchSubmit: data => { dispatch({type: 'user/USERNAME_PASSWORD_LOGIN', payload: {...data}}) },
```
goldibex: is this really the right way to namespace events? It means you're going to have to parse them at the reducer, which costs time.

Two alternative proposals:
1. `type` should be an array of namespaces, e.g.:
```
// option 1
dispatch({ type: ['user', 'USERNAME_PASSWORD_LOGIN'] });
// option 2
dispatch({ type: 'user', name: 'USERNAME_PASSWORD_LOGIN' });
```

Thoughts?

src/app/components/cards/PostsList.jsx
```diff
@@ -211,12 +211,12 @@ class PostsList extends React.Component {
 
 export default connect(
     (state, props) => {
-        const pathname = state.app.get('location').pathname;
-        const current = state.user.get('current')
-        const username = current ? current.get('username') : state.offchain.get('account')
-        const content = state.global.get('content');
-        const ignore_result = state.global.getIn(['follow', 'getFollowingAsync', username, 'ignore_result']);
-        const userPreferences = state.app.get('user_preferences').toJS();
+        const pathname = state.getIn(['app', 'location']).pathname;
```
goldibex: this is an argument in favor of option 1 for event `type`s as described above

src/app/components/elements/SaveLogin.jsx
```diff
@@ -1,42 +0,0 @@
-import React, {Component, PropTypes} from 'react'
```
goldibex: where'd this file go?

src/app/components/pages/About.jsx
```diff
@@ -1,17 +1,20 @@
 import React from 'react';
-import { APP_NAME, APP_URL } from 'app/client_config';
+import { APP_NAME, LANDING_PAGE_URL } from 'app/client_config';
 import tt from 'counterpart';
 
+console.log('-- APP_NAME -->', APP_NAME);
```
goldibex: remove this

src/app/redux/AppReducer.js
```diff
@@ -24,7 +24,7 @@ const defaultState = Map({
     user_preferences: Map({
         locale: null,
         nsfwPref: 'warn',
-        theme: 'light',
+        nightmode: false,
```
goldibex: why change this?

src/app/redux/MarketReducer.js
```diff
@@ -1,40 +1,29 @@
...
-        }
-    ]
-});
+
+const defaultState = Map({status: {}});
+
+export default function reducer(state = defaultState, action) {
+    const payload = action.payload;
+
+    if (action.type === 'market/RECEIVE_ORDERBOOK') {
```
goldibex: `switch`

bnchdrff:
to avoid react-redux warnings 😭 https://github.com/gajus/redux-immutable#problem
> because Redux combineReducers treats state object as a plain JavaScript object.

luckily it's just about as lightweight of a dependency as we can ever add: https://github.com/gajus/redux-immutable/blob/master/package.json#L20 (no sub-deps)

src/app/redux/TransactionReducer.js
```diff
@@ -0,0 +1,123 @@
...
+const defaultState = fromJS({
+    operations: [],
+    status: { key: '', error: false, busy: false, },
+    errors: null
+});
+
+export default function reducer(state = defaultState, action) {
+    const payload = action.payload;
+
+    if (action.type === 'transaction/CONFIRM_OPERATION') {
```
goldibex: `switch`

src/app/redux/TransactionReducer.js
```diff
@@ -0,0 +1,123 @@
...
+                state = state.update('errors', errors => {
+                    return errors ? errors.set(errorKey, errorStr) : Map({[errorKey]: errorStr});
+                });
+            }
+        }
+        if (errorCallback) try { errorCallback(errorKey) } catch (error2) { console.error(error2) }
+        return state
+    }
+
+    if (action.type === 'transaction/DELETE_ERROR') {
```
goldibex: `switch`

src/app/redux/TransactionReducer.js
```diff
@@ -0,0 +1,123 @@
...
+                if (/Missing Active Authority /.test(errorKey)) {
+                    errorKey = "Not your valid active key.";
+                    errorStr = "Transaction failed: Not your valid active key.";
+                }
+                state = state.update('errors', errors => {
+                    return errors ? errors.set(errorKey, errorStr) : Map({[errorKey]: errorStr});
+                });
+            }
+        }
+        if (errorCallback) try { errorCallback(errorKey) } catch (error2) { console.error(error2) }
```
goldibex: 1. always brace conditional statements.
2. Do not hide such a massive, consequential application behavior by removing newlines.
3. do not `On Error Resume Next`. If there is an error it should blow up the application in a suitably panicky way.

src/app/redux/UserSaga.js
```diff
@@ -26,7 +25,7 @@ export const userWatches = [
     uploadImageWatch,
 ]
 
-const highSecurityPages = Array(/\/market/, /\/@.+\/(transfers|permissions|password)/, /\/~witnesses/)
+const highSecurityPages = [/\/market/, /\/@.+\/(transfers|permissions|password)/, /\/~witnesses/]
```
goldibex: do not use regular expressions for testing security conditions.

src/app/redux/tests/global.test.js
```diff
@@ -19,7 +17,7 @@ describe('global reducer', () => {
         const state = Immutable.fromJS(require('./global.json'));
         //const action = {type: 'global/RECEIVE_STATE', payload: state};
         expect(
-            reducer(undefined, actions.receiveState(state))
+            reducer(undefined, {type: 'global/RECEIVE_STATE', payload: state})
```
goldibex: why is `undefined` being passed as an argument

src/app/components/elements/SaveLogin.jsx
bnchdrff: good q... removed by @valzav i think in the first batch of changes; it's not referenced anywhere in the app:

```
bc@zebrina:~/steemit/condenser$ grep -r SaveLogin src
bc@zebrina:~/steemit/condenser$ 
```

goldibex:
`Immutable.js` is a 143kb lead weight that can add hundreds of milliseconds to TTFI even on year-old devices. If we were already using it before I won't hold the release up over it; but we will be having a discussion about these gigantic dependencies and how much effort is involved in ridding ourselves of them.

State management can be done effectively for applications much, much larger than Condenser using plain old React and Redux without any bells or whistles.

src/app/redux/MarketReducer.js
bnchdrff: good idea; this, as well as many other changes, are just following the style previously set up in the reducers before. this is a good time to refactor them, though

src/app/redux/UserSaga.js
bnchdrff: this can be handled in another PR unless you want me to do it here

src/app/redux/tests/global.test.js
bnchdrff: i didn't write this test & that's a good question :)

bnchdrff:
the thing is, we're already using Immutable all over the place. this PR is not about that. it's about removing redux-modules. i didn't know we wanted to remove Immutable. we can do that i guess?

i find it pretty useful though.

src/app/redux/TransactionReducer.js
bnchdrff: again this is a port of the previously defined reducer. i can work on refactoring it now if you think it's a good idea.

src/app/components/elements/SaveLogin.jsx
goldibex: @valzav what did this file do and why is it gone?

src/app/components/cards/PostSummary.jsx
bnchdrff: i'm into option 1... thoughts, others?

src/app/redux/UserSaga.js
goldibex: yeah, you didn't add that. i'll file an issue. disregard for now.

src/app/redux/tests/global.test.js
goldibex: @valzav ?

goldibex:
@bnchdrff I've created an issue for reviewing TTFI and Javascript bundle size and assigned it to you. Disregard my comments on Immutable.js on this PR, they don't block merge.

src/app/redux/TransactionReducer.js
goldibex:
```js
if (errorCallback) {
  errorCallback(errorKey);
} else {
  throw new Error('PANIC: no callback registered to handle error ' + errorKey);
}
```

that was fast

src/app/redux/AppReducer.js
bnchdrff: it was setting an incorrect default state previously; `nightmode` is the correct prop name

bnchdrff: (my bug)

src/app/components/cards/PostsList.jsx
bnchdrff: not sure what this line has to do with actions -- selecting from the store here

src/app/redux/TransactionReducer.js
bnchdrff: looks like the original behavior is old as dirt https://github.com/steemit/condenser/blame/master/src/app/redux/Transaction.js#L113

i'm not super familiar with how this part of the app works, and would like others' feedback first

src/app/components/cards/PostSummary.jsx
bnchdrff: i don't think there are any huge disadvantages to using an array for action.type but on the flipside, i think it might be nice to follow the ducks pattern -- it's something that many developers are familiar with

bnchdrff: we can discuss in https://github.com/steemit/condenser/issues/2036

src/app/redux/UserSaga.js
bnchdrff: discuss in https://github.com/steemit/condenser/issues/2037

src/app/redux/tests/global.test.js
bnchdrff: discuss in https://github.com/steemit/condenser/issues/2038

bnchdrff:
see https://github.com/steemit/condenser/issues/2021 for lib size discussion

src/app/components/elements/SaveLogin.jsx
valzav: SaveLogin was not in use

src/app/redux/tests/global.test.js
valzav: because when reducer is called with undefined state it will use the default one

valzav:
@goldibex: I would recommend to keep Immutable - it plays very nice with redux and makes comparing parts of the state supplied into components much easier.

src/app/redux/AppReducer.js
goldibex: got it

goldibex:
lgtm

bnchdrff:
closing for now; redux-immutable is incompatible with redux-form 5.x and upgrading to 7.x is a very big project. will reopen a new PR without redux-immutable changes.

Issue #1666, opened 2017-08-16
Add full "og" tags to blog/profile pages


kellyjanderson:
Right now if you share the link to your blog/profile on Facebook the preview is only text and titled Steemit.

This should have the title "User Name - Steemit" at least and should pick a reasonable image for Facebook.

![screen shot 2017-08-16 at 11 09 31 am](https://user-images.githubusercontent.com/643583/29378483-1e3e47a4-8274-11e7-9d71-494ca1280ef9.png)

plink01001:
Are there plans to make any changes for this issue, or should it be moved to the `enhancement-review` category?

originated:
I just address this issue, the naming declaration on the facebook OG tags was not correct. Closing. PR https://github.com/steemit/condenser/pull/2024

PR #2024, opened 2017-11-18
Correct issue with naming on meta tags not rendering properly

originated:


originated:
Fix for https://github.com/steemit/condenser/issues/1666

PR #2028, opened 2017-11-19
Enable a check on the transfer form to ensure users are not sending transfers to themselves

netuoso:
## Issue
Users are able to send a transfer to their own account through the wallet

## Solution
When a user attempts to send a transfer to their own account, display a warning and disable the send button.

## Summary
Right now, STEEM can be a very confusing places for users. Due to this, I believe it is a good idea to let the frontend provide as much assistance as possible while educating the user. If a user has multiple accounts, they will often need to send transfers between them. On condenser, a user has the ability to send a transfer to the account they are currently logged in as. While not very damaging, it is still confusing and might result in some users not getting their funds to where they were hoping.

## Screenshots
screen shot 2017-11-18 at 11 28 23 pm
screen shot 2017-11-18 at 11 28 28 pm

### DIFF
--- a/src/app/components/modules/Transfer.jsx
+++ b/src/app/components/modules/Transfer.jsx
@@ -72,7 +72,9 @@ class TransferForm extends Component {
             initialValues: props.initialValues,
             validation: values => ({
                 to:
-                    ! values.to ? tt('g.required') : validate_account_name(values.to, values.memo),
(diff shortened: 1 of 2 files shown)


bnchdrff:
let's piggyback this one onto #1964

netuoso:
@bnchdrff Done. Merged into #1964.

Closing this.

PR #2029, opened 2017-11-19
Fix error message when a user attempts to power down with less than the minimum required amount

netuoso:
## Issue
When a user with less than the minimum required amount (~2SP) tries to power down it tells them it requires ~300SP to powerdown.

## Solution
Change the error message to represent the current account power down minimum.

## Summary
This is definitely an edge case, but there will be users with less than 2SP that attempt the power down from their account. These users will run into an error telling them it will take about 300SP to be able to begin powering down. Updating the error message is a good solution in removing confusion around the frontend error messages. It would be good to internationalize these error messages.

## Screenshots
image


gl2748:
See: https://github.com/steemit/condenser/pull/2121

PR #2030, opened 2017-11-19
Detect when a user presses escape key on settings update login screen

netuoso:
## Issue
When a user hits ESCAPE to cancel settings update, the "update" button never returns and is constantly a loading indicator.

## Solution
Detect when a user presses ESCAPE to cancel updating user settings and reset the loading state to false to unhide the "update" button

## Summary
On the user settings page, things get a bit unfriendly for the user when they dismiss the login modal after pressing "update" using ESCAPE. Dismissing the login modal via ESCAPE is not firing off an event to change the application state for the loading value. This pull request detects when a user presses escape only on the User Settings page due to the loading indicator hiding the submit button.

## Screenshots
image

### DIFF
--- a/src/app/components/modules/Settings.jsx
+++ b/src/app/components/modules/Settings.jsx
@@ -19,6 +19,7 @@ class Settings extends React.Component {
         }
         this.initForm(props);
         this.onNsfwPrefChange = this.onNsfwPrefChange.bind(this);
+        this.escFunction = this.escFunction.bind(this);
     }
 
     initForm(props) {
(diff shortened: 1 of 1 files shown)


bnchdrff:
This is an intersting bug and a good solution for this specific instance of the bug, but I'd like to pause & consider how to construct a more general fix for the underlying issue.

Ideally an escape would cancel whatever request saga is currently happening, and would pop up a toaster message saying "<whatever action> cancelled" -- that'd be pretty cool & nice ux. @pkattera opinions on this?

relativityboy:
This  is nice a non-interfering patch for the `esc` action.

I think the bug is that a spinner is created before the update interaction with the server is started.

I'm going to close this issue and file a bug recommending we change the loading state to true only when we've actually sent a request to update user settings.

PR #2031, opened 2017-11-19
Fix body length check on post and comment

netuoso:
## Issue
`Uncaught Error: [sprintf] property "maxKb" does not exist` error encountered when body length exceeds maximum

## Solution
Fix the syntax of the problematic code to send the correct prop to the internationalizer.

## Summary
Currently, there is an error that is encountered when a user inputs too many characters in the post or comment reply editor. When this error is encountered, it prevents the reply editor from properly displaying further errors until the user completely solves the body length error (without being told it is too long). This pull request fixes the syntax of the translator that is being used when the error message is triggered. The max body length was also changed from 100 to 65 in this pull request.

## Screenshots
- error - 
screen shot 2017-11-19 at 1 28 31 pm
- error fixed - screen shot 2017-11-19 at 1 27 46 pm


src/app/components/elements/ReplyEditor.jsx
```diff
@@ -144,7 +144,7 @@ class ReplyEditor extends React.Component {
     initForm(props) {
         const {isStory, type, fields} = props
         const isEdit = type === 'edit'
-        const maxKb = isStory ? 100 : 16
+        const maxKb = isStory ? 65 : 16
```
relativityboy: you've changed the max size for a story from 100 -> 65 here... that's a significant change. Can you point to an issue or bug? If not please revert.

roadscape: Need to check if the backend has some firm limit, but currently the max block size (as voted on by witnesses) is 64kb so it would not be possible to publish a comment bigger than that.

netuoso: Exactly this. The limit should probably actually be 64, but I figured if it was a pasting issue that users would at least realize there is this limit. It would be misleading to currently allow a total TX size larger than 65,536 bytes.

PR #2039, opened 2017-11-21
Pass flag status for voting dispatch, remove curation rewards message…

originated:
… for flag removal


bnchdrff:
@originated i added a space before the "will reset your curation rewards for this post" fragment

before:

screenshot from 2017-12-18 13-14-59

after:

screenshot from 2017-12-18 13-17-21

bnchdrff:
on second thought i think we should not compound sentence fragments like this because they will be hard to translate

bnchdrff:
created https://github.com/steemit/condenser/issues/2176 for the sentence fragement l10n issue

PR #2044, opened 2017-11-22
Whistle signup flow updates

### LINKED ISSUE #1989: Condenser signup process changes for Whistle
We need to make the `Got an account? Login` to trigger `postMessage` if clicked
We need to make the last page of signup process trigger `postMessage` when the last page is activated

Each trigger must give a separate status message/json so that we can differentiate which  `postMessage` is been triggered

relativityboy:
Closes #1989 
Remove navigation that takes users out of signup flow. Remove possibility of 'beta' appearing anywhere.

Of necessity, the UI is very sparse. @pkattera approved.

### DIFF
--- a/src/app/Main.js
+++ b/src/app/Main.js
@@ -9,6 +9,7 @@ import universalRender from 'shared/UniversalRender';
 import ConsoleExports from './utils/ConsoleExports';
 import {serverApiRecordEvent} from 'app/utils/ServerApiClient';
 import * as steem from 'steem';
+import {determineViewMode} from "app/utils/Links";
 
 
 window.onerror = error => {
@@ -89,7 +90,8 @@ function runApp(initial_state) {
         window.$STM_csrf = initial_state.offchain.csrf;
         delete initial_state.offchain.csrf;
     }
-    initial_state.app.viewMode = (window.location.search.indexOf(PARAM_VIEW_MODE + '=' + VIEW_MODE_WHISTLE) > -1)? VIEW_MODE_WHISTLE : '';
+
+    initial_state.app.viewMode = determineViewMode(window.location.search);
 
     const location = `${window.location.pathname}${window.location.search}${window.location.hash}`;
     universalRender({history, location, initial_state})
--- a/src/app/Main.js
+++ b/src/app/Main.js
@@ -89,7 +89,7 @@ function runApp(initial_state) {
         window.$STM_csrf = initial_state.offchain.csrf;
         delete initial_state.offchain.csrf;
     }
-    initial_state.app.viewMode = (window.location.search.indexOf(PARAM_VIEW_MODE + '=' + VIEW_MODE_WHISTLE))? VIEW_MODE_WHISTLE : '';
+    initial_state.app.viewMode = (window.location.search.indexOf(PARAM_VIEW_MODE + '=' + VIEW_MODE_WHISTLE) > -1)? VIEW_MODE_WHISTLE : '';
 
     const location = `${window.location.pathname}${window.location.search}${window.location.hash}`;
     universalRender({history, location, initial_state})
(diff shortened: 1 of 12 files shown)


src/app/Main.js
```diff
@@ -87,6 +89,7 @@ function runApp(initial_state) {
         window.$STM_csrf = initial_state.offchain.csrf;
         delete initial_state.offchain.csrf;
     }
+    initial_state.app.viewMode = (window.location.search.indexOf(PARAM_VIEW_MODE + '=' + VIEW_MODE_WHISTLE) > -1)? VIEW_MODE_WHISTLE : '';
```
goldibex: this ternary expression is really confusing, please rewrite it

src/app/components/pages/Approval.jsx
```diff
@@ -16,6 +17,14 @@ class Approval extends React.Component {
     }
 
     render() {
+        if(process.env.BROWSER && this.props.viewMode === VIEW_MODE_WHISTLE) {
+            try {
+                window.postMessage(WHISTLE_SIGNUP_COMPLETE);
+            } catch(e) {
+                console.error('Cannot window.postMessage when not in whistle app');
```
goldibex: why does this try/catch block exist?

src/app/utils/Links.js
```diff
@@ -35,8 +35,42 @@ export default {
     ipfsPrefix: /(https?:\/\/.*)?\/ipfs/i,
 }
 
+//todo: possible this should go somewhere else.
```
goldibex: 1. please write TODO, not todo
2. please resolve these TODOs. where else would this code go?

src/app/utils/Links.test.js
```diff
@@ -48,6 +48,27 @@ describe('Links', () => {
     })
 })
 
+describe('makeParams', () => {
```
goldibex: thank you

src/server/app_render.jsx
```diff
@@ -111,8 +111,13 @@ async function appRender(ctx) {
                 offchain.recover_account = account_recovery_record.account_name;
             }
         }
+        const initial_state = {
+            app: {
+                viewMode: (ctx.request.search.indexOf(PARAM_VIEW_MODE + '=' + VIEW_MODE_WHISTLE) > -1)? VIEW_MODE_WHISTLE : ''
```
goldibex: refactor ternary expression as requested above

src/shared/constants.js
```diff
@@ -0,0 +1,9 @@
+export const PARAM_VIEW_MODE = 'view_mode';
```
goldibex: i heartily endorse this methodology for constants.

do not export default an object. this pointlessly removes es6 module magic.

you can really just write
```export const FOO = 'bar';```
it's ok, i promise

src/app/components/pages/Approval.jsx
relativityboy: Because the implementation of postMessage in react-native is different from the spec for web browsers. Calling window.postMessage in a browser with only one argument results in an Error. Because we cannot guarantee a user will never land on this page from a browser, we handle the exception.

src/app/utils/Links.js
relativityboy: Most likely in a file called utils/Url.js that's currently in the `notifications-ui` branch. Some recent conversations with @bnchdrff suggest that file may be changing or going away.

The TODO is not currently resolvable, but I will capitalize the word for you.

src/app/components/pages/Approval.jsx
goldibex: you're not handling the error, you're ignoring it. why can't you guarantee a user will never land on the page from an error?

src/app/utils/Links.js
goldibex: if we haven't formally decided to deprecate the file, that's where the code should go.

src/app/components/pages/Approval.jsx
relativityboy: Can you rephrase
> why can't you guarantee a user will never land on the page from an error?

PR #2056, opened 2017-11-26
Some names are not visible in the comments user tooltip - Nightmode

netuoso:
## Issue
Some names are not visible in the comments user tooltip due to NightMode colors

## Solution
Change the colors displayed for visited links in the comment tooltip in NightMode.

## Summary
When a user clicks on another user's profile image in a comment reply while in NightMode, links will not be visible because they are the same color as the background of the tooltip box. In an effort to improve end user experience, this PR contains the code necessary to ensure these links are not washed out behind a while background.

## Screenshots
image
screen shot 2017-11-26 at 4 32 55 pm


relativityboy:
@netuoso - master is a bit of a moving target right now, and this branch has gotten out of date. Would you update your fork so I can merge cleanly?

relativityboy:
@netuoso - thanks for this PR. It looks like master has already made the text visible with different commits so we don't need to merge it, but you've inspired a further improvement, #2082 that will be worked in short order.

Thanks for this, and keep up the good work!

netuoso:
@relativityboy no problem i might even take a look at that other issue tomorrow. didnt see the comment from a couple days ago.

Issue #2059, opened 2017-11-27
Purge unused components


gl2748:
## Audit of extant components:
```
Icon: 68
LoadingIndicator: 20
Tooltip: 9
NotifiCounter: 9
Voting: 8
Userpic: 8
FoundationDropdown: 8
Author: 7
SvgImage: 6
FoundationDropdownMenu: 6
TagList: 5
MarkNotificationRead: 5
DropdownMenu: 5
Callout: 5
VerticalMenu: 4
UserList: 4
Orderbook: 4
Keys: 4
Follow: 4
WalletSubMenu: 3
TransactionError: 3
Reputation: 3
Reblog: 3
ChangePassword: 3
VotesAndComments: 2
QrReader: 2
PageViewsCounter: 2
OrderbookRow: 2
HelpContent: 2
GeneratedPasswordInput: 2
YoutubePreview: 1
UserNames: 1
UserKeys: 1
SuggestPassword: 1
SidebarStats: 1
SidebarNewUsers: 1
SidebarLinks: 1
ShareMenu: 1
SavingsWithdrawHistory: 1
QrKeyView: 1
PasswordReset: 1
PasswordInput: 1
OrderHistory: 1
HorizontalMenu: 1
FormattedAsset: 1
DepthChart: 1
DateJoinWrapper: 1
CountryCode: 1
ConvertToSteem: 1
CheckLoginOwner: 1
AuthorDropdown: 1
Template: 0
SlateEditor: 0
SignupProgressBar: 0
SaveLogin: 0
ReplyEditor: 0
OrderhistoryRow: 0
LocalizedCurrency: 0
ExperimentWrapper: 0
```

To reproduce:
#### warning, this executes a bash script.

From root dir:
`wget https://gist.githubusercontent.com/gl2748/3b466ed0bc903ed9b6eb3ff7af25cb93/raw/fa1901fef592bd3f88a455c4ef7f8542266aafb1/yezhovshchina.sh && chmod +x yezhovshchina.sh && ./yezhovshchina.sh src/app/components/elements`

## Suspects
```
Template: 0
SlateEditor: 0
SignupProgressBar: 0
SaveLogin: 0
ReplyEditor: 0
OrderhistoryRow: 0
LocalizedCurrency: 0
ExperimentWrapper: 0
```

I therefore propose investigating those components that are not explicitly invoked elsewhere in the code base, and removing if deemed superfluous.

gl2748:
[...]

PR #2060, opened 2017-11-27
purge vestigal components, refs #2059

gl2748:
see #2059


src/app/components/elements/OrderHistory.jsx
```diff
@@ -1,5 +1,4 @@
 import React from "react";
-import HistoryRow from "./OrderhistoryRow";
```
bnchdrff: this is being used https://github.com/steemit/condenser/pull/2060/files#diff-153a69ec7ac931570754a21a62d2b9bdL35

PR #2089, opened 2017-12-03
Strings spanish translate

oPensyLar:
Completed some missing strings to spanish


relativityboy:
@oPensyLar - thank you!

Issue #2081, opened 2017-12-01
use "prettier" to automatically format codebase


goldibex:
Code formatting in condenser consumes an unhealthy amount of developer cognitive burden. Therefore we should do the following work:

1: ADD PRETTIER
- [x] add prettier to devDependencies
- [x] set up prettier with some reasonable conventions
- [x] create a new npm script called "fmt" that runs prettier

2: RUN PRETTIER
- [x] run Prettier over the codebase
- [x] Make sure Prettier doesn't break anything
- [x] make Prettier a precommit hook

After 2, all existing pull requests will be conflicted out.

BUT to resolve the merge conflicts we only need to fast-forward each branch to 1 and run prettier. We could even do this automatically for all outstanding PRs.

bnchdrff:
additional step for `2: RUN PRETTIER` can be
- [ ] sync up eslint rules with what prettier does, and add it back to pre-commit hook

PR #2092, opened 2017-12-03
add prettier & yarn command `yarn run fmt`

### LINKED ISSUE #2090: add prettier
the first step for #2081 

- [ ] add prettier to devDependencies
- [ ] set up prettier with some reasonable conventions
- [ ] create a new npm script called "fmt" that runs prettier

bnchdrff:
closes #2090 

see #2081 for the backstory

also, remove the existing precommit hooks until we fix them -- are ppl ok with this?


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    tabWidth: 4,
+    singleQuote: true,
+    trailingComma: "es5"
```
goldibex: Let's use JSON for this file, as that's already the format for .babelrc and .eslintrc

package.json
```diff
@@ -23,6 +24,7 @@
   "license": "MIT",
   "dependencies": {
     "@steem/crypto-session": "git+https://github.com/steemit/crypto-session.git#83a90b319ce5bc6a70362d52a15a815de7e729bb",
+    "@steemit/steem-js": "0.6.7",
```
goldibex: this shouldn't be here?

goldibex:
and yes, removing the existing precommit hooks is fine for the moment

package.json
bnchdrff: ah it got re-ordered by yarn

PR #2111, opened 2017-12-07
Replace iso.bootstrap escaped html char parsing

relativityboy:


bnchdrff:
this PR was a fix for another PR #1810 which was reverted

PR #1992, opened 2017-11-14
Add a small search input field next to the search button on the header.

netuoso:
## Issue
Currently users have to click the small search icon to be redirected to the search page. They then have to click the input on the search page, begin typing and press enter to see results.

## Solution
Add a search input field next to the search icon if users prefer to be more direct.

## Summary
At the moment, it is kind of a pain to search the website. We have the solution in place of the google site search link, but it isnt very user friendly. In an effort to approve the end user experience, this pull request adds a small input field next to the search icon on the header. It respects the rules of staying hidden on small screens.

## Screenshots
screen shot 2017-11-14 at 2 46 33 pm
screen shot 2017-11-14 at 2 46 13 pm

PR #2122, opened 2017-12-08
Add search input to the header

gl2748:
Supersedes:  https://github.com/steemit/condenser/pull/1992
I have reviewed and tested locally - all is good.

### DIFF
--- a/src/app/components/modules/TopRightMenu.jsx
+++ b/src/app/components/modules/TopRightMenu.jsx
@@ -113,7 +113,7 @@ function TopRightMenu({
         ];
         return (
             <ul className={mcn + mcl}>
-                <li className={lcn + ' Header__search'}>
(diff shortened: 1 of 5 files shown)


bnchdrff:
the search box has a border when focused in firefox:

screenshot from 2017-12-09 12-15-47

src/app/components/modules/TopRightMenu.jsx
```diff
@@ -122,7 +122,7 @@ function TopRightMenu({
                         )}
                     </a>
                 </li>
-                {!pathCheck ? submit_story : null}
```
bnchdrff: why is this changed?

gl2748: good spot, not so sure, undoing...

gl2748:
@bnchdrff addressed your comment.

PR #2130, opened 2017-12-08
Fix for promotion text field 1930 2

gl2748:
Supercedes: PR: https://github.com/steemit/condenser/pull/1982


gl2748:
# before
promote_before

# after
promote_after

gl2748:
# as deployed to staging
deployed

Issue #2128, opened 2017-12-08
mentions can break URLs


roadscape:
![image](https://user-images.githubusercontent.com/5168676/33787039-1ea1f012-dc31-11e7-95f0-eab8d8ef3a14.png)

PR #2131, opened 2017-12-08
match more mentions

roadscape:
Reverts "Revert "match more mentions"" steemit/condenser#2129

Original PR: https://github.com/steemit/condenser/pull/1953
Issue: https://github.com/steemit/condenser/issues/2128
Revert: https://github.com/steemit/condenser/pull/2129


bnchdrff:
alright, my fix is in -- this change excludes words where a forward slash precedes the potential username.

src/shared/HtmlReady.js
```diff
@@ -262,18 +262,24 @@ function linkify(content, mutate, hashtags, usertags, images, links) {
...
-        if (valid && usertags) usertags.add(userLower);
-        if (!mutate) return user;
-        return (
-            space +
-            (valid ? `<a href="/@${userLower}">@${user2}</a>` : '@' + user2)
-        );
-    });
+    // Cribbed from https://github.com/twitter/twitter-text/blob/master/js/twitter-text.js#L90
+    content = content.replace(
+        /(^|[^a-zA-Z0-9_!#$%&*@＠\/]|(^|[^a-zA-Z0-9_+~.-\/]))[@＠]([a-z][-\.a-z\d]+[a-z\d])/gi,
```
relativityboy: This is getting pretty hairy. Why is the middle block `(^|[^a-zA-Z0-9_+~.-\/])` needed? Can it be merged with the first?

Also, it looks like that expression will match `[@s0u1` and `@@s0u1` is that what we want? What if the user wrote` [@lbah has a bad website](https://badwebsite.com)`
@roadscape @bnchdrff

bnchdrff: looks like twitter just rewrote their library, which i was referencing. i fixed the link to point at the specific revision: https://github.com/twitter/twitter-text/blob/v1.14.7/js/twitter-text.js#L90

yes it's complicated, because what we're doing is complicated ;) yes we want to match both of those things. markdown links will have already been converted to HTML by the time it gets here. i added a test to cover links with mentions in them.

PR #2132, opened 2017-12-09
Allowing steem URI scheme in markdown content

aaroncox:
These two small changes allow posts to render links that use the `steem://` URI scheme. I have successfully tested these changes running a dev version of condenser and viewing the following post:

[https://steemit.com/spam/@webdev/test-6#@jesta/re-webdev-test-6-2017128t175053871z](https://steemit.com/spam/@webdev/test-6#@jesta/re-webdev-test-6-2017128t175053871z)

This URI scheme will allow launching of external apps (e.g. Vessel) to further interact with readers.

Currently you can workaround this issue and embed `steem://` links into posts by using a URL shortener (e.g. tinyurl.com). This takes the links and converts them shorter `https://` links, which render as valid links. This isn't preferable though - since the user would have to generate the `steem://` link and then run each one through a URL shortener.


gl2748:
@aaroncox Vessel looks really cool - a lot like metamask? Great stuff, I have opened a PR that includes this work along with tests to cover the changes:
https://github.com/steemit/condenser/pull/2134

aaroncox:
@gl2748 Awesome, thank you very much! 

Yeah - I think it's evolving in a similar direction as metamask, except as a standalone app rather than a browser extension. Excited to see what people end up building with it!

gl2748:
@sneak how do we want to handle this work? 'steem' is obviously not a standard URI scheme per:
https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

bnchdrff:
i think we should close this PR and discuss in another forum, probably on a steemit.com post

aaroncox:
Does this need to be a standard in order to allow through the sanitization filters? 

This is rather discouraging.

sneak:
@aaroncox We need to do a security analysis of the potential impacts and attack vectors such a change could potentially enable. For obvious reasons, we take a conservative approach. I would love to support this feature once we’ve had a chance to do that.

aaroncox:
It's a reasonable concern and should be the utmost priority, but please realize this change doesn't allow for anything you can't already do today. If you discover attack vectors that are capable of being performed through a custom protocol - those attacks are possible today (using a URL shortener and swapping the link to HTTPS) and wouldn't actually be "enabled" by allowing `steem:` links through the sanitizer. 

If a security analysis is what's needed to proceed, I'll respect the process and await the results. I'm happy to help if and where I can.

gl2748:
In the event this is all o-k, the PR here covers this work and has tests for the changes. https://github.com/steemit/condenser/pull/2134

roadscape:
@gl2748 - PR #2134 was closed yet it contains tests for this feature. It's also on a local branch which is required for testing. So I'm reopening it and moving the feature/discussion there.

gl2748:
@roadscape sounds good. I would say that this has been languishing for quite some time and perhaps the consensus (albeit by implication) is that this work will never be merged and that all associated PRs can be closed?

roadscape:
The change is simple; I'll dust it off, and check still relevant/safe.

PR #2137, opened 2017-12-10
Closes #2136 - Add warning to SBD conversion page

### LINKED ISSUE #2136: Add warning to SBD->STEEM conversion page
The SBD->STEEM conversion page is an advanced user feature. Many users `convert` SBD to STEEM, thinking this is the best way to get STEEM for their SBD, when in a lot of cases they would be better off using the internal market or exchange to trade SBD for STEEM. Update the conversion page to warn users that it is an advanced feature, and recommend other options for them to get STEEM for their SBD.
(I will be working on this.)

TimCliff:


TimCliff:
Before:
image

After:
image

relativityboy:
We just removed access to this UI. It may not longer be needed. 

On hold pending further investigation.

relativityboy:
Closing. Looks like we don't need it, but thanks for the work!

PR #2140, opened 2017-12-11
add a test for the transaction saga

gl2748:
This is fallout from the work on PR #1771.
A test for the transaction saga.

PR #2145, opened 2017-12-13
update redux saga to v0.16.0

gl2748:
There is some debate over the best way to test Redux-Saga.

I have been frustrated with [writing unit tests for them](https://github.com/steemit/condenser/pull/2140/files#diff-5842cb9a8b07e8c503a252d37680ec6b). Updating Redux-saga will let me try a more general integrative approach -
https://github.com/jfairbank/redux-saga-test-plan

As well as allowing for a concise TDD for future Sagas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "react-timeago": "^3.1.2",
     "redux": "^3.3.1",
     "redux-form": "5.3.4",
-    "redux-saga": "^0.9.5",
+    "redux-saga": "^0.16.0",
     "remarkable": "^1.7.1",
     "sanitize-html": "^1.11.4",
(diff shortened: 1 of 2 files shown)


bnchdrff:
i'm going through the redux-saga release logs to make sure we aren't missing anything big

gl2748:
Yes there is 1.0.0 in beta. That might be the bigger issue but getting to 0.16.0 will help...

bnchdrff:
our sagas aren't compatible with this update; i am unable to log in, for example. i think we should postpone this work until there's a very clear reason to update.

bnchdrff:
closing this for now, see https://github.com/steemit/condenser/issues/2175

PR #2156, opened 2017-12-14
add test for the AppReducer

### LINKED ISSUE #2160: Create Tests for the Reducers.
Why test?

- Because the business relies on the site being operational and bug free.

- So that we can add new features to the App (e.g. communities) and have confidence we have not broken something when we do.

- So that we can understand how the app works and think about improving it / simplifying it.

- So that when we are working we only have to worry about one thing at a time - the task at hand and not the wider context.

# Reducers.

- [x] App Reducer. -  #2156
- [x] Global Reducer. - #2232
- [ ] Market Reducer. - #2230 
- [x] Offchain Reducer. - #2147
- [ ] Root Reducer.
- [ ] Transaction Reducer.
- [ ] User Reducer.

gl2748:
A test for the App reducer. fixes #2160


src/app/redux/AppReducer.js
```diff
@@ -2,18 +2,19 @@ import { Map, OrderedMap } from 'immutable';
 import tt from 'counterpart';
 
 // Action constants
+export const appActions = {
```
bnchdrff: why this change? this breaks API: https://github.com/steemit/condenser/blob/master/src/app/components/elements/ConvertToSteem.jsx#L6

bnchdrff: i'd rather have it named something else, like actionConsts or something

src/app/redux/AppReducer.test.js
```diff
@@ -0,0 +1,173 @@
...
+});
+
+describe('App reducer', () => {
+    it('should provide a nice initial state', () => {
+        const initial = reducer();
+        expect(initial).to.equal(defaultState);
+    });
+    it('should return correct state for a LOCATION_CHANGE action', () => {
+        const initial = reducer();
+        const actual = reducer(initial, mockActions['LOCATION_CHANGE']);
```
bnchdrff: why not use the action creator as the reducer input here?

bnchdrff: if we use the action creator function as input for the reducer test, we get coverage for both functions

PR #2184, opened 2017-12-19
remove unneeded HtmlReady test scenario

### LINKED ISSUE #2178: htmlready test error needs explanation

bnchdrff:
closes #2178 

wheni wrote this test suite i thought this was a needed consideration but it turns out it is not, and the test runner gets a (not-test-fail-causing) error when we run it. because we don't ever pass in a non-html string this test is safe to remove.


bnchdrff:
actually i withdraw this PR this is dumb

Issue #2187, opened 2017-12-19
Update contribution guidelines.


relativityboy:
Condenser's enhancement contribution guidelines are changing from a PULL REQUEST first to ISSUE first policy. The .md files need to reflect that.

PR #2192, opened 2017-12-20
remove htmlready console.error call

### LINKED ISSUE #2178: htmlready test error needs explanation

bnchdrff:
closes #2178

this call is not needed

### DIFF
--- a/src/shared/HtmlReady.js
+++ b/src/shared/HtmlReady.js
@@ -109,8 +109,6 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
             ...state,
         };
     } catch (error) {
-        // Not Used, parseFromString might throw an error in the future
-        console.error(error.toString());
         return { html };
     }
 }


relativityboy:
Just saw @bnchdrff's comment on #2187 
> HtmlReady never receives non-html content;

bnchdrff:
i understand the original author's instinct to use a try..catch block; the xmldom parser could concievably have weird errors and it's a reasonable stopgap to return the original html if a problem comes up. on the flipside it could be more secure to display nothing in that case.

changing that behavior is outside the scope of this PR.

roadscape:
Please tag me on any HtmlReady changes!

roadscape:
If this is the case that was causing the error:

```
    it('should return plain text without html unmolested', () => {
        const teststring = 'teststring lol';
        expect(HtmlReady(teststring).html).to.equal(teststring);
    });
```

Why is this one fine:
```
     it('should detect only valid mentions', () => {
         const textwithmentions =
             '@abc @xx (@aaa1) @_x @eee, @fff! https://x.com/@zzz/test';
         const res = HtmlReady(textwithmentions, { mutate: false });
         const usertags = Array.from(res.usertags).join(',');
         expect(usertags).to.equal('abc,aaa1,eee,fff');
     });
```

If I'm looking at it right, the comment above the log line is incorrect and it's probably referring to `XMLSerializer.serializeToString`.

roadscape:
Confirmed; full trace:
```
     TypeError: Cannot read property 'prefix' of null
      at Document.nodeSerializeToString [as toString] (node_modules/xmldom/dom.js:920:22)
      at XMLSerializer.serializeToString (node_modules/xmldom/dom.js:914:31)
      at exports.default (src/shared/HtmlReady.js:109:39)
      at Context.<anonymous> (src/shared/HtmlReady.test.js:97:16)
```

Why would `serializeToString` fail on a DOM Document?
```
Document {
  implementation: { _features: {} },
  childNodes:
   { '0':
      Text {
        ownerDocument: [Circular],
        data: 'teststring lol',
        nodeValue: 'teststring lol',
        length: 14,
        previousSibling: null,
        nextSibling: null,
        parentNode: [Circular] },
     length: 1 },
  doctype: null,
  firstChild:
   Text {
     ownerDocument: [Circular],
     data: 'teststring lol',
     nodeValue: 'teststring lol',
     length: 14,
     previousSibling: null,
     nextSibling: null,
     parentNode: [Circular] },
  lastChild:
   Text {
     ownerDocument: [Circular],
     data: 'teststring lol',
     nodeValue: 'teststring lol',
     length: 14,
     previousSibling: null,
     nextSibling: null,
     parentNode: [Circular] },
  _inc: 2 }
```

bnchdrff:
@roadscape i think the issue is that it fails when the input is not valid xml, i.e. a text string

bnchdrff:
the established behavior of HtmlReady is to output the original string if the xml parser throws; when i wrote the test suite i added in a bit to cover this behavior but not the `console.warn` output

roadscape:
It doesn't seem like a case we should silence; if anything we should reraise the exception. Especially if all code paths are "supposed" to provide proper HTML input. I overlooked this pattern when writing the 'should detect only valid mentions' test, so really it seems that test should be failing.

bnchdrff:
it def. seems like we should throw & not output the un-readied html if xmldom has an error. not sure why it was written this way.

roadscape:
> This suggests we shouldn't have the try-catch block at all.

^ I'm thinking this is the best solution for now. If the usages are audited and this should never happen then we need to have it fail loudly if it does.

PR #2193, opened 2017-12-20
fixes #2191, revert changes to UserList, fixes #2191

### LINKED ISSUE #2191: Followers and followed lists are empty
followed

followers

gl2748:
Fixes #2191 
# Before
screenshot from 2017-12-20 13-41-48

# After
screenshot from 2017-12-20 13-38-19

### DIFF
--- a/src/app/components/elements/UserList.jsx
+++ b/src/app/components/elements/UserList.jsx
@@ -30,11 +30,11 @@ class UserList extends React.Component {
         let user_list = users.map(user => (
             <UserListRow user={user} key={idx++} />
         ));
+        user_list = user_list.toArray();
 
         let currentIndex = -1;
-        const usersLength = users.length;
+        const usersLength = users.size;
         const limitedIndex = Math.min(historyIndex, usersLength - PER_PAGE);
-
         user_list = user_list.reverse().filter(() => {
             currentIndex++;
             return (


src/app/components/elements/UserList.jsx
```diff
@@ -23,18 +23,19 @@ class UserList extends React.Component {
 
     render() {
         const { state: { historyIndex } } = this;
+        const account = this.props.account;
```
bnchdrff: why not destructure?

src/app/components/elements/UserList.jsx
```diff
@@ -23,18 +23,19 @@ class UserList extends React.Component {
...
         const users = this.props.users;
         const title = this.props.title;
 
         let idx = 0;
         let user_list = users.map(user => (
-            <UserListRow user={user} key={idx++} />
+            <UserListRow account={account} user={user} key={idx++} />
```
bnchdrff: this is a bit out of scope for the bugfix going on here but can't we use user ids as a key?

src/app/components/elements/UserList.jsx
```diff
@@ -23,18 +23,19 @@ class UserList extends React.Component {
...
         ));
+        user_list = user_list.toArray();
```
bnchdrff: rather than `let`ing `user_list` we should make it a `const` and chain the `toArray` call after the `map`:

```
const user_list = users.map(blahblah).toArray()
```

src/app/components/elements/UserList.jsx
```diff
@@ -23,18 +23,19 @@ class UserList extends React.Component {
...
 
         let currentIndex = -1;
-        const usersLength = users.length;
+        const usersLength = users.size;
         const limitedIndex = Math.min(historyIndex, usersLength - PER_PAGE);
-
         user_list = user_list.reverse().filter(() => {
```
bnchdrff: ahhh it's being redefined again down here...ok maybe we should split off this refactor stuff into another issue 🙃

goldibex:
@gl2748 did you see my last comment and decide to annoy me? this is your second notice. write better PR titles.

src/app/components/elements/UserList.jsx
gl2748: Yeah so this bug was introduced by  me not understanding that users was an immutable type.

PR #1975, opened 2017-11-13
Show the image upload helper on all replies, even comments

netuoso:
## Issue
Users want to be able to use the image upload helper that they see on posts when replying to comments. Many users are browsing SteemIt on mobile and trying to upload images any other way is a PAIN!

## Solution
Show the image upload helper at the bottom of the reply box for comments as well as posts.

## Summary
Due to the fact that a large number of users are browsing the SteemIt website on mobile devices, it would be wise to enable the image upload helper on all reply boxes. Often times users wish to be able to upload memes or whatever they want in comments. Restricting such features will only serve to frustrate users familiar with other forms of social media.

## Screenshots
- Original - screen shot 2017-11-12 at 9 19 11 pm
- Image helper - screen shot 2017-11-12 at 9 17 35 pm
- File selector - screen shot 2017-11-12 at 9 17 55 pm

PR #2197, opened 2017-12-20
Show the image upload helper on all replies, even comments

### LINKED ISSUE #2196: Use the image upload helper that they see on posts when replying to comments.
Users want to be able to use the image upload helper that they see on posts when replying to comments. Many users are browsing SteemIt on mobile and trying to upload images any other way is a PAIN!

Associated PR: #1975

gl2748:
Closes #2196 
Original PR: #1975


src/app/components/elements/ReplyEditor.jsx
```diff
@@ -538,24 +538,20 @@ class ReplyEditor extends React.Component {
                                             tabIndex={2}
                                         />
                                     </Dropzone>
-                                    {type === 'submit_story' && (
```
gl2748: removing this line is the only change.

PR #2208, opened 2017-12-22
change default sort to hot

### LINKED ISSUE #2207: default trending page very stale
I suspect there's a feedback loop with trending content; we should consider using hot instead, it cycles faster and generally looks more welcoming to guest users.

image

image

roadscape:
close #2207

untested


sneak:
Disagree with the change but am approving it anyway; it will become irrelevant when communities ships.

bnchdrff:
this will be addressed with hive; closing after discussion with @roadscape and @pkattera

PR #2216, opened 2017-12-23
Add Ability To Burn Author Rewards Via UI

netuoso:
### Issue
Currently the only option to denying a payout is to burn ALL rewards. This means curators do not get paid, and the rewards are simply redistributed to the rewards pool.

### Proposal
Add the "Burn Author Rewards" comment option. This option will burn all author rewards to `null` via a beneficiary channel.

### Summary
Since some users are not really dependent on their post earnings, they might wish to burn the author rewards. One reason for this action would be to ensure those rewards are not added to the available supply of STEEM. Burning this STEEM results in decreased supply with an unaffected demand, therefore likely higher market prices.

## Screenshots
screen shot 2017-12-22 at 7 34 04 pm
screen shot 2017-12-22 at 7 54 22 pm
screen shot 2017-12-22 at 7 58 52 pm

### DIFF
--- a/src/app/components/elements/ReplyEditor.jsx
+++ b/src/app/components/elements/ReplyEditor.jsx
@@ -653,7 +653,7 @@ class ReplyEditor extends React.Component {
                                         onChange={this.onPayoutTypeChange}
                                         style={{
                                             color:
-                                                this.state.payoutType == '0%'
+                                                ['0%','nullburn'].indexOf(this.state.payoutType) >= 0
                                                     ? 'orange'
                                                     : '',
                                         }}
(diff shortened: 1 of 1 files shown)


clayop:
Thanks a lot! It is a really reauired feature!

bnchdrff:
@sneak @pkattera input on this?

sneak:
Hard no to burning any tokens in the UI.

sneak:
(Also, the use case for this is already covered under "Decline Payout".)

clayop:
@sneak Declining payout and burning reward are totally different. Burning payout provides users with opportunity to reduce inflation while still give curation rewards, which decline payout doesn't.

iamsmooth:
Perhaps a good middle ground here is to accept the feature into the code base (assuming it is well-implemented and not being rejected on technical grounds) but allow it to be enabled or disabled at the deployment level. So steemit.com could disable it if desired but others using the condenser code base could enable it. 

The open source condenser repo is not intended to be specific to the steemit.com deployment as I understand things. As such, I would expect to see PRs accepted or rejected more on technical grounds than steemit.com policies.

sneak:
Yes, entirely open to that. Please have the setting default to not giving the burn option.

sneak:
Condenser is indeed going to get many sdc-specific features soon, but should gracefully degrade if you run it without our full backend service constellation.

PR #2218, opened 2017-12-23
Urgent (Wallet is broken in Korean language mode)

### LINKED ISSUE #2217: Wallet does not work due to invalid formatting (only in Korean version)
The last Korean translation work came with a defect in the formatting so that thhe Wallet page did not work. I got a fix and will send pull request immediately.

ianpark:
Fixes #2217.

Wallet page does not work in Korean version due to the wrong string formatting in ko.json.
Please review and deliver as soon as possible, as Korean are unable to use the wallet.
Sorry for the trouble.

Issue Number:
#2217 

Workaround:
Change the language to English


ianpark:
@relativityboy @originated Please take a look if you can. Thanks ;)

Issue #2152, opened 2017-12-13
Log errors better


bnchdrff:
currently we use overseer to track browser errors https://github.com/steemit/condenser/blob/master/src/app/Main.js#L14-L16

it doesn't collect enough information -- most errors (~40k per day) come through as `Script error` -- not very helpful!

let's log more details.

also let's log errors in our reducers and maybe other places too.

gl2748:
cf. https://github.com/steemit/condenser/blob/master/src/app/redux/AppReducer.js#L47-L52

gl2748:
refs:
https://github.com/steemit/condenser/issues/2078

relativityboy:
Some addll input from #2209 We want better error reporting and user behavior logging. Based on configuration it should be able to, as desired

Log locally (to the console)
Remotely to some data analysis server
Server target should be configurable outside of git (others may have their own builds of condenser)
Support log levels
Transparently to the log statements include
Relevant information about the client machine eg: features enabled, condenser flavor, etc
Be turned off entirely

relativityboy:
This is a bit of speculation but - some kinds of errors users would consider a bug might trigger a non-blocking UI interface for reporting said bug. The logger could post user feedback right to the logging server. It'd give us a human angle on what the users wanted at the time.

PR #2221, opened 2017-12-23
change htmlready to throw error if xmldom throws an error

### LINKED ISSUE #2178: htmlready test error needs explanation

bnchdrff:
closes #2178

see https://github.com/steemit/condenser/pull/2192 for discussion

DO NOT MERGE until we (especially @roadscape) take a long hard look!

i did quite a bit of testing and auditing and didn't see any circumstances where we give htmlready known-to-crash-xmldom strings & are expecting the raw html to be returned. if we deploy this PR it should be in tandem with #2152 so we can track any errors if they happen.

### DIFF
--- a/src/shared/HtmlReady.js
+++ b/src/shared/HtmlReady.js
@@ -109,9 +109,8 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
             ...state,
         };
     } catch (error) {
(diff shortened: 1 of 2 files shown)


roadscape:
This is an improvement; raising a new exception (as opposed to rethrow) won't show us a trace pinpointing the issue but at least we're not silencing it.

PR #2224, opened 2017-12-26
log stack traces if available

### LINKED ISSUE #2222: log stack trace to overseer
related to #2152 

currently, when an error happens in the browser, we log a message to overseer:

https://github.com/steemit/condenser/blob/410af32615b58ec01f12eb5899edbb29b28907c6/src/app/Main.js#L14-L16

this only logs one part of the error, though, and most errors only get a vague `Script error` logged!

we should log the full stack trace as well as any metadata we can collect from the session - current route and username (if our privacy policy allows it) are easy ones for now; redux state in the future (but that can be later in #2152's work).

i've already completed this work in the `log-errors-better` branch, but it'd be very handy to backport it into its own PR so we can use it immediately:

https://github.com/steemit/condenser/blob/log-errors-better/src/app/Main.js#L14-L18

bnchdrff:
closes #2222

related to #2152 

to test this, you need to do a production build -- it will not work when webpack does a development (sourcemaps included) build.

@jnordberg it looks like overseer is OK with an arbitrary object shape being passed as `metadata`, right? https://github.com/steemit/overseer/blob/master/src/server.ts#L61-L64

it'd probably be wise to decide on a format for these logs -- we can look at sentry for inspiration maybe

### DIFF
--- a/src/app/Main.js
+++ b/src/app/Main.js
@@ -10,10 +10,9 @@ import ConsoleExports from './utils/ConsoleExports';
 import { serverApiRecordEvent } from 'app/utils/ServerApiClient';
 import * as steem from '@steemit/steem-js';
 import { determineViewMode } from 'app/utils/Links';
+import frontendLogger from 'app/utils/FrontendLogger';
 
-window.onerror = error => {
-    if (window.$STM_csrf) serverApiRecordEvent('client_error', error);
-};
+window.addEventListener('error', frontendLogger);
 
 const CMD_LOG_T = 'log-t';
 const CMD_LOG_TOGGLE = 'log-toggle';
--- a/src/app/utils/FrontendLogger.js
+++ b/src/app/utils/FrontendLogger.js
@@ -1,0 +1,45 @@
+import { serverApiRecordEvent } from 'app/utils/ServerApiClient';
+
+/**
+ * Handles window error events by logging to overseer.
+ *
+ * This function relies on these globals:
+ * - process.env.VERSION
+ * - window.location.href
+ *
+ * @param {ErrorEvent} event
+ */
+export default function frontendLogger(event) {
+    if (window.$STM_csrf) {
+        const report = formatEventReport(
+            event,
+            window.location.href,
+            process.env.VERSION
+        );
+        serverApiRecordEvent('client_error', report);
+    }
+}
+
+/**
+ * Format a browser error event report
+ *
+ * @param {ErrorEvent} event
+ * @param {string} href
+ * @param {string} version
+ *
+ * @return {object}
+ */
+export function formatEventReport(event, href, version) {
+    const trace =
+        typeof event.error === 'object' &&
+        event.error !== null &&
+        typeof event.error.stack === 'string'
+            ? event.error.stack
+            : false;
+    return {
+        trace,
+        message: event.message,
+        href,
+        version,
+    };
+}
--- a/src/app/utils/FrontendLogger.test.js
+++ b/src/app/utils/FrontendLogger.test.js
@@ -1,0 +1,35 @@
+import { formatEventReport } from './FrontendLogger';
+
+describe('formatEventReport', () => {
+    it('should handle modern firefox/chrome errors with a stacktrace', () => {
+        const modernErrorEvent = {
+            error: {
+                stack: 'i am a stacktrace',
+            },
+            message: 'i am a message',
+        };
+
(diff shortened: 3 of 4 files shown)


src/app/Main.js
```diff
@@ -11,9 +11,18 @@ import { serverApiRecordEvent } from 'app/utils/ServerApiClient';
 import * as steem from '@steemit/steem-js';
 import { determineViewMode } from 'app/utils/Links';
 
-window.onerror = error => {
-    if (window.$STM_csrf) serverApiRecordEvent('client_error', error);
-};
+window.addEventListener('error', error => {
```
gl2748: Can we name this argument `event`  or `errorEvent` per:
https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

src/app/Main.js
```diff
@@ -11,9 +11,18 @@ import { serverApiRecordEvent } from 'app/utils/ServerApiClient';
...
+    const trace =
+        typeof error.error === 'object' && typeof error.error.stack === 'string'
```
gl2748: It looks like the ErrorEvent.error is an experimental part of the API ... 
Might need an additional condition to poly-fill for it,
for example if I throw a simple string like `throw 'this is just a test'`  there is no `error` object... and this breaks.

gl2748:
# Before
screenshot from 2017-12-27 16-48-32

# After
screenshot from 2017-12-27 16-46-43

I see that it has two event listeners for errors.

src/app/Main.js
bnchdrff: it shouldn't break; it should just define the stacktrace property as `false`. unfortunately there is no way to polyfill a stacktrace.

bnchdrff:
there is `window.onerror` which is what we were using before and then this `window.addEventListener` style I've switched us to... not sure where the additional listener is coming from! what environments are you using for before and after @gl2748?

src/app/Main.js
bnchdrff: good idea; will do

src/app/Main.js
gl2748: Like you need to check that the errorEvent has that object... per docs
error_msg

gl2748:
@bnchdrff before is the live site. After is `yarn run build && yarn run production`

gl2748:
@bnchdrff actually not able to reproduce the double listener now. I thin it is fine. Please ignore.

src/app/Main.js
gl2748: Ey, actually `typeof error.error === object` probably should suffice ... it's just if it is `errorEvent.error` is object and does not have a stack atttribute it will break here.

gl2748: Ok so the thing here is that for some errors the `ErrorEvent.error` is `null` and in this case the double check for: `typeof error.error === 'object' && typeof error.error.stack === 'string'` is insufficient...
demo in console...
```
const e = null
const p = typeof e === 'object' // evals to true
const q = typeof e === 'object' && typeof e.nonExistant === 'string' ? "hello" : "world"
```

bnchdrff: which errors would have a null error?

i've added the test but from what i understand the prop just won't be present if it isn't supported.

here's a test i ran in ie10:

screenshot from 2017-12-28 09-18-20

gl2748: Try:
```
throw "This is a string"
```

bnchdrff: that's a funny way to throw an error ;) anyways i've added a `null` check

gl2748: screenshot from 2017-12-28 09-49-30

gl2748: You can fling strings

bnchdrff:
@jnordberg are we ok passing this data shape to overseer?

jnordberg:
Format looks fine. Need to start signing requests to overseer for it to accept them now but that's a separate issue

bnchdrff:
can you spec that out in another issue @jnordberg ? that's in steemit/steem-js i'm guessing?

bnchdrff:
we need to take into consideration any recent changes made to overseer - @jnordberg do we need to do a signed call for `overseer.collect`? is this the right call to make, even?

bnchdrff:
@gl2748 please review again when you have a moment

gl2748:
Tested, LGTM:
screenshot from 2018-03-06 11-34-05

Issue #1610, opened 2017-08-03
Witness voting does not display properly after casting a vote (Firefox)


TimCliff:
After changing the vote for a witness on the ~witness voting page, the 'upvote' icon sometimes does not change based on the vote. Lots of users seem to have this issue, because I frequently see three witness votes in steemd for users voting on witnesses (vote, unvote, then vote again) all within a few seconds. It seems to only happen the first time a vote is cast in the browser session, and only on Firefox.

After unvoting a witness, the icon was still blue. I clicked on it again and got this error:

![image](https://user-images.githubusercontent.com/20735105/28901111-95fc2716-77bb-11e7-8fb2-ec75fda2dafb.png)

TimCliff:
I recorded a video of this issue:
https://www.youtube.com/watch?v=RJ7w1DFZ7PY

I have been able to consistently reproduce on Firefox, Edge, and Chrome using the following steps:
1. Have an account where at least one witness vote is already cast.
2. Do a fresh load of the witness voting page (to ensure that the new transaction must be signed with the active/owner key).
3. Click the unvote button on a witness vote that is already cast.

After following these steps, the page will show the with the vote still cast, even though it has been removed. At this point you can refresh the page, and the vote will (correctly) disappear.

This seems to intermittently happen when placing a vote too (instead of removing it) but for some reason I can only consistently reproduce when removing a vote. I have no idea why it behaves differently in the two cases, and I see no reason in the code why this would be the case.

The function in question is here:
https://github.com/steemit/condenser/blob/1842b43aef55d58a089f7e2e735fd2c4c0885f31/src/app/components/pages/Witnesses.jsx#L357

I can think of two possible explanations for why the action does not immediately reflect on the page, and a refresh is needed:
1. At the time the page grabs the new data, the action has not been accepted by the blockchain yet.
2. The react dispatch is not properly updating the props for the page. (More info [here](https://stackoverflow.com/questions/39419237/what-is-mapdispatchtoprops))

Any thoughts?

TimCliff:
This appears fixed. Great work @eonwarped !!!!!

PR #2228, opened 2017-12-27
Show disabled/inactive witnesses in a different style on Witness page

netuoso:
### Issue
There is no distinction between a disabled and active witness on the witness page.

### Solution
Add a logic check and change the CSS for the name of the witness.

### Summary
If there are disabled witnesses in the top 50, users are unaware because the Condenser witness page does not show any distinction. This pull request contains code that will perform a basic check on the witness price feed and signing key. If the price feed is blank or the signing key is the disabled key, the witness name will be grey and strikethrough.

### Screenshots
- Without CSS
screen shot 2017-12-26 at 11 08 24 pm
- With CSS
screen shot 2017-12-26 at 11 08 12 pm

### Extra Notes
This pull request should also address https://github.com/steemit/condenser/issues/1610 by adding a `forceUpdate` after applying the witness vote. Sometimes a simple `setState` does not properly trigger the `ComponentShouldUpdate`.

### DIFF
--- a/src/app/components/pages/Witnesses.jsx
+++ b/src/app/components/pages/Witnesses.jsx
@@ -34,6 +34,8 @@ class Witnesses extends React.Component {
         this.onWitnessChange = e => {
             const customUsername = e.target.value;
             this.setState({ customUsername });
+            // Force update to ensure witness vote appears
+            this.forceUpdate();
         };
         this.accountWitnessProxy = e => {
             e.preventDefault();
@@ -76,6 +78,10 @@ class Witnesses extends React.Component {
             const owner = item.get('owner');
             const thread = item.get('url');
             const myVote = witness_votes ? witness_votes.has(owner) : null;
+            const signingKey = item.get('signing_key');
+            const witnessIsDisabled = /STM1111111111111111111111111111111114T1Anm/.test(signingKey);
+            const priceFeed = item.get('sbd_exchange_rate');
+            const noPriceFeed = /0.000 STEEM/.test(priceFeed.get('base'));
             const classUp =
                 'Voting__button Voting__button-up' +
                 (myVote === true ? ' Voting__button--upvoted' : '');
@@ -96,7 +102,9 @@ class Witnesses extends React.Component {
                 }
             }
             return (
-                <tr key={owner}>
+                <tr
+                    key={owner}
+                    style = { witnessIsDisabled || noPriceFeed ? { opacity: '0.4' } : null }>
                     <td width="75">
                         {rank < 10 && '0'}
                         {rank++}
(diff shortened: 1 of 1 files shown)


src/app/components/pages/Witnesses.jsx
```diff
@@ -76,6 +78,10 @@ class Witnesses extends React.Component {
             const owner = item.get('owner');
             const thread = item.get('url');
             const myVote = witness_votes ? witness_votes.has(owner) : null;
+            const signingKey = item.get('signing_key');
+            const witnessIsDisabled = /STM1111111111111111111111111111111114T1Anm/.test(signingKey);
```
relativityboy: This needs to be a constant with a good name and explanation. 

Anything that _might-ever be referenced by other components_ probably should go in src/shared/constants otherwise @ the top of the file.

src/app/components/pages/Witnesses.jsx
```diff
@@ -76,6 +78,10 @@ class Witnesses extends React.Component {
...
+            const priceFeed = item.get('sbd_exchange_rate');
+            const noPriceFeed = /0.000 STEEM/.test(priceFeed.get('base'));
```
relativityboy: and 0.000 STEEM - also, is this strict enough? If this is a match over the entire string choose the faster of 
1. a good regex
2. ===

src/app/components/pages/Witnesses.jsx
```diff
@@ -96,7 +102,9 @@ class Witnesses extends React.Component {
                 }
             }
             return (
-                <tr key={owner}>
+                <tr
+                    key={owner}
+                    style = { witnessIsDisabled || noPriceFeed ? { opacity: '0.4' } : null }>
```
relativityboy: Needs to be in a stylesheet.

src/app/components/pages/Witnesses.jsx
```diff
@@ -115,7 +123,7 @@ class Witnesses extends React.Component {
                             </a>
                         </span>
                     </td>
-                    <td>
+                    <td style = { witnessIsDisabled || noPriceFeed ? { textDecoration: 'line-through' } : null }>
```
relativityboy: Needs to be in a stylesheet.

relativityboy: @netuoso - The intent of this change is a good one. We have [new contribution guidelines](https://github.com/steemit/condenser/blob/master/CONTRIBUTING.md#enhancement-suggestions). Please check them out and follow them for future PRs.

We have a lot of work coming down the pipe for the steemit team. For the next couple months feature adds by community members will be slower to be integrated, and PR reviews will need to be corrected by the contributors somewhat quickly or risk being closed (likely to be revisited later)

bnchdrff:
@netuoso we're interested in incorporating this idea ; feel free to re-open when you have time to respond to our feedback! thanks!

pkattera:
@netuoso Sorry for the delayed response. This is a great suggestion and we should get it in condenser. 

While I agree that we should disable witnesses who have their signing key disabled, I'm not sure about disabling witnesses without a price feed.

 In addition, can we please add a tooltip on hover anywhere along the disabled row (like an image alt tag), that says, 'This witness is currently not active', to give a little bit more context to end users.

roadscape:
Review continues in local branch at #2828

PR #2232, opened 2017-12-27
Test global reducer

### LINKED ISSUE #2160: Create Tests for the Reducers.

gl2748:
Fixes #2160

Rebased here - #2402


src/app/redux/GlobalReducer.js
```diff
@@ -352,18 +352,16 @@ export default function reducer(state = defaultState, action) {
         case CLEAR_META: {
             return state.deleteIn(['metaLinkData', payload.id]);
         }
-
+        // does not have an action creator...
```
bnchdrff: ooooooops i missed that one!!!! i'll patch it into this PR

src/app/redux/GlobalReducer.test.js
```diff
@@ -0,0 +1,719 @@
...
+    hasPendingPayout: false,
+    gray: false,
+    flagWeight: 0,
+    up_votes: 0,
+    total_votes: 0,
+    authorRepLog10: undefined,
+    allowDelete: true,
+});
+
+const mockPayloads = {
```
bnchdrff: should we define these payloads at the top like this, or define them inline with each test? having them inline makes each test a bit easier to read.

PR #2235, opened 2017-12-28
run component tests with jest & enzyme

### LINKED ISSUE #2203: Test components
Figure out how to test components, and set up some examples.

bnchdrff:
closes #2203

- modify existing component test to work with jest runner
- remove mocha-related test libs
- remove defunct tests
- port existing unit tests to jest / jasmine


bnchdrff:
note: this PR includes a `react` & `react-dom` version update from 15.4 to 15.6.2 -- this is in order to get some react-dom capabilities necessary for component tests. also it's just a good idea :)

src/app/components/modules/Header.test.js
```diff
@@ -1,11 +1,18 @@
 /*global describe, it, before, beforeEach, after, afterEach */
 import React from 'react';
-import { expect } from 'chai';
-import { shallow } from 'enzyme';
+import { configure, shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15';
```
gl2748: If we add a `jestSetup.js` file per this commit  then we will not have to import the Adapter in every test...
https://github.com/steemit/condenser/commit/4f233a29ad734b4bccb13acd0937ef590bed8786

bnchdrff: ooh good call, will `cherry-pick`

bnchdrff: actually i don't think we should run these before every test -- they are only necessary for component tests & will slow down unit tests

gl2748: Sounds good.

PR #2259, opened 2018-01-02
add create_user endpoint

### LINKED ISSUE #2243: Add endpoint to create users
The endpoint should be password protected (set in env var) and take username email and insert it into the database to be used by the old account recovery system until it is replaced.

bnchdrff:
closes #2243

note that this requires an env var `CREATE_USER_SECRET` to be defined at condenser runtime


src/server/api/general.js
```diff
@@ -269,6 +269,50 @@ export default function useGeneralApi(app) {
...
+                name: esc(name),
+            });
+            if (existingUser) {
+                this.body = JSON.stringify({
+                    success: false,
+                    error: 'user with this email or name already exists',
+                });
+                this.status = 400;
+            } else {
+                const user = yield models.User.create({
```
jnordberg: This isn't enough to create an user that can be recovered. We need an `Account` instance created and attached to the `User` as well as an `Identity` instance with the email.

bnchdrff: ah, ok. i thought you just wanted the user record. for the Account row, what should we use for `owner_id`? and for Identiy what should we use for `provdier` and `provider_id`? these columns are referenced in the account recovery code.

jnordberg: For Identity set it up as an verified email, not sure about owner_id check what the creation process is doing now and set it to the same. This endpoint should do the same thing as the last step of the current signup process minus the blockchain account creation which will have happened already when this endpoint is hit by faucet

bnchdrff: sorry i meant `owner_key` -- it's generated here: https://github.com/steemit/condenser/blob/master/src/app/components/pages/CreateAccount.jsx#L86-L114

bnchdrff: i pushed a change to create the `account` and `identity` rows -- no keys yet

jnordberg: Ah, good question. We might need to pass the public key along from faucet as well...

sneak:
I am closing this pull request.  @bnchdrff, please conduct more through testing on this branch (perhaps by deploying it in dev, or writing some unit tests) if you feel it's not ready to be merged and deployed into production.

PRs should only be opened when the pull requesting user thinks the code is production-ready.  If you think it needs a WIP tag, close the PR and open it when it's fully baked.

sneak:
FYI the branch remains, so continue your work and reopen when ready.

jnordberg:
Resurrecting this

PR #2273, opened 2018-01-04
init FeatureFlag HOC

### LINKED ISSUE #2179: Add feature flags (ability to turn aspects of condenser on and off)
# Subissues

* [x] create feature flag HoC `Flag`: model its API after unbounce/flag
* [x] add feature flag `Map` to our state tree, probably next to other user settings in AppReducer; model its shape after unbounce/flag
* [x] create a `withFlagState` container for `Flag` which subscribes to the flag part of the state tree - #2273 
* [x] update steem-js to 0.7.1 so we can fetch flags (done in PR)
* [ ] add steemitapi.get_feature_flags to jussi
* [ ] figure out what is blocking the get_feature_flags endpoint & solve any issues
* [x] add saga to fetch flags - saga should deal gracefully with the flag endpoint not working at all #2375 (done in PR)
* [x] figure out issue with initial state not having defaults #2377 (done in PR)
* [x] add documentation w/ example #2376

It would be really nice to have "feature flags" that would either be part of condenser's configuration, or found via an API call. The flags would allow us to turn features (ex for non-extant features: notifications, live comment count/second tickers) on or off for a given instance of condenser.

The API call component of this would be handled by [conveyor](https://github.com/steemit/conveyor) (formerly steemitapi). It was [issue 5](https://github.com/steemit/conveyor/issues/5) and [PR 4](https://github.com/steemit/conveyor/pull/4) in that codebase.

JSON-RPC Request Example:
```
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "conveyor.get_feature_flags",
    "params": {"username": "jnordberg"}
}
```
The return value is an object with key value pairs `{flag_name: boolean}`. Absence of a flag is equivalent to false.

As a part of condenser's local configuration, they would mimic the structure of the RPC call's response, and deferring to the RPC result unless a css like _!important_ was added to a flag (mechanism tbd). There will also be a local setting for turning off the API call entirely; relying only on the local config.

**Logging & Other**
This Issue only covers the consumption of the feature flags themselves, and enabling/disabling the associated features. It is not intended to include special capabilities _implied_ by feature flags.

gl2748:
Fixes #2179

### DIFF
--- a/src/app/components/modules/Flag/Flag.test.jsx
+++ b/src/app/components/modules/Flag/Flag.test.jsx
@@ -1,0 +1,31 @@
+/*global describe, it, before, beforeEach, after, afterEach */
+import React from 'react';
+import { configure, shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15';
+import LoadingIndicator from 'app/components/elements/LoadingIndicator';
+import Icon from 'app/components/elements/Icon';
+
+import Flag from 'app/components/modules/Flag';
+
+configure({ adapter: new Adapter() });
+
+describe('Flag', () => {
+    let component = LoadingIndicator;
+    let fallback = Icon;
+    it('should render the component only when the flag condition passes', () => {
+        const FlagComponent = Flag(component);
+        let wrapper = shallow(<FlagComponent flag={true} />);
+        expect(wrapper.html()).not.toBe(null);
+    });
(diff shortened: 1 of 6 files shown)


bnchdrff:
let's do the whole thing in one PR once it's all ready, closing this for now

Issue #2266, opened 2018-01-03
Comment with <?php crashes steemit.com


DoctorLai:
I was trying to place a comment that contains a PHP code:

```
<?php
  echo base64_decode("GVsbG8gQ29sZEhhaXJ+"); 
```

and on steemit.com  it keeps loading, after 1 minute, I refresh the page, and the steemit.com crashes  https://steemit.com/cn-stats/@dailychina/cn--2017-12-31

![image.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1514720803/azj4h0daefza0svhxydx.png)

And if I open busy.org  it can load, but the comment displays as:

![image.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1514720836/xiy5lzn5rymr5s9ssysx.png)

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@justyy/comment-with-php-crash-the-steemit-com">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

netuoso:
https://github.com/steemit/condenser/pull/2275

bnchdrff:
fixed in #2277

PR #2275, opened 2018-01-04
Strip php tags from posts, comments, and replies

netuoso:
## Issue
https://github.com/steemit/condenser/issues/2266
- Comments/posts that include a `<?php>` tag do not render.

## Solution
Strip the `<?php>` tags from Comment/Post body before rendering with MarkdownViewer

## Summary
If a user submits a post or comment with `<?php` in the body, the entire page will be broken because it will encounter an error in the `xmldom` DOMParser. If we simply strip the `<?php>` tags, the page will load as expected. This pull request implements a simple solution where the `<?` are removed from the tag rendering it safe.

## Screenshots
screen shot 2018-01-04 at 4 33 07 pm

### DIFF
--- a/src/app/components/cards/Comment.jsx
+++ b/src/app/components/cards/Comment.jsx
@@ -287,10 +287,11 @@ class CommentImpl extends React.Component {
         let controls = null;
 
         if (!this.state.collapsed && !hide_body) {
+            var sanitizedBody = comment.body.replace(/(\<\?{0,}php)/, 'php');
             body = (
                 <MarkdownViewer
                     formId={post + '-viewer'}
-                    text={comment.body}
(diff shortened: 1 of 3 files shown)


bnchdrff:
fixed in #2277

src/app/components/cards/Comment.jsx
```diff
@@ -287,10 +287,11 @@ class CommentImpl extends React.Component {
         let controls = null;
 
         if (!this.state.collapsed && !hide_body) {
+            var sanitizedBody = comment.body.replace(/(\<\?{0,}php)/, 'php');
```
mktcode: Wouldn't this simply change all tags, including those in \`code` formatting for example?

PR #2277, opened 2018-01-04
return a message rather than throwing if xmldom cannot parse

### LINKED ISSUE #2276: rather than throwing an error, return something when an xmldom error is caught
modify the behavior we set up in #2221 so that htmlready will return an empty string or something if xmldom chokes

bnchdrff:
closes #2276

we should log these

### DIFF
--- a/src/shared/HtmlReady.js
+++ b/src/shared/HtmlReady.js
@@ -110,7 +110,8 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
         };
     } catch (error) {
         // xmldom error is bad
-        return { html: 'rendering error' };
+        console.log('rendering error', { error: error.message, html });
+        return { html: '' };
     }
 }
 
--- a/src/shared/HtmlReady.js
+++ b/src/shared/HtmlReady.js
@@ -110,7 +110,10 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
         };
     } catch (error) {
         // xmldom error is bad
-        console.log('rendering error', { error: error.message, html });
+        console.log(
+            'rendering error',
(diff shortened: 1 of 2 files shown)


bnchdrff:
in the case of that post linked in #2266 here's what happens:

screenshot from 2018-01-04 18-27-27

what should we display for posts like this?

src/shared/HtmlReady.js
```diff
@@ -110,7 +110,7 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
         };
     } catch (error) {
         // xmldom error is bad
```
jnordberg: Good idea to log it, not elegant but a a simple `console.log('RENDER ERROR: '+error.message)` would let us keep track of it server-side at least

jnordberg:
`<span class="render-error>Rendering error</span>` and color it red or something to differentiate it from a working comment

jnordberg:
Or just an empty string would do imo

src/shared/HtmlReady.js
bnchdrff: i just set it up to also log the problematic html... this may or may not be a good idea ;)

src/shared/HtmlReady.js
```diff
@@ -110,7 +110,8 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
...
-        throw new Error('HtmlReady: xmldom error');
+        console.log('rendering error', { error: error.message, html });
```
jnordberg: Do a `JSON.encode` here so we are sure to get it on a single line in the logs

bnchdrff: smart; done

PR #2280, opened 2018-01-05
Hebrew Translation and RTL alignment

doriitamar:
I have translated the platform to Hebrew + made some CSS adjustments so RTL content will be shown correctly.

All of my changes were quite straightforward and the biggest was to change the `dir` attribute on the `<html>` tag.


src/app/components/modules/TopRightMenu.jsx
```diff
@@ -117,7 +118,7 @@ function TopRightMenu({
                 <LinkWithDropdown
                     closeOnClickOutside
                     dropdownPosition="bottom"
-                    dropdownAlignment="right"
+                    dropdownAlignment={alignment || 'right'}
```
doriitamar: I noticed nobody ever used the `alignment` prop except me and it was hardcoded as always `"right"`

src/app/components/modules/Settings.jsx
```diff
@@ -157,6 +157,12 @@ class Settings extends React.Component {
         const locale = event.target.value;
         const userPreferences = { ...this.props.user_preferences, locale };
         this.props.setUserPreferences(userPreferences);
+
+        if (locale === 'he') {
```
doriitamar: I'm not sure that this piece of code is correctly written. Yeah, it works, but I do use the DOM directly..

src/app/components/modules/Header.jsx
```diff
@@ -82,12 +82,12 @@ class Header extends React.Component {
         const current_account_name = this.props.current_account_name;
         let home_account = false;
         let page_title = route.page;
-
         let sort_order = '';
         let topic = '';
         let user_name = null;
         let page_name = null;
         this.state.subheader_hidden = false;
+        this.context.intl = this.context.intl || { locale: 'en' };
```
doriitamar: It's a redundant initialization though required to pass the tests. See https://gist.github.com/mirague/c05f4da0d781a9b339b501f1d5d33c37/

gl2748:
Thanks for this @doriitamar! I am tagging @originated as they are handling all of our translation work.

src/app/components/modules/Settings.jsx
bnchdrff: while your code will definitely cause the attribute to change after a user changes their language to hebrew, the setting won't persist to the next session. i think the right way to do this would be to connect the attribute to [the redux state](https://github.com/steemit/condenser/blob/master/src/app/redux/AppReducer.js#L34).

src/server/server-html.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 export default function ServerHTML({ body, assets, locale, title, meta }) {
     let page_title = title;
     return (
-        <html lang="en">
+        <html lang={locale} dir={locale == 'he' ? 'rtl' : 'ltr'}>
```
bnchdrff: good catch on the `lang` attribute!

originated:
I'm currently closing this issue until our current translation additions are ready to not cause any conflicts.  We'll re-open this ticket when this language has been properly added/on-boarded into Transfix as well.  Great work @doriitamar

PR #2293, opened 2018-01-06
Code div blocks are now overflow: scroll (issue #2290)

### LINKED ISSUE #2290: code div css should scroll
The code div blocks shouldn't clip the background, instead they should be set to overflow: scroll

I've made this change in my local copy of the condenser, but I don't know what a new_grass label is or how to "tag @sneak" so I'm just going to leave this here...

https://github.com/not-a-bird/condenser/commit/d8aaea60c146624717278eebe1b03a378cef4382

not-a-bird:
Fixes #2290.

[Edit from @sneak - our internal process for PRs is that they should always open with the issue number they fix - I added the line above for you.]

I was getting annoyed at how scrolling in a code div would clip the
background and basically make it impossible to read what was there.
This tiny addition (smaller than this comment, in fact) fixes that.

Here's the before image, notice how the gray background ends, but the code block continues past it?

before

And here is the after, note that it correctly continues now with the text:
after


not-a-bird:
I would think everything code.

Is there something specific that it interferes with?
-------- Original Message --------

julienbh:
Not that I know of... Maybe someone who knows more the code than me could chime in.

sneak:
I think it is reasonable to put this on all `<code>` blocks.

sneak:
PS: @not-a-bird - thank you for the issue and PR!  We really appreciate stuff like this from the community.

bnchdrff:
this is a good start; we can add more `overflow: scroll`s to other `code`s if needed (probably just in comments)

PR #2302, opened 2018-01-08
Add buildwhale to BadActorList.js

themarkymark-steem:
Add scam account buildwhale to bad actor, account is used to catch misspellings


pfunks:
I've updated my PR here to include buildawhale impostor accounts, #2328

PR #2304, opened 2018-01-08
Chinese menu

originated:


src/app/Translator.js
```diff
@@ -7,10 +7,11 @@ import ru from 'react-intl/locale-data/ru';
 import fr from 'react-intl/locale-data/fr';
 import it from 'react-intl/locale-data/it';
 import ko from 'react-intl/locale-data/ko';
+import zu from 'react-intl/locale-data/zu';
 import { DEFAULT_LANGUAGE } from 'app/client_config';
 import tt from 'counterpart';
 
-addLocaleData([...en, ...es, ...ru, ...fr, ...it, ...ko]);
+addLocaleData([...en, ...es, ...ru, ...fr, ...it, ...ko, ...zu]);
```
bnchdrff: why are we using this langcode? also is this chinese simplified or traditional?

originated: @bnchdrff This is simplified @valzav would have better insight for that bit of code.

bnchdrff: chinese simplified is `zh-Hans` not `zu`which is Zulu -- if we plan on having a chinese traditional localization (`zh-Hant`) we should maybe rename this translation set to `zh-Hans`. if we don't plan on having another chinese dialect, we can just name it `zh`.

PR #2311, opened 2018-01-09
Use a regular popup & share link rather than fb-sdk for facebook sharing

### LINKED ISSUE #1990: Only show share buttons after a user interaction
Avoid loading the Facebook SDK or showing the share buttons until the user shows some sort of intention to use them.

bnchdrff:
Closes #1990 

This sharer link on the fb side might stop working someday.

For the record, we'd previously delayed loading of the fb sdk but that made the popup get blocked more frequently (https://github.com/steemit/condenser/pull/1151).


bnchdrff:
this is a good option if we're OK not loading the SDK at all -- i.e. we aren't using fb for analytics or something else.

PR #2317, opened 2018-01-10
remove tarantool and related functionality

### LINKED ISSUE #2309: spike: audit tarantool usage
how do we currently use tarantool? what changes have to happen in order for old notifications to get mothballed? how might these changes effect other usage?

related to #2102

bnchdrff:
closes #2309 (if we want it to!)

this leaves quite a bit of notification frontend code in place, as there are some ambiguous dependencies that i don't really want to dig into right now with non-tarantool-related notifications generated in browser.


src/app/redux/PollDataSaga.js
```diff
@@ -1,41 +1,17 @@
...
-                    webpush_params = { error };
-                }
-            }
-            const nc = yield call(getNotifications, username, webpush_params);
-            yield put(appActions.updateNotificounters(nc));
-        }
-
         try {
-            yield call([api, api.getDynamicGlobalPropertiesAsync]);
+            yield call([api, api.getDynamicGlobalPropertiesAsync]); // TODO why?????
```
bnchdrff: i'm pretty sure we can just get rid of polldatasaga at this point

bnchdrff: this call never puts anything. it used to: https://github.com/steemit/condenser/blob/45267c401ecb002462e9bb9d9c6fbf3d1d2a9621/src/app/redux/PollDataSaga.js#L38

bnchdrff:
@jredbeard let's discuss when you're available

relativityboy:
FTR, the only concern I had with this was removing the Tarantool code for account signup, and the requirement to have our signup process be no less reliable than it is now; so we can use it as a backup for the new faucet.

If we're set phase out Tarantool completely, this looks great.

src/server/api/general.js
```diff
@@ -124,32 +123,11 @@ export default function useGeneralApi(app) {
...
-                '-- /accounts tarantool is not available, fallback to another method',
-                e
-            );
-            const rnd_wait_time = Math.random() * 10000;
-            console.log('-- /accounts rnd_wait_time -->', rnd_wait_time);
-            yield new Promise(resolve =>
-                setTimeout(() => resolve(), rnd_wait_time)
-            );
-        }
+        const rnd_wait_time = Math.random() * 10000;
```
goldibex: wat

goldibex: Seriously, what? is this supposed to be a spinlock? i get that this isn't your code, but exactly what is happening here?

goldibex: This is not even remotely idiomatic Javascript. Can you please repair this?

bnchdrff: +1 idk why it has to be random. this is the status quo, though.

bnchdrff: just discussed this on the call; the random wait can be removed.

valzav: @goldibex, I've added this as fallback solution for those rare cases when Tarantool db is down/not available. I admit its pure quality from idea to implementation - I was time constrained at this moment and the issue I was trying to fix was a real exploit costing us a couple of hundred bot accounts registered each day.

src/server/api/general.js
```diff
@@ -133,6 +134,18 @@ export default function useGeneralApi(app) {
...
             }
 
+            // Is an account creation already happening for this user, maybe on another request?
+            // If so, throw an error.
+            if (user.creation_hash !== null) {
+                throw new Error('An account creation is in progress');
+            }
+            // If not, set this user's creation_hash.
+            // We'll check this later on, just before we create the account on chain.
+            const creationHash = hash
```
goldibex: couldn't this just be `crypto.randomBytes(32)`? the first rule of entropy exhaustion is  use as little entropy as you can get away with

src/server/api/general.js
```diff
@@ -166,17 +179,38 @@ export default function useGeneralApi(app) {
...
+                    new_account_name: account.name,
+                    delegation: config.get('registrar.delegation'),
+                    owner: account.owner_key,
+                    active: account.active_key,
+                    posting: account.posting_key,
+                    memo: account.memo_key,
+                });
+            } catch (e) {
+                yield user.update({ creation_hash: null });
+                throw new Error('Account creation error, try again.');
```
goldibex: when will this execute? I'm not 100% familiar with JS generator syntax, but in Python, execution would stop right here and you wouldn't reach the `throw` until the next invocation of the method

src/server/api/general.js
bnchdrff: i was kinda wondering how big to make this :) i'll change it to 32

src/server/api/general.js
bnchdrff: this generator method is getting called over and over again, so the `throw` happens after the call to `user` is done -- doing the `yield` just lets Koa will move on to other work if the sql takes a minute

bnchdrff:
to test locally, apply this diff & create a user (if needed):

```diff
diff --git a/src/server/api/general.js b/src/server/api/general.js
index 03b76a1..5f8c873 100644
--- a/src/server/api/general.js
+++ b/src/server/api/general.js
@@ -105,7 +105,7 @@ export default function useGeneralApi(app) {
         const params = this.request.body;
         const account =
             typeof params === 'string' ? JSON.parse(params) : params;
-        if (!checkCSRF(this, account.csrf)) return;
+        //if (!checkCSRF(this, account.csrf)) return;
         logRequest('accounts', this, { account });
         if ($STM_Config.disable_signups) {
             this.body = JSON.stringify({
@@ -115,7 +115,37 @@ export default function useGeneralApi(app) {
             return;
         }
 
-        const user_id = this.session.user;
+        //const user_id = this.session.user;
+        // create a user in your local db with values like this:
+/*
+mysql> select * from users where id = 8\G
+*************************** 1. row ***************************
+            id: 8
+          name: mfmmppfoop
+         email: fbarm@barm.mmbarmmmmmmmmmmmmmmmm
+           uid: NULL
+    first_name: NULL
+     last_name: NULL
+      birthday: NULL
+        gender: NULL
+ picture_small: NULL
+ picture_large: NULL
+   location_id: NULL
+ location_name: NULL
+        locale: NULL
+      timezone: NULL
+      verified: NULL
+           bot: NULL
+    created_at: 2018-01-02 17:54:39
+    updated_at: 2018-01-11 22:18:37
+  waiting_list: NULL
+     remote_ip: NULL
+account_status: approved
+  sign_up_meta: NULL
+ creation_hash: NULL
+1 row in set (0.00 sec)
+*/
+                               const user_id = 8;
         if (!user_id) {
             // require user to sign in with identity provider
             this.body = JSON.stringify({ error: 'Unauthorized' });
@@ -195,6 +225,8 @@ export default function useGeneralApi(app) {
             }
 
             try {
+                // either comment this out (to simulate a successful creation)
+                // or just let it run and it will throw (to simulate an onchain error)
                 yield createAccount({
                     signingKey: config.get('registrar.signing_key'),
                     fee: config.get('registrar.fee'),

```

then do a request like this:

```
curl -d '{"name":"accountname","owner_key":"bleh","active_key":"activekey","posting_key":"postingkey","memo_key":"memokey"}'  -H 'content-type: application/json' http://localhost:8080/api/v1/accounts
```

PR #2328, opened 2018-01-12
Update BadActorList.js

pfunks:
I alphabetized the list for easy checking/insertion of new accounts in the future.

Many new accounts impersonating/typofishing exchanges have been added, along with some prominent Steem service and user typofishing accounts.  The accounts that could be considered borderline such as `@bitt` were checked to make sure they were not a real users account, and as it turns out a scammer was successful at taking someone's funds meant for Bittrex in that case.


bnchdrff:
thank you @pfunks & sorry for the delay in merging all of these in; your work is very much appreciated!

bnchdrff:
also good call on the alphabetization, that'll make maintenance of this list much easier

Issue #2167, opened 2017-12-15
figure out why #2113 passed tests but wasn't deployable, fix tests to fail in this case


sneak:
For some reason #2113 passed tests but wasn't deployable.  Make sure that whatever caused it to be undeployable also causes it to fail tests/fail build.

goldibex:
The packages `medium-editor-insert-plugin` (which was the subject of #2113) and `blueimp-file-upload` are both direct dependencies of condenser as specified [here](https://github.com/steemit/condenser/tree/master/webpack/base.config.js#L64). This has been the case since the initial commit of the repository (06c90aa).

`medium-editor-insert-plugin` and `blueimp-file-upload` as packages are both referred to by  `webpack/base.config.js` using `require.resolve`. This effectively conceals them from any normal static analysis tool.

On 7 December 2016, in commit ad8ffb7, @jcalfee removed the dependency `blueimp-file-upload` from `package.json` without removing the corresponding `require.resolve` from `webpack/base.config.js`. This left condenser in an invalid state; starting from ad8ffb7, the package `blueimp-file-upload` only existed as a transitive dependency of `medium-editor-insert-plugin`.

On 28 October, in commit 6c137dd, @roadscape merged PR #551, previously approved by @jcalfee, which claimed to remove the usage of `medium-editor-insert-plugin` from the codebase. This PR was erroneous in that it neither removed `medium-editor-insert-plugin` nor `blueimp-file-upload` from `package.json`, nor did it remove the `require.resolve` entries from `webpack/base.config.js`.

From 6c137dd onward, both `medium-editor-insert-plugin` and `blueimp-file-upload` were hanging dependencies. They were not referred to by any code inside condenser, but their presence was requested in an inconsistent and irregular way by `base/webpack.config.js`.

@jnordberg noticed that `medium-editor-insert-plugin` was not referred to as an import anywhere in the codebase and opened #2113 to remove it. He should have examined the entire repository, not just the `src` directory, for references to the file. If he had he would have discovered the hanging references in `base/webpack.config.js`.

Our automated test suite continued to pass after #2113 was merged because the unit tests as currently configured do not rely on Webpack. This is not unusual for a React/ES6 application, as Webpack is exceptionally slow and in some instances takes more than a minute to finish a build.

I therefore submit the following proposals for additional followup work, in order of priority:
- [ ] Remove the references to `medium-editor-insert-plugin` and `blueimp-file-upload` from `webpack/base.config.js`.
[...]

PR #2340, opened 2018-01-14
Revert "Merge pull request #2274 from steemit/revert-2235"

### LINKED ISSUE #2203: Test components

bnchdrff:
closes #2203

This reverts commit 1c1451275e1f22f55af7cf6c7ae39e427c11e67a, reversing
changes made to a665b02e812469051a1aae5f4bbdcfb8434cb349.


gl2748:
Looking at this now...

gl2748:
For whatever reason, if you have a locally built version of the site at `/lib` the tests will run against that code and fail. When testing locally and checking our this branch for the first time, delete `lib` dir:
Example error:
```
 FAIL  lib/shared/HtmlReady.test.js
  ● Test suite failed to run

    Cannot find module 'chai' from 'HtmlReady.test.js'

       5 | var _from2 = _interopRequireDefault(_from);
       6 | 
    >  7 | var _chai = require('chai');
       8 | 
       9 | var _chai2 = _interopRequireDefault(_chai);
      10 | 
      
      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:194:17)
      at Object.<anonymous> (lib/shared/HtmlReady.test.js:7:13)

```

bnchdrff:
we need to exclude `lib/` from jest; i'll work on that

bnchdrff:
oh
actually.... shouldn't the tests pass from `lib` anyways? is your `lib` build stale?

bnchdrff:
the compiled tests pass for me.. try `yarn build && yarn test`

gl2748:
Yarn run test should probably run build if it's testing the `/lib` dir?

bnchdrff:
jest will run any *.test.js it finds -- if you've got a build in `lib` it'll run them in there also -- i think this is good

gl2748:
I think we should add a jest config to this work. I think a reasonable expectation is that tests are running against one version of the code and that version is what's in `/src`.
https://facebook.github.io/jest/docs/en/configuration.html

gl2748:
https://facebook.github.io/jest/docs/en/configuration.html#rootdir-string

bnchdrff:
@gl2748 ok we will avoid testing `lib/` -- i changed our jest config

gl2748:
@bnchdrff Actually, am i right in this, we had issues with Mocha not testing code as compiled ... perhaps testing lib is the right way to go after all?

bnchdrff:
if we want to mandate testing of transpiled code in `lib` we can do that in another PR; the status quo is to let the test runner transpile when it runs

gl2748:
Ok, sounds alright. Here is the issue that referenced the divergence between our code and the tests:
https://github.com/steemit/condenser/issues/2167

package.json
```diff
@@ -199,5 +195,14 @@
       "prettier --write",
       "git add"
     ]
+  },
+  "jest": {
+    "rootDir": "src",
+    "modulePaths": [
+      "<rootDir>/"
```
relativityboy: not `<rootDir>/src`? 

`https://github.com/steemit/condenser/pull/2235/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R200`

package.json
```diff
@@ -199,5 +195,14 @@
...
+    ],
+    "moduleNameMapper": {
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js"
+    }
   }
 }
```
relativityboy: None of the versions are `^`. In #2235 they were.

package.json
bnchdrff: yup; changed the rootDir above in e60d6bf

package.json
bnchdrff: correct; we're moving to pinned dependency versions and this is one step in that direction

PR #2346, opened 2018-01-15
deal with hard to test functions a bit more gracefully

bnchdrff:
i did a bad job of porting iain's mocha test workarounds for TransactionSaga's impure (timestamp-related) code; this is a better way of doing it.


gl2748:
This is good, later on we can make these hard to test functions pure.

PR #2347, opened 2018-01-15
Add debounce for onChange validation #2282

### LINKED ISSUE #2282: Memo field in transfer dialog gets slower and slower
There is something going on in the memo field.
I noticed this before, if you type a long memo field it gets slower and slower as you type.

Second demo

The cpu profiler shows some time spent in one section 
image

Putting a breakpoint I get this being called many times :
image

```
value (vendor.018876a8…ormatted:11103)
a (app.018876a897e71d4c1f78.js:1)
validation (app.018876a897e71d4c1f78.js:1)
a (app.018876a897e71d4c1f78.js:1)
(anonymous) (app.018876a897e71d4c1f78.js:1)
t.notifyAll (vendor.018876a8…ormatted:18624)
close (vendor.018876a8…formatted:7046)
closeAll (vendor.018876a8…ormatted:11226)
perform (vendor.018876a8…ormatted:11199)
perform (vendor.018876a8…formatted:7062)
S (vendor.018876a8…formatted:7070)
closeAll (vendor.018876a8…ormatted:11226)
perform (vendor.018876a8…ormatted:11199)
batchedUpdates (vendor.018876a8…ormatted:19228)
o (vendor.018876a8…formatted:6979)
dispatchEvent (vendor.018876a8…ormatted:33284)
```

```
 {
                key: "toPublicKey",
                value: function() {
                    return this.toPublic()
                }
            }
```

The more text I have the more this breakpoint fires.

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@h4ck3rm1k3st33m/memo-field-in-transfer-dialog-gets-slower-and-slower">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

relativityboy:
Closes #2282

### DIFF
--- a/src/app/utils/ReactForm.js
+++ b/src/app/utils/ReactForm.js
@@ -5,6 +5,8 @@
     @arg {object} initialValues required for checkboxes {save: false, ...}
     @arg {function} validation - values => ({ username: ! values.username ? 'Required' : null, ... })
 */
+const USER_INPUT_DEBOUNCE_MS = 400;
+
 export default function reactForm({
     name,
     instance,
@@ -12,6 +14,7 @@ export default function reactForm({
     initialValues,
     validation = () => {},
 }) {
+    let debounce = { timeout: null };
     if (typeof instance !== 'object')
         throw new TypeError('instance is a required object');
     if (!Array.isArray(fields))
@@ -121,7 +124,13 @@ export default function reactForm({
             }
 
             instance.setState({ [fieldName]: v }, () => {
-                setFormState(name, instance, fields, validation);
+                setFormStateDebounce(
+                    name,
+                    instance,
+                    fields,
+                    validation,
+                    debounce
+                );
             });
         };
 
@@ -132,6 +141,13 @@ export default function reactForm({
             instance.setState({ [fieldName]: v });
         };
     }
+}
+
+function setFormStateDebounce(name, instance, fields, validation, debounce) {
+    clearTimeout(debounce.timeout);
+    debounce.timeout = setTimeout(function() {
+        setFormState(name, instance, fields, validation);
+    }, USER_INPUT_DEBOUNCE_MS);
 }
 
 function setFormState(name, instance, fields, validation) {


relativityboy:
> Is there a reason why there needs to be onChange validation on this form?

That was a call made when the form was designed. Probably to give the users real-time feedback. My guess is that it makes the site a bit friendlier to use when it's not slowing down input.

To alter that would be more than a bugfix.

src/app/utils/ReactForm.js
```diff
@@ -121,7 +124,13 @@ export default function reactForm({
             }
 
             instance.setState({ [fieldName]: v }, () => {
-                setFormState(name, instance, fields, validation);
+                setFormStateDebounce(
```
bnchdrff: why should we debounce the form state setter, rather than [the validation function in Transfers.jsx](https://github.com/steemit/condenser/blob/master/src/app/components/modules/Transfer.jsx#L172)?

bnchdrff: (asking because this change will affect all users of our ReactForm module)

relativityboy: I considered that.
The validation routine expects validators to be synchronous calls (and if well written, side-effect free).
Adding a debounce at this level keeps us from having to re-engineer ReactForm to handle validation calls that may never complete while still handling sync calls; or handle async only and rewriting all the validators to support async behavior.

This lets us keep the same api, gives us improved performance across all forms (reduced number of validations in case of other expensive validations like memo), and keeps our code change minimal.

relativityboy: Also, I'm not saying I like the way ReactForm does things. But this came out as the best solution to the situation.

bnchdrff: cool, this makes sense to me. debouncing the setState rather than the validation call feels dirty, but this will get us by for now.

gl2748: The other thing I would say here is that the debounce function could take whatever function we want as an argument...

PR #2358, opened 2018-01-17
Fix account not loading error when tags contain an array of arrays

### LINKED ISSUE #2254: Account disappears after posting from SC2
I used Steem Connect v2 to post from my steemit account @makafuigdzivenu(https://steemd.com/@makafuigdzivenu. The app I used was a https://peerquery.com - a new client I am working on.

After posting a few test posts, my @makafuigdzivenu account has disappeared from https://steemit.com/@makafuigdzivenu and Condenser.

What could be the cause and how do I resolve it?

netuoso:
Closes #2254 

## Issue
https://github.com/steemit/condenser/issues/2254
When a user submits a comment that has an Array of Arrays for the tags field, Condenser breaks due to expecting a single array.

## Solution
When parsing the tags, concat the arrays together to create a single Array from the multiples if they exist.

## Summary
Due to the fact that users are able to submit arbitrary data in the json_metadata field that Condenser parses, users are able to completely break their own account by submitting improperly formatted tags in the json_metadata. One user in particular `@makafuigdzivenu` encountered this issue and is unable to load any of his pages via Condenser. This pull request contains a quick solution to the problem by just concatenating multiple arrays in the tags field to just a single array.

### Improperly formatted tags breaking Condenser
screen shot 2018-01-17 at 1 15 06 pm

### Improperly formatted tags concatenated into a single Array works fine
screen shot 2018-01-17 at 1 42 05 pm

### DIFF
--- a/src/app/utils/StateFunctions.js
+++ b/src/app/utils/StateFunctions.js
@@ -166,6 +166,9 @@ export function contentStats(content) {
         if (typeof tags == 'string') {
             tags = [tags];
         }
+        if (Array.isArray(tags)) {
+            tags = [].concat.apply([], tags);
+        }
         if (!Array.isArray(tags)) {
             tags = [];
         }


gl2748:
Hey @netuoso I like this fix, but in the interest of simplicity, perhaps we can resort to the `catch` in the `try...catch` statement if the tags array is not a very simple flat array.

Regarding the solution here -  does it handle a nested array of depth - n - for example `const tags = ['help', ['me', 'to', ['understand']]]`?

Another thing I would add is that we have recently upgraded our testing suite on condenser to Jest, and a test to accompany this functionality and the change introduced by your code would be much appreciated.
An example of a test have a look here:

https://github.com/steemit/condenser/blob/08f8545e8d8df3e2d6ea73510719b62f95fc9447/src/app/utils/FormatDecimal.test.js

roadscape:
Let's not encourage non-standard data in the tags field. If `tags` is not an array of strings just throw it out. To handle this case (of programmer error?) in some frontends is to increase the complexity for any app which reads this field.

bnchdrff:
Hi (again) @netuoso, we're cleaning up our pull request queue and need to close this PR for now; this isn't to say that we don't want to merge it, though -- rather, feel free to re-open it when you have a chance to make the requested changes. Thanks for your contributions!

Issue #2372, opened 2018-01-22
Not able to show posting key in Spanish interface


bruce-forks:
Copied from [this](https://utopian.io/utopian-io/@suesa/not-able-to-show-posting-key-in-spanish-interface) bug report on Utopian.io:

Expected behavior
Clicking on the "show" button next to the posting key (while logged in) shows posting key, when using Steemit in Spanish.

Actual behavior
Clicking on the "show" button next to the posting key (while logged in) doesn't show the posting key, only when the user chose "Spanish" as language. All other languages function as they're supposed to.

How to reproduce
Switch your language in the settings to "Spanish". Go to the permissions tab and click the button to show the posting key. Nothing will happen.

Browser: Chrome and Firefox
Operating system: ??? (probably Windows, was by a different user) and Windows 10

bnchdrff:
fixed in #2373

PR #2373, opened 2018-01-22
Fixed showing private keys in Spanish

bruce-forks:
**Issue**
[#2372](https://github.com/steemit/condenser/issues/2372)
When a user sets their language to Spanish, they are unable to view their private keys.

**Solution**
In the locale file for Spanish, the value `userkeys_jsx.public_something_key` and `userkeys_jsx.private_something_key` were referencing a value called `count`, which doesn't exist. Changed this to `key` which mirrors all the other locale files.


bnchdrff:
good catch @bruce-forks -- thank you!

PR #2318, opened 2018-01-10
add a server-side-rendered path that does not require the network

gl2748:
closes #2269 

adds an endpoint that can be served without any network requests, for testing or benchmarking.

see README additions for instructions.

PR #2400, opened 2018-01-25
Benchmarking page

### LINKED ISSUE #2269: Create a pure client side page that we can run benchmarks against.
In order to measure performance on Condenser we would like a page that has no external api calls (pure) or dependencies. It should be available as a URL on both development & production to be used by benchmarking & healthchecks. In this way any performance penalties or gains associated with any change to  condenser can be measured in isolation.

gl2748:
closes #2269
Rebase of PR #2318 

adds an endpoint that can be served without any network requests, for testing or benchmarking.
see README additions for instructions.


bnchdrff:
pls clean up squashed commit msg

PR #2284, opened 2018-01-05
Im/2182/add more components to storybook

gl2748:
fixes #2182 #2281

PR #2401, opened 2018-01-25
More Component Stories and init component file-structure.

### LINKED ISSUE #2182: Extend Storybook reach
Expands on the storybook components integrated from issue #2165 / pr #2146

### Progress
- [x] TimeAgoWrapper: 12
- [x] Keys: 4
- [x] Follow: 4
- [x] WalletSubMenu: 3
- [x] TransactionError: 3
- [x] Reputation: 3
- [x] Reblog: 3
- [x] ChangePassword: 3

### LINKED ISSUE #2237: New layout for component files
Our component directories are getting gigantic; what if we switched to a structure where each component has its own subdirectory with an `index.jsx` entrypoint for the component itself, and `style.scss` for its associated styles?

gl2748:
closes #2237 and #2182 

Rebase of pr #2284 

# Before
story_before

# After
story_after


src/app/components/elements/Author/index.jsx
```diff
@@ -5,7 +5,7 @@ import shouldComponentUpdate from 'app/utils/shouldComponentUpdate';
 import Icon from 'app/components/elements/Icon';
 import { Link } from 'react-router';
 import { authorNameAndRep } from 'app/utils/ComponentFormatters';
-import AuthorDropdown from './AuthorDropdown';
+import AuthorDropdown from '../AuthorDropdown';
```
bnchdrff: since AuthorDropdown is only used in this component would it make sense to colocate it in the Author component's subdirectory?

gl2748: I'm not so sure, as a component's usefulness is partially prospective - that it *can* be used elsewhere... and one day might be...

bnchdrff: makes sense yeah; let's just leave it

bnchdrff:
looks like prettier didn't run on a lot of these files; also check the test failures (out-of-date yarn.lock)

webpack/storybook.config.js
```diff
@@ -48,6 +48,11 @@ module.exports = (baseConfig, env) => {
         ]
     };
     config.plugins.push( new ExtractTextPlugin('[name]-[chunkhash].css'));
+    config.plugins.push(new webpack.DefinePlugin({
+        'process.env': {
+            BROWSER: JSON.stringify(true),
+        }
+    }));
```
bnchdrff: eventually we should share webpack config between storybooks and our main system with a [`makeConfig` like they're doing in faucet](https://github.com/steemit/faucet/blob/master/webpack/makeConfig.js)

src/app/redux/TransactionSaga.test.js
```diff
@@ -14,6 +14,11 @@ import {
 } from './TransactionSaga';
 import { DEBT_TICKER } from 'app/client_config';
 
+import { configure, shallow } from 'enzyme';
```
bnchdrff: i'm gonna rip this out real quick, not necessary

src/app/components/elements/WalletSubMenu.story.jsx
```diff
@@ -0,0 +1,18 @@
...
+import rootReducer from 'app/redux/RootReducer';
+import { Provider } from 'react-redux';
+import { createStore } from 'redux';
+const store = createStore(rootReducer);
+
+storiesOf('Elements', module)
+    .addDecorator(withKnobs)
+    .addDecorator(getStory => <Provider store={store}>{getStory()}</Provider>)
+    .add('WalletSubMenu', () => (
+        <Center>
```
bnchdrff: should we add default styles to the storybook to just center everything so we don't have to `<Center>` in each file?

gl2748: Yes, not everything is centered. I would agree that the center wrapper should go into the storybooks config as a decorator. Like this:
https://storybook.js.org/addons/introduction/#storybook-decorators

webpack/storybook.config.js
gl2748: Yes that looks good.

PR #2432, opened 2018-01-30
add feature flags support

### LINKED ISSUE #2179: Add feature flags (ability to turn aspects of condenser on and off)

bnchdrff:
closes #2179

rebase of #2305

rebase of im/2179/feature-flags

see src/app/components/modules/ConnectedFlag/README.md for docs

requests flags from conveyor at user login

### DIFF
--- a/src/app/components/modules/Flag/index.jsx
+++ b/src/app/components/modules/Flag/index.jsx
@@ -1,0 +1,8 @@
+import React, { Component } from 'react';
+
+const Flag = ({ flagged, FlagComponent, Fallback = null, children }) => {
+    if (flagged && children) return { ...children };
+    else return flagged ? FlagComponent : Fallback;
+};
+
+export default Flag;
--- a/src/app/components/modules/Flag/index.jsx
+++ b/src/app/components/modules/Flag/index.jsx
@@ -5,4 +5,24 @@ const Flag = ({ flagged, FlagComponent, Fallback = null, children }) => {
     else return flagged ? FlagComponent : Fallback;
 };
 
+Flag.propTypes = {
+    flagged: React.PropTypes.bool.isRequired,
+    FlagComponent: (props, propName, componentName) => {
+        // First ensure it is a React element
+        React.PropTypes.checkPropTypes(
+            { FlagComponent: React.PropTypes.element },
+            props,
+            'FlagComponent',
+            'Flag'
+        );
+        // Also issue a warning if children are also supplied
+        if (props[propName] && props.children) {
+            return new Error(
+                'Supplied both a FlagComponent and children to Flag; rendering children!'
+            );
+        }
+    },
+    Fallback: React.PropTypes.element,
+};
+
 export default Flag;
--- a/src/app/redux/UserSaga.js
+++ b/src/app/redux/UserSaga.js
@@ -7,6 +7,7 @@ import { PrivateKey, Signature, hash } from '@steemit/steem-js/lib/auth/ecc';
 import { accountAuthLookup } from 'app/redux/AuthSaga';
 import { getAccount } from 'app/redux/SagaShared';
 import * as userActions from 'app/redux/UserReducer';
+import { receiveFeatureFlags } from 'app/redux/AppReducer';
 import { browserHistory } from 'react-router';
 import {
     serverApiLogin,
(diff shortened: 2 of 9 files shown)


bnchdrff:
see https://github.com/steemit/condenser/pull/2395 for the steem-js update PR -- really we should merge that one first!!!

src/app/components/modules/Flag/index.jsx
```diff
@@ -0,0 +1,8 @@
+import React, { Component } from 'react';
+
+const Flag = ({ flagged, FlagComponent, Fallback = null, children }) => {
+    if (flagged && children) return { ...children };
```
relativityboy: If there are children, and there's also a FlagComponent specified it would be good for us to warn.

src/app/redux/UserSaga.js
```diff
@@ -372,13 +373,36 @@ function* usernamePasswordLogin2({
         // Does not need to be fatal
         console.error('Server Login Error', error);
     }
+
+    // Feature flags
+    yield fork(
```
relativityboy: I have't used forks before, nor generators in the way `getFeatureFlags` appears to implement the pattern.

I have a strong appreciation for indicating this code is async, but is there a technical reason for us to have `getFeatureFlags` implemented as a classic function using `await` in place of `yield`?

bnchdrff: `fork` is a redux-saga concept (not the syscall fork)

generators are used in redux-saga to simplify flow control - rather than actually calling the async methods, a saga yields actions (the `put`s and `call`s and `forks`) -- redux-saga will handle pausing the generator function while those methods run async.

src/app/components/modules/Flag/index.jsx
bnchdrff: i'll add to the README that child components will be used before a provided FlagComponent

bnchdrff: done in https://github.com/steemit/condenser/pull/2432/commits/9e4809b2bafad7805c97e665b35951e29136024c

bnchdrff: i'm going to add proptypes validation to the Flag component actually

relativityboy:
@bnchdrff - as I've read the code, if we wanted to have some flags defaulted to `true` we would put them in the initial state at `featureFlags['flagname']`

bnchdrff:
like in a test? if we want a flag to default to true in prod, that'd be handled in conveyor

src/app/components/modules/Flag/index.jsx
bnchdrff: added in https://github.com/steemit/condenser/pull/2432/commits/4c86617bf4f6f76448424b7c67cb3e27d28123dd

src/app/components/modules/Flag/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+        React.PropTypes.checkPropTypes(
+            { FlagComponent: React.PropTypes.element },
+            props,
+            'FlagComponent',
+            'Flag'
+        );
+        // Also issue a warning if children are also supplied
+        if (props[propName] && props.children) {
+            return new Error(
+                'Supplied both a FlagComponent and children to Flag; rendering children!'
```
relativityboy: The value of propName should be included in the error message.

relativityboy:
> like in a test? if we want a flag to default to true in prod, that'd be handled in conveyor
That, or someone else is running condenser and they want to use feature flags in a simple way without relying on their conveyor.

src/app/components/modules/Flag/index.jsx
bnchdrff: it is -- propName is FlagComponent!

bnchdrff:
i don't think we need to account for that use case in this PR. if someone doesn't want to use conveyor then they can load the redux state however they want; the store doesn't care where the state comes from.

PR #2435, opened 2018-01-31
Improve mobile experience by not having first tag capitalized

kirkins:
I went to create a post on the mobile website.

My first tag was auto-capitalized, I then went to post, and received feedback that tags cannot be capitalized.

screenshot_20180130-205337

This should fix that.

### DIFF
--- a/src/app/components/cards/CategorySelector.jsx
+++ b/src/app/components/cards/CategorySelector.jsx
@@ -72,6 +72,7 @@ class CategorySelector extends React.Component {
                 ref="categoryRef"
                 tabIndex={tabIndex}
                 disabled={disabled}
+                autocapitalize="none"
             />
         );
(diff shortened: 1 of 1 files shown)


pkattera:
@kirkins Thanks for fixing this!

sneak:
I'd love to see us migrate to a component like the following:

https://github.com/NdYAG/react-tageditor

demo:  https://ndyag.github.io/react-tageditor/

It disambiguates what is and isn't a tag boundary by creating a new subcomponent when you hit tab.  I think we should expand this to create a new subcomponent when you hit tab or space or comma, so any of those common tag delimiters clearly indicate to the user what set of tags they have chosen.

sneak:
Related: replace tag picker component entirely: #2437

Issue #2429, opened 2018-01-30
Showing real rewards value


bitcoiners:
Now that SBD USD price parity is broken, steemit.com is not showing correct value of rewards. This should be fixed.

roadscape:
Most posts pay out 50% in SP and 50% in SBD; for example, a post with a "Payout: $100" shown receives ~$50 of SP plus 50 SBD. SBD is worth $6 now, so the value of the payout is closer to $350. 

At SBD = $6, our displayed payouts are ~71% lower than they should be.

options:
 - fix SBD peg (blockchain issue)
 - account for the discrepancy in UI:
    - remove the dollar sign from payout and treat it as points, or
    - account for SBD price in payout display (be sure to use `post.percent_steem_dollars`)
      - current SBD price can be calculated as midpoint of spread on internal market

eonwarped:
I propose that when clicking the payout and the little card expands, it simply shows you how much projected SBD and STEEM you expect to receive. Hell, we have enough to predict all of those with the current parameters. Even curation, if we wanted to go that far, but for now I'd say it'll be simpler just to do SBD/STEEM POWER break down.

Smittyvb:
@themarkymark-steem This could cause issues with users comparing post values. Many currencies are similar — Canadian Dollars, American Dollars, and Australian Dollars all have the same symbol, but their worth can vary a lot — 10 USD is 13 AUD. This will be confusing to people talking about their post value, for example, in other posts. Also, people might ignore the currency sign, and mistake yen, for dollars, for example. (¥108 = $1, so someone might think that they’ve got $108, instead of $1).

PR #2460, opened 2018-02-04
Add price feed info and sbd/steem breakdown

eonwarped:
This is to address issue #2429 .

Preliminary quick hack for adding SBD and STEEM POWER break down, affecting the expanded view for payouts, the "Pending Payout" and "Author Payout", using the steem price feed. Also takes into account the % SBD setting. But does not take into account beneficiaries or curation.

- Pending Payout shows SBD and SP breakdown.
- For Author Payout, it is not accurate, as it is should be based on a historical price feed (SBD's portion is accurate, but not SP).

I'm not even sure this is desirable, so I leave it to people to chat about it.

(If we do want this: Any suggested locations for adding testing, if necessary?)

Here are the screenshots

For pending payout:
cr_normal_pending

For 100% Power up:
cr_powerup100

After payout:
cr_authorrew


bnchdrff:
@pkattera feedback: let's push it through, maybe apply lighter text style to new info

netuoso:
The payout values on this (for the SP portion) are going to be wildly incorrect as the price of STEEM changes. The payouts are based on a price average over the course of the payout period for the post. This PR will only have *semi-accurate* payout breakdowns on very recently paid out posts when the price hasn't moved much.

PR #2462, opened 2018-02-05
Updated DMCAList.js

Gandalf-the-Grey:
- two urls added
- removed (unneeded) tags from links


sneak:
@Gandalf-the-Grey Has this been tested with and without the category tag?  Does it still block them if the URL used to access them still includes the category tag?

Gandalf-the-Grey:
@sneak yes, it's redirected to url with category tag and blocked

Issue #2436, opened 2018-01-31
prereq install fails on latest stable node version


sneak:
I'm a developer/entrepreneur.  I use steemit all the time and want to add a feature to condenser.  I make a blog post and it gets widespread support.  I make an issue in the issue tracker proposing it and TPTB say it'd get merged if I wrote it.

I sit down to my brand new freshly installed Macintosh with High Sierra and install Xcode and homebrew.  I `brew install node` and `npm install -g yarn` to bring up a dev environment.  I clone condenser and check out `master` with the plan to run the tests to verify my dev environment baseline is set up correctly.  I get an error because it can't install on the latest stable version of node that homebrew installs.

```
✔ ~/tmp/condenser [master|…1] 
23:15 $ yarn install
yarn install v1.3.2
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
error @steemit/rpc-auth@1.1.0: The engine "node" is incompatible with this module. Expected version "^8".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
✘-1 ~/tmp/condenser [master|…1] 
23:18 $ node -v
v9.4.0
✔ ~/tmp/condenser [master|…1] 
23:18 $ 
```

I know node 8 is LTS, but that doesn't actually matter at all if you read that user story.

sneak:
I believe this just requires an rpc-auth bamp because https://github.com/steemit/rpc-auth/pull/16 is already done and merged.

Issue #2365, opened 2018-01-19
Vulnerabilities in Dependencies


jayserdny:
Steemit condenser has several Dependencies that are exposed to vulnerabilities. Some of these can be remedied by updating to the latest version.

##### Vulnerable Modules
***
- shelljs **introduced through:** git-rev-sync@1.9.1
- koa-static-cache **introduced through:** koa-static-cache@3.2.0
- qs **introduced through**: co-body@4.2.0 and purest@2.0.1
- minimatch **introduced through:** sequelize-cli@2.8.0
- jquery **introduced through: **medium-editor-insert-plugin@2.4.1
- request **introduced through:** purest@2.0.1
- http-signature **introduced through:** purest@2.0.1
- tunnel-agent **Introduced through:** node-sass@4.7.2, twilio@2.11.1
- mime **introduced through: **url-loader@0.5.9
- ms **Introduced through:** twilio@2.11.1

##### Detailed Vulnearibilities
***
###### Command Injection: git-rev-sync@1.9.1
"shelljs is a portable Unix shell commands for Node.js. It is possible to invoke commands from shell.exec() from external sources, allowing an attacker to inject arbitrary commands.""

**Remediation: Not fix available**

Source: [https://snyk.io/vuln/npm:shelljs:20140723](https://snyk.io/vuln/npm:shelljs:20140723)
###### Directory Traversal: koa-static-cache@3.2.0
"Affected versions of the package are vulnerable to Directory Traversal. When in dynamic mode, a malicious user can traverse through the servers files, by entering %2E%2E/ into the url, allowing the attacker to obtain the contents of any file on the server's filesystem."

**Remediation: Upgrade to koa-static-cache@4.1.1.**

Source: [https://snyk.io/vuln/npm:koa-static-cache:20170613](https://snyk.io/vuln/npm:koa-static-cache:20170613)

###### Prototype Override Protection Bypass:  co-body@4.2.0 and purest@2.0.1
"In versions of the package affected by this vulnerability, it is possible to circumvent this protection and overwrite prototype properties and functions by prefixing the name of the parameter with [ or ]. e.g. qs.parse("]=toString") will return {toString = true}, as a result, calling toString() on the object will throw an exception."

**Remediation: Upgrade to co-body@5.0.0 & Upgrade to purest@3.0.0.**

Source: [https://snyk.io/vuln/npm:qs:20170213](https://snyk.io/vuln/npm:qs:20170213)

[...]

PR #2476, opened 2018-02-06
add nsp, update packages with vulnerabilities

### LINKED ISSUE #1637: Node Security Platform should be integrated and risk should be mitigated
NSP should be run every build and the build should fail if NSP finds vulnerabilities.

bnchdrff:
fixes #1637, #2365, and #2436 

see notes starting at https://github.com/steemit/condenser/issues/2365#issuecomment-363161969


bnchdrff:
for the record it seems like nsp's vulnerability db isn't as comprehensive as snyk

Issue #2459, opened 2018-02-04
condenser’s layouts do not reserve space for images before images load


sneak:
This causes redrawing/reshuffling of the list when the images finally load.  Condenser should know in advance how big the thumbnails will be and render placeholders until the images load from the imagehoster.

pkattera:
A dev team that I worked with a couple of years ago used this technique to stop pages from jumping around while images load: 

http://aspiringwebdev.com/stop-your-web-pages-from-jumping-around-while-images-load/

gl2748:
First phase do this on the PostsIndex list.

gl2748:
# Before:
![before_slow](https://user-images.githubusercontent.com/3374538/35921964-227c6e52-0c14-11e8-839c-6e714a8cbcff.gif)

jnordberg:
https://github.com/steemit/imagehoster/issues/60
https://github.com/steemit/hivemind/issues/81

PR #2477, opened 2018-02-06
logo-type in header, move sort order dropdown, refs #2391

### LINKED ISSUE #2391: Include the full logotype on mobile
**The problem**
Currently, we do not show the full logotype within the mobile primary navigation. This means we're missing an important branding opportunity when new users come from a google search or social link.

screen shot 2018-01-23 at 11 16 03 am

**The solution**
Display the full steemit logotype for all screen sizes both for authenticated and non-authenticated users.

screen shot 2018-01-23 at 11 19 31 am

_Home page_

To implement this, we'll need to remove the content filter within the primary navigation on the home page. Instead, we'll add a dropdown to the page title. 

screen shot 2018-01-23 at 11 19 20 am

----

Other elements and icons also need to be updated but I'll create another ticket for that.

gl2748:
closes #2391 

# Desktop
demo

# Mobile
mobile_not_logged_in

# Known feedback.
- [ ] Make all header icons uniform size and make sure they align in a pleasing fashion.
- [ ] Get feed back on the topic selector / and topic title both being displayed. Perhaps the title is redundant?
- [ ] Make sure elements align as desired.
- [ ] When user is on their feed - the 'Home' Sort Order, the 'Home' List element does not receive  the 'active' class. It should.


gl2748:
I think these gifs also demonstrate an issue that can be addressed in #2459 - you can see that the column that holds the list of posts collapses when its content is loading. Ideally this would not happen and instead a 'Flash-of-unstyled' content would be displayed in such a way that this isn't jarring.

gl2748:
screenshot from 2018-02-07 16-41-13

screenshot from 2018-02-07 16-41-07

gl2748:
Superceded by PR:
#2591

PR #2482, opened 2018-02-07
add placeholder for loading images in blogmode, refs #2459

### LINKED ISSUE #2459: condenser’s layouts do not reserve space for images before images load

gl2748:
Fixes #2459 

# Before:
before_slow

# After:
place_holder

### DIFF
--- a/src/app/assets/stylesheets/foundation-overrides.scss
+++ b/src/app/assets/stylesheets/foundation-overrides.scss
@@ -109,7 +109,7 @@ tbody tr:nth-child(even) {
 }
 
 .reveal {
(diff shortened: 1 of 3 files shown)


src/app/assets/stylesheets/foundation-overrides.scss
```diff
@@ -119,3 +119,21 @@ tbody tr:nth-child(even) {
     min-width: 100px;
   }
 }
+
```
bnchdrff: mind throwing these in PostSummary.scss?

gl2748: Yes makes more sense.

PR #2485, opened 2018-02-07
add clickwall & style to bad actor memos

### LINKED ISSUE #2206: Make transfers from the "bad actors" list appear differently in the UI
Motivation: bad actors can perform phishing attempts by sending transfer messages that appear to come from legit steem service partners such as exchanges. 

Maybe grey these transfer memos with a warning when you hover over them.

bnchdrff:
closes #2206

also, clean up the memo component a bit


bnchdrff:
redesigned based on feedback starting at https://github.com/steemit/condenser/issues/2206#issuecomment-363909518

screenshot from 2018-02-13 10-24-45

src/app/components/cards/TransferHistoryRow.scss
```diff
@@ -0,0 +1,18 @@
+.Memo--badActor {
+  .bad-actor-caution {
+    color: $color-red;
+    font-size: 75%;
+    font-weight: bold;
+    text-transform: uppercase;
```
pkattera: Can we please include some letter-spacing to the 'CAUTION'
e.g. `letter-spacing: 2px`

bnchdrff:
added some spacing -- 1px seems quite nice:

screenshot from 2018-02-13 13-47-51

bnchdrff:
@sneak mind reviewing again?

sneak:
Looks good.  My nonblocking feedback is that I would desaturate the red a bit (but keep it red) because the CAUTION is very eye-catching - we are trying to encourage them to ignore the memo/not click.

sneak:
PS by desaturate I mean reduce brightness, not pink-ify.  pre-coffee this morning...

bnchdrff:
made it darker:

screenshot from 2018-02-20 14-49-02
screenshot from 2018-02-20 14-48-54

pkattera:
Note: We have to be especially careful about changing text colors since we have two different themes. Any new text color either has to be accessibility contrast checked to be legible on both the light and dark background, or we need to theme it with slightly different shades via theme.scss.

bnchdrff:
the darken()ed red is probably inaccessible, very true -- also yeah this should be a themed color

sneak:
I love where this ended up; the results look great. I’m not too worried about the accessibility due to the fact that “hide memos from bad actors, period” would also have been an acceptable solution.

sneak:
Also, as long as the text is legible, the CAUTION tag is functionally optional.

Issue #2420, opened 2018-01-29
Going to other user's feed, page title says "My Feed" should say "${account}'s feed"


relativityboy:
Title says it all, but here's a screenshot of user 'valuproposition' browsing to user 'lovejoy' 's feed page.
<img width="682" alt="screen shot 2018-01-28 at 9 47 20 pm" src="https://user-images.githubusercontent.com/599528/35492897-1a25f3dc-0475-11e8-80ec-8bfd25529801.png">

roadscape:
Browsing others' feeds is an undocumented feature; a PR would be welcome for such a tweak but closing the issue for now.

PR #2498, opened 2018-02-10
Viewing another user's feed shows their username in title

bruce-forks:
**Issue**
[#2420](https://github.com/steemit/condenser/issues/2420)
When you visit the feed of another user, it is still labeled as "My Feed" even though it belongs to another user.

**Solution**
In the PostsIndex page, I added a variable that contains the username of the user's feed that we are viewing. If that string is equal to the username of the user who is logged in, the title says "My Feed". Otherwise, the title says "___'s Feed", where "___" is the username.

### DIFF
--- a/src/app/ResolveRoute.js
+++ b/src/app/ResolveRoute.js
@@ -55,9 +55,6 @@ export default function resolveRoute(path) {
     if (path === '/approval') {
         return { page: 'Approval' };
     }
-    if (path === '/pick_account') {
-        return { page: 'PickAccount' };
-    }
     if (path === '/recover_account_step_1') {
         return { page: 'RecoverAccountStep1' };
     }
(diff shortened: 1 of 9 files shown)


gl2748:
@bruce-forks congrats on your first contribution to Steemit and your first PR! We're gonna get this into the codebase!

One thing I would request is that the following user story also be satisfied by the work in this PR:

>  As an anonymous (not-logged-in) user, when I visit another user's feed, the page_title is:
`account_name + "'s Feed"`. 

Basically, extend your work beyond the condition [`if this.props.username !== 'undefined'` ](
https://github.com/steemit/condenser/pull/2498/files#diff-499e339aab8691ef8271f774c3ad9553R159)

Screenshot of current `page_title` for a user's feed as seen by an anonymous user:
screenshot from 2018-02-15 12-37-40

PR #2510, opened 2018-02-13
remove rate limit from create_user endpoint

### LINKED ISSUE #2509: remove condenser-side rate limiter from create_user endpoint
faucet is hitting this limit when new users are added.

bnchdrff:
closes #2509

nginx will still apply a higher rate limit so we are still protected.

also a request will get denied quickly if it doesn't supply the correct secret.

i moved the secret check above the logger line so that we can't get logspammed.


gl2748:
LGTM

Issue #2529, opened 2018-02-16
translatable strings should be able to accept react dom elements as interpolated variables


bnchdrff:
https://github.com/steemit/condenser/blob/19536bc638d47ca069306915086659e6fb1b4e7e/src/app/components/cards/TransferHistoryRow/index.jsx#L178-L187

this type of language string concatenation makes translation hard or impossible. while we can interpolate plain text with our current translation framework (react-intl and counterpart) we cannot interpolate react dom elements (links or styled text, for example).

we can look at https://github.com/Gandi/react-translate as well as https://github.com/martinandert/react-translate-component and https://github.com/martinandert/react-interpolate-component for hints

see #1929, #2349, #2351

Issue #2351, opened 2018-01-16
Adjectives are inappropriately reused in translation


goldibex:
From #1929.

A little more explanation:

In some cases we're inappropriately reusing translations for words like "New" where semantically a translation can't follow the English structure. For instance, you can't reuse the translation for "New" between "New Post" and "15 New Messages" for 2 reasons: 1) "new post" is actually an abbreviated command, and 2) in other languages, the form of the word "new" may change depending on whether it's modifying the word "post" or "messages".

Generally speaking, our policy should be never to reuse a translation except semantically. For instance, a whole phrase like "You cannot undo this action" could be reused.

---------------------------------

For example:
![](https://camo.githubusercontent.com/dca35cff899fce8297fc0b2df07534033d487a45/68747470733a2f2f7332302e706f7374696d672e6f72672f67627763366d7063742f696d61676531312e706e67)
In the top page, this "new" means "new posts".

I want to translate it into "新着" in Japanese.

In Change Password page, this word is also used.

It means "new password", so I want to translate it into "新しい" in Japanese, or to leave it in English.
"新着" is unsuitable.

However, in the souce code, these words are defined by same term 'g.new'.
https://github.com/steemit/condenser/blob/master/src/app/components/modules/Header.jsx
![](https://camo.githubusercontent.com/1edcc1844a40bff20951fbcae115162ea4fae571/68747470733a2f2f7332302e706f7374696d672e6f72672f6867716563643673642f696d61676531332e706e67)
https://github.com/steemit/condenser/blob/master/src/app/components/elements/ChangePassword.jsx

Unless they are used in exactly the same meaning, they should be defined separately.

bnchdrff:
find translated strings that are reused, and for each of them, if they are semantically different, create unique translated strings

sneak:
(even if the english version is identical)

bnchdrff:
this needs more work; i forgot to put in the locale strings for the first component i did (TransferHistoryRow) and it'll take a minute; closing this until i fix that.

bnchdrff:
darn it i meant to close my PR. this issue is still open.

PR #2531, opened 2018-02-16
de-frag translations

### LINKED ISSUE #2349: Fix hard-coded word order in translations.
From #1929. The outcome is to review the referenced code, see if the problem still exists, and if so resolve it.

------------------------------

In some languages, their grammar is vastly different from English, so if word order is fixed, translators cannot make natural translation.

For example:
screen

The order of these words is completely fixed in source code.
https://github.com/steemit/condenser/blob/master/src/app/components/cards/TransferHistoryRow.jsx

I want to translate this message "Transfer 0.200 SBD to inoue" into "inoueに0.200SBDを送りました" in Japanese.
However, because the word order is fixed, I translated it as "送信 0.200 SBD 宛先:inoue" in Japanese.
This is very unnatural.

Another one:

In left side of top page, if the user clicks "Show more", it will change to "Show fewer".

https://github.com/steemit/condenser/blob/master/src/app/components/pages/PostsIndex.jsx
line 177: <small><a className="c-sidebar__more-link" onClick={this.onShowSpam}>{showSpam ? tt('g.next_3_strings_together.show_less') : tt('g.next_3_strings_together.show_more')}</a>{' ' + tt('g.next_3_strings_together.value_posts')}</small>
The order of "Show more" (or "Show fewer") and "low value posts" can not be swapped.

I couldn't translate these into Japanese.
Because Japanese word order is SOV (subject-object-verb).
So I had to write "low value posts" blank and I wrote full text on "Show more" or "Show fewer".

bnchdrff:
closes #2349

fixes some of #2351 but that will need more work

still need to figure out a way to interpolate react dom elements into a translation (#2529)

also, this commit adds a storybook to cover all cases for TransferHistoryRow.

this story provides knobs for any incoming prop that affects branching and can be used as the basis for a real test eventually


src/app/components/cards/TransferHistoryRow/index.jsx
```diff
@@ -0,0 +1,345 @@
...
+                type === 'transfer_to_savings'
+                    ? 'to_savings'
+                    : type === 'transfer_from_savings'
+                      ? 'from_savings'
+                      : 'not_savings';
+
+            if (data.from === context) {
+                // Semi-bad behavior - passing `type` to translation engine -- @todo better somehow?
+                // type can be to_savings, from_savings, or not_savings
+                // Also we can't pass React elements (link to other account) so its order is fixed :()
```
gl2748: Bummer. For future reference what library supports passing components?

src/app/components/cards/TransferHistoryRow/story.jsx
```diff
@@ -0,0 +1,136 @@
...
+import { receiveState } from 'app/redux/GlobalReducer';
+import mockState from '../../../../../api_mockdata/get_state';
+
+import TransferHistoryRow from './';
+
+const store = createStore(rootReducer);
+
+// Load some fake blockchain state.
+// This component relies on many parts of the state tree - messy & hard to test!
+store.dispatch(receiveState(mockState));
```
gl2748: This is a good point of reference for working on other stories that will need store data.

gl2748:
Hey @bnchdrff is there an order of work for updating the locales other than (en) to use the tokens now being passed?

bnchdrff:
> Hey @bnchdrff is there an order of work for updating the locales other than (en) to use the tokens now being passed?

that's an @originated question!

src/app/components/cards/TransferHistoryRow/index.jsx
bnchdrff: https://github.com/martinandert/react-interpolate-component along with https://github.com/martinandert/react-translate-component is one option that is sorta workable with our existing system

https://react.i18next.com is another one

my fave so far is https://github.com/lingui/js-lingui

gl2748:
@bnchdrff  - Getting missing translations warnings on the new storybook story for Transfer History Row.
Is this to be expected? Locale is set to 'EN' via the storybooks-intl knob-a-roo.
screenshot from 2018-02-20 09-11-51

bnchdrff:
those errors are to be expected if i did a bad job filling in all of the new translations ;) fixing it now

bnchdrff:
this needs more work; i forgot to put in the locale strings for the first component i did (TransferHistoryRow) and it'll take a minute; closing this until i fix that.

bnchdrff:
@originated what should we do here? a bunch of new translations will be needed for this PR

originated:
@bnchdrff Updates look good. To answer above currently flow is to make sure the en.json is always up to date in Transifix and have them translated first in the system. 

As far as getting the keys done promptly it's a product question for @pon if we should indeed be using a service, etc.

Issue #2486, opened 2018-02-08
FAQ Updates


plink01001:
Several updates are needed to the FAQ page. This issue (and the associated PR) will address the following:
- Improve the language for the opening questions at the top of the FAQ
- Review all FAQ material to ensure it is still accurate and up to date
- Review all external links to ensure they still work
- Add information on SMTs to FAQ page #2096
- Add “Can I change my username?” #2097
- Add new question to FAQ for which exchanges STEEM and SBD are listed on #2252
- Add information on bandwidth to FAQ #2298
- Update information on password security

plink01001:
Also include a fix for https://github.com/steemit/condenser/issues/2419

plink01001:
Also add: _Can I speed up the account creation process?_

plink01001:
PR has been merged. @jnordberg this can be removed from the backlog.

PR #2533, opened 2018-02-17
Update faq.md

parmsang:
Not the best first impression when you're greeted with a typo.


plink01001:
@parmsang thank you very much for your contribution! I am working on a larger change to the FAQ page here: https://github.com/steemit/condenser/issues/2486 and will include a fix for this as part of that change.

PR #2538, opened 2018-02-18
ignore case in phishy link checker

### LINKED ISSUE #2536: phishy links detector should be case-insensitive
as a malicious user, when i try to fool users by linking to steamit.com with text that contains Steemit.com, my link should be caught by HtmlReady's phishy link detector.

bnchdrff:
closes #2536


src/shared/HtmlReady.js
```diff
@@ -146,8 +146,8 @@ function link(state, child) {
 
             // Unlink potential phishing attempts
             if (
-                child.textContent.match(/(www\.)?steemit\.com/) &&
-                !url.match(/https?:\/\/(.*@)?(www\.)?steemit\.com/)
+                child.textContent.match(/(www\.)?steemit\.com/i) &&
```
gl2748: Outside the ambit of this work but I would suggest a more aggressive phish detector to cover someting like:
**Steemit is so cool**: [www.steemit.cool](http://www.lotsofmany.com/)
**New british steemit**: [www.steemit.co.uk](http://www.eggalone.com/)
Suggest:
```
child.textContent.match(/(steemit)+\.([\w\d\.])*/i)
```

bnchdrff: too many false positives w/ that approach

sneak: they are using `steemiit.com` IRL: https://steemit.com/pretty/@siennawu/snow-day#@soniia/re-siennawu-snow-day-20180218t140601956z

PR #2570, opened 2018-02-27
add hide url

ludorumjeoun:
the author asked for hide this URL.
/blog/@dreamya/korea-bucheon-cartoon-festival-cosplay-card-capture-sakura

author's [request](https://steemit.com/kr/@dreamya/kuzhd) google translate.

After posting to Steamit in April of last year, it is when I did not know that it can not be deleted or modified after payout. I went to the Bucheon Manga Festival event and took a picture of a language and post it. Of course, I was allowed to take pictures, and I was allowed to post them on my blog. After posting, the contents posted via email were sent by e-mail with the link. I received an e-mail saying, "I want to delete the image in the article after a few months or so in September, and it was impossible to delete or edit it already.

I tried to delete it later in various ways, but it has not been deleted yet and I would like to ask for help. Thank you for downloading the link below.

Currently, we do not take portraits and we are doing our best. I hope that this will not happen in the future.


sneak:
That file is only to be modified upon Steemit receiving a properly formatted DMCA request.

PR #2579, opened 2018-02-28
Fix Bug Where Comment sorting changes don't apply until refresh

### LINKED ISSUE #2264: Comment sorting changes don't apply until refresh
As a user, I want to be able to sort comments for a post by: 
-  'trending'
-  'votes'
- 'age'.

To do this I navigate to the comment thread on a `Post` and then in the upper right hand corner select the sort order I desire. After I have selected a sort order the order of the displayed comments should update accordingly.

Right now this does not happen. Upon selecting a sort order, the argument is simply added as a URL param and the user must manually refresh the page for the comment order to update.

Likely a bug in the state management.

screenshot from 2018-01-02 15-04-17

caleblogan:
fixes #2264 

The problem here is App doesn't update when `props.location.query` changes. This leaves Post with an out of date location prop. I added another check to `App.shouldComponentUpdate` which solves the problem. However this seems to add lag when sorting since the App component has to update.

## Other Possible Fixes would be
- Grab the location from `window`.
- Subscribe to react-router outside of the parent component, if that's possible.
- Add local state to keep track of the selected sort value.

### DIFF
--- a/src/app/components/App.jsx
+++ b/src/app/components/App.jsx
@@ -103,6 +103,7 @@ class App extends React.Component {
         const n = nextProps;
         return (
             p.location.pathname !== n.location.pathname ||
+            p.location.query !== n.location.query ||
             p.new_visitor !== n.new_visitor ||
             p.flash !== n.flash ||
             this.state.open !== nextState.open ||


roadscape:
Thanks for your contribution @caleblogan, I've updated your branch in #2629 and testing shortly.

Issue #2542, opened 2018-02-19
Minor UI updates to Settings


pkattera:
Some UI and structural cleanup for the Settings page, enabling us to add future user preferences, like rewards preferences.

![slice 2](https://user-images.githubusercontent.com/1121139/36365899-5c6c7f40-1500-11e8-89f6-e191e7cabb18.png)

gl2748:
Confine to this page.

originated:
Please see PR https://github.com/steemit/condenser/pull/2589

PR #2589, opened 2018-03-02
Update settings page with new styling, formatting

originated:
Handles #2542


jnordberg:
screen shot 2018-03-02 at 12 47 24

PR #2591, opened 2018-03-02
logo-type in header, move sort order dropdown, refs #2391

### LINKED ISSUE #2391: Include the full logotype on mobile

gl2748:
closes #2391 
New header.


bnchdrff:
search & post buttons should have the same outline style; currently the search button has an inner shadow

screenshot from 2018-03-02 15-00-54

bnchdrff:
and then if you're logged in w/ the default avi, you get another slightly different circle button:

screenshot from 2018-03-02 15-03-12

bnchdrff:
in Firefox, the header overflows <640px:

screen shot 2018-03-02 at 15 06 03
screen shot 2018-03-02 at 15 05 54

bnchdrff:
i think it's funny looking how the blog/list mode toggle is slightly shorter than the two select dropdowns (tags/trending) -- it kinda makes it look like two squinting eyes with a small crinkled nose in the middle

bnchdrff:
on iphone se, buttons overlap logo:

screenshot from 2018-03-06 11-39-40

src/app/components/elements/SearchInput/SearchInput.story.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs, boolean } from '@storybook/addon-knobs';
+import SearchInput from './index';
+import { Center } from '../Tooltip.story';
+
+storiesOf('Elements', module)
+    .addDecorator(withKnobs)
```
bnchdrff: vestigal knobs

src/app/components/elements/SearchInput/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+const SearchInput = ({ type }) => {
+    return (
+        <span>
+            <form
+                className={'search-input'}
+                action="/static/search.html"
+                method="GET"
+            >
+                <svg
```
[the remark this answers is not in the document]
gl2748: I don't really like what the icons component is doing. It's not very clear, and I have found it hard to work with.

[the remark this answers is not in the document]
bnchdrff: cool, yeah it's a mess i agree

src/app/utils/getPageTitle.js
```diff
@@ -0,0 +1,79 @@
+import tt from 'counterpart';
+import normalizeProfile from 'app/utils/NormalizeProfile';
+
+const getPageTitle = (route, account_meta) => {
+    let page_title = route.page;
+    let sort_order = '';
+    let topic = '';
+    let user_name = null;
+    if (route.page === 'PostsIndex') {
```
bnchdrff: a switch statement would be a little easier to read & maintain

src/app/redux/UserReducer.js
```diff
@@ -394,3 +404,13 @@ export const uploadImage = payload => ({
     type: UPLOAD_IMAGE,
     payload,
 });
+
+export const showSidePanel = () => ({
+    type: SHOW_SIDE_PANEL,
+});
+
+export const hideSidePanel = () => {
```
bnchdrff: use shortand syntax like you do in showSidePanel above

PR #2595, opened 2018-03-02
2430 add ability to upload profile and cover photo images

### LINKED ISSUE #2430: Community - Add ability to upload profile and cover photo images
Currently the settings page only accepts URLs for cover photo and profile images. This causes users to use third party websites to upload their images, instead of uploading them directly to the site. It would be a much better user experience if users could upload the images directly, similar to how it is done for posts and comments.

voorash:
Worked on this at the request of witness followbtcnews and suggestion of @TimCliff 

The change from steemit chat to steem chat should not have been in this branch but assuming that by the time this pull request is merged my other one that had that change in it will already be merged so it shouldn't produce any conflict.

closes #2430


voorash:
snap

roadscape:
Thank you for this contribution! I've moved the PR to a local branch for review: https://github.com/steemit/condenser/pull/2652

PR #2628, opened 2018-03-07
add blackbox test suite

### LINKED ISSUE #2507: research a blackbox / selenium test suite
create a quick set of selenium scripts to test loading the homepage, finding certain important content items, ensuring no js errors happen, logging in on testnet, posting, upvoting, logging out.

bnchdrff:
closes #2507 

- migrate to circleci v2
- add a docker image definition for a nightwatch test suite (uses selenium & chromedriver) with a simple home page load test as an example
- run test suite in cirleci alongside prod condenser image


blackboxtest/nightwatch-deps.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
jnordberg: What is this file for?

blackboxtest/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM circleci/openjdk:jdk-node-browsers
+
+WORKDIR /home/circleci
+COPY package.json yarn.lock nightwatch-deps.sh /home/circleci/
+RUN yarn install --non-interactive --frozen-lockfile
+RUN curl -s https://selenium-release.storage.googleapis.com/3.9/selenium-server-standalone-3.9.1.jar > /home/circleci/selenium-server-standalone.jar
+RUN curl -s https://chromedriver.storage.googleapis.com/2.36/chromedriver_linux64.zip | funzip > /home/circleci/chromedriver && chmod +x /home/circleci/chromedriver
```
jnordberg: Isn't there a selenium image we can use directly?

blackboxtest/nightwatch.json
```diff
@@ -0,0 +1,44 @@
+{
+  "src_folders" : ["/home/circleci/nightwatch-tests"],
```
jnordberg: Use relative paths if possible

circle.yml
```diff
@@ -1,11 +1,15 @@
...
-        - echo "Ignore CircleCI detected dependencies"
-
-test:
-    override:
-        - docker build -t steemit/steemit.com .
+version: 2
+jobs:
+    build:
+        docker:
+            - image: circleci/openjdk:jdk-node-browsers
```
jnordberg: Only docker is needed here, I suggest using `docker:17`

circle.yml
```diff
@@ -1,11 +1,15 @@
...
+        steps:
+            - checkout
+            - setup_remote_docker:
+                docker_layer_caching: true
+            - run: docker build -t steemit/condenser:$CIRCLE_BRANCH .
+            - run: docker build -t steemit/condenser-blackboxtest:$CIRCLE_BRANCH blackboxtest/
+            - run: |
+                docker run -ditp 8080:8080 --name condenser steemit/condenser:$CIRCLE_BRANCH
+                sleep 10 # let condenser start
```
jnordberg: Did you try the circleci 2.0 workflows with dependencies? Seems like we should be able to run the builds in parallel using that

https://circleci.com/docs/2.0/workflows/

blackboxtest/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "condenser-blackboxtest",
+  "version": "1.0.0",
+  "main": "false",
```
jnordberg: Why is main set to false and why is it a string?

blackboxtest/nightwatch-tests/homepage.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
jnordberg: Idea for test: Load a specific permlink and verify that the post content and a couple of comments renders correctly

circle.yml
jnordberg: It would also allow us to specify in GitHub that only the main docker build step is a requirement for merging so that a blackboxtest failure won't block

blackboxtest/nightwatch-deps.sh
bnchdrff: totally vestigal; will remove

blackboxtest/Dockerfile
bnchdrff: there aren't any circleci-provided ones but yes there probably is one out there; i found this way to work pretty well & it gets cached anyways. now we have our own selenium image ;)

blackboxtest/nightwatch.json
bnchdrff: good point.

circle.yml
bnchdrff: sounds cool, will try it out!

blackboxtest/package.json
bnchdrff: good point

circle.yml
bnchdrff: 10 4; done

blackboxtest/Dockerfile
jnordberg: Looks like there is an official selenium image, if that's a drop in replacement I suggest using that just for the free updates :)

https://hub.docker.com/r/selenium/standalone-chrome

circle.yml
bnchdrff: so i set up a basic workflow; the fancy thing to do would be to refactor our existing Dockerfile into steps for `yarn install`, `unit & integration tests`, and `prod build` -- the latter two can run in parallel, and then blackboxtest would be dependent on prod build.

jnordberg:
@jredbeard before this can be merged we need to change the GitHub required check from `ci/circleci` to `ci/circleci: build`

blackboxtest/nightwatch-tests/homepage.js
bnchdrff: put test ideas here https://github.com/steemit/condenser/issues/2637

blackboxtest/nightwatch-tests/ssr.js
```diff
@@ -0,0 +1,44 @@
...
+						'/change_password',
+						'/create_account',
+						'/approval',
+						'/recover_account_step_1',
+						'/recover_account_step_2',
+						'/waiting_list.html',
+						'/market',
+						'/~witnesses',
+						'/submit.html',
+        ].forEach(path => testHttpGet('localhost', 8080, path, browser));
```
jnordberg: Does this not need some callback? Or some eqvl of `test.plan(numUrls)`

(ps indents are wonky)

bnchdrff: nope, this works! fixing indents now.

jnordberg: What happens if the get request to one of the urls never returns?

bnchdrff: our test suite would hang!

we could do a timeout handler for `http` -- actually that is a good idea; i'll set it up to a short-ish limit since a slow rendering page is a bad sign anyways

bnchdrff: done; applied a timeout of 5000ms

PR #2623, opened 2018-03-07
Suggestion - Add Feature to Sort By Author Reputation

caleblogan:
Not sure if you want this. I thought I would just throw it up here and see what you thought.
Suggestion post: https://busy.org/@therealwolf/feature-request-sorting-comments-by-author-reputation

PR #2634, opened 2018-03-08
sort comments by author reputation

roadscape:
Replaces #2623 by @caleblogan:

> Not sure if you want this. I thought I would just throw it up here and see what you thought.
> Suggestion post: https://busy.org/@therealwolf/feature-request-sorting-comments-by-author-reputation


roadscape:
Only issue I see is inconsistent capitalization:

image

Otherwise this feature works great.

bnchdrff:
that's a pre-existing bug though; let's throw in a `text-transform` to fix it in this PR though

bnchdrff:
(i'm doing it now)

roadscape:
👌it looks like the FoundationDropdownMenu itself will need a new class applied to it, right now it's all generic

bnchdrff:
created a new set of translatable strings for the sort order options, which lets us capitalize them without weird styling, and also makes sense for providing context to translators

PR #2642, opened 2018-03-09
kick-docker

gl2748:


gl2748:
No longer necessary ... dockerhub was reborn.

PR #2645, opened 2018-03-10
Fix App query PropType and object compare

### LINKED ISSUE #2635: App component expects query prop to be string, but it's an object
introduced in #2629 

```
  <-- GET /
Warning: Failed prop type: Invalid prop `query` of type `object` supplied to `App`, expected `string`.
    in App (created by Connect(App))
    in Connect(App) (created by RouterContext)
    in RouterContext
    in IntlProvider (created by Translator)
    in Translator (created by Connect(Translator))
    in Connect(Translator)
    in Provider

```

caleblogan:
fixes #2635 

- use lodash isEqual to compare query object
- change query propType from string to object

### DIFF
--- a/src/app/components/App.jsx
+++ b/src/app/components/App.jsx
@@ -16,6 +16,7 @@ import { serverApiRecordEvent } from 'app/utils/ServerApiClient';
 import { key_utils } from '@steemit/steem-js/lib/auth/ecc';
 import resolveRoute from 'app/ResolveRoute';
 import { VIEW_MODE_WHISTLE } from 'shared/constants';
+import isEqual from 'lodash/isEqual';
 
 const pageRequiresEntropy = path => {
     const { page } = resolveRoute(path);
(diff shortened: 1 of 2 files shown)


teukumuhas:
(Y)

bnchdrff:
since we're only worried about re-rendering if the sort param is changed why not specifically check for it rather than doing a potentially expensive deep equals? also i'm curious, when you did your original work, why did you choose to re-render App rather than one of the lower-down-the-tree components like PostFull? (i didn't look into this deeply)

caleblogan:
I didn't see `withRouter` or something similar in the codebase so I was trying to avoid that. I pushed another commit as a proposed fix using `withRouter`.

caleblogan:
I notice there is a SagaCancellationException for `watchLocationChange` thrown sometimes using withRouter. Is there a better way to get the location than `withRouter`?

bnchdrff:
@caleblogan check out my commit i pushed to your branch -- i think this is a nice way of handing it!

bnchdrff:
(we aren't using react router in the most kosher way in condenser and it makes work like this a little difficult!)

caleblogan:
That looks good! Thanks for working with me, learning a lot.

PR #2652, opened 2018-03-12
profile and cover image uploader

### LINKED ISSUE #2430: Community - Add ability to upload profile and cover photo images

roadscape:
Replaces PR #2595 by @voorash:

> Worked on this at the request of witness followbtcnews and suggestion of @TimCliff 
> 
> The change from steemit chat to steem chat should not have been in this branch but assuming that by the time this pull request is merged my other one that had that change in it will already be merged so it shouldn't produce any conflict.
> 
> closes #2430


src/app/components/modules/Settings.jsx
```diff
@@ -80,6 +82,55 @@ class Settings extends React.Component {
...
+                    field = this.state.cover_image;
+                } else {
+                    return;
+                }
+                field.props.onChange(image_md);
+            } else {
+                this.setState({ progress });
+            }
+            setTimeout(() => {
+                this.setState({ progress: {} });
```
roadscape: @voorash progress is cleared unconditionally after 4s? Is there a success/failure signal we can use?

voorash: Logic is the same as the upload image on the post page.  I will take a look at that tomorrow and get back to you.

voorash: So here is the code from the reply editor. I didn't include the comment when copying it over. It pretty much just clears the message that says it is uploading as well.  
```
setTimeout(() => { this.setState({ progress: {} }) }, 4000) // clear message
```
I ran a test on this to make sure it did what I thought it does.  If it took over 4s for the server to respond the uploading message would be cleared.  If it ultimately failed it would show a failure message as the timeout does nothing but clear the message.  If on the other hand at 6 seconds it makes a connection and moves on with the upload the progress is automatically set from the user saga to show the percentage uploaded and all will proceed as normal.

During the actual upload the progress is tracked with this
```
xhr.upload.onprogress = function (event) {
        if (event.lengthComputable) {
            const percent = Math.round((event.loaded / event.total) * 100)
            progress({message: `Uploading ${percent}%`})
             console.log('Upload', percent)
        }
    }
```

gl2748: If the error message is not dismissable, the time-out might have some validity... If however our error display component is dismissable ... we might not need the timeout.

pkattera:
Unfortunately, this change isn't working for me in my local environment (chrome 65 on Mac). I'm able to select a picture, but the image address is not populated in the field and the image doesn't update. Any idea why this isn't working for me?

cover

voorash:
@pkattera  I will take a look at it on chrome on my mac.  Also, I saw something pop up that said you were missing a translation can you confirm that there is nothing funky going on with your local copy of the branch.

roadscape:
Tested working & UI has been approved.

free999enigma:
It maybe a bug but you can upload an image on other profiles for avatar or cover which should not be possible (still you can't update other profiles after the image was uploaded)

https://i.gyazo.com/449a567fcecb106a9ac84104acbb7bf1.png

Without being logged in

https://i.gyazo.com/b5a6559ec91f3ab4ef801c7f6cb2e425.png

pkattera:
Thanks for catching this @free999enigma – were you actually able to change the photo for another user? I can press the upload link, select a photo and see the newly generated URL but I can't actually change the other person's photos.

@roadscape If anyone who is not the logged in user arrives at a settings page, we should probably redirect them one level up back to `/@username`. Otherwise, it's super confusing on the off chance you land on someone else's settings page. Thoughts?

free999enigma:
@pkattera i mentioned "(still you can't update other profiles after the image was uploaded)"

roadscape:
@pkattera nothing on the page will save until the user authenticates; we probably shouldn't render the settings page at all unless it's the logged in user. It's not possible to get to anyone's setting page (other than your own) via condenser interface, so it's a rare case users would end up on someone else's.

PR #2660, opened 2018-03-13
remove user related ssr

### LINKED ISSUE #2614: remove user-related SSR
see https://github.com/steemit/steempunks/issues/236 and steemit/condenser#2466

This might be more of a spike, until we are ready to fully rip out the account creation & recovery parts of our codebase.

bnchdrff:
closes #2614

the gist: during SSR we used to try & load some basic user details from
the db to partially hydrate state. by removing this step, server load
will go down a bit, and we can avoid some dangerous situations by
limiting user related queries to API requests only.

stop doing a "probably logged in" in-between state in the header -- this
fixes a disorienting in-between state in the header.

we only use koa-flash during account creation, so we can scrap it now
that we have a new faucet.

the only bits of session-related state we're using are:

* csrf token
* login challenge token
* user prefs (nightmode, locale)
* new_visit (used for the welcome banner)

also remove some tarantool danglers

finally, remove unused ref to signup bonus


src/shared/UniversalRender.jsx
```diff
@@ -351,9 +350,9 @@ async function universalRender({
                 onchain.content[key]['stats'] = contentStats(
                     onchain.content[key]
                 );
-                onchain.content[key]['active_votes'] = onchain.content[key][
-                    'active_votes'
-                ].filter(vote => vote.voter === offchain.account);
+                //onchain.content[key]['active_votes'] = onchain.content[key][
+                //    'active_votes'
+                //].filter(vote => vote.voter === offchain.account);
```
roadscape: This feature is used to reduce page size drastically by removing (potentially 1000's of) vote records which are not relevant to the logged in user; would be ideal to preserve it if at all possible.

bnchdrff: cool, i'll work this part back in.

bnchdrff: actually what if we excluded these vote records for anon users, and then for logged-in users we request that data on the client side?

roadscape: When a client loads a SSR page do they immediately request `get_state` again? If yes: then we could probably trim all vote records here, because `get_state` will return the whole set anyway (w/o trimming). It would probably result in visual flashing however. If no: then they would probably need to; I'm not sure there's any other call to get this data (currently).

bnchdrff: the client ends up doing another `get_state` anyways

bnchdrff: ok i've excluded all vote data from the SSR response & i don't notice any visual flashing on page loads.

roadscape:
Testing on dev with cache disabled reveals some flickering:

image

roadscape:
@pkattera is the flashing a concern? It should only affect initial page load; successive navigation within the same tab will not exhibit this, afaict.

pkattera:
Flickering is never ideal. I want to test this for myself on steemitdev before giving a recommendation. Are there any techniques we can use to minimize the flickering effect?

pkattera:
The two items that seem to be most affected by the flickering from this change are:
1. The right hand side module on desktop homepage for logged in users which changes from 'New to Steemit' to 'Links'
2. The page title on the desktop home page, which flickers from '@username's Feed' to 'My feed'

Is there any way we can reduce this flicker? Since the right hand side module isn't the core focus of the page, we could set the opacity to 0, until know if the user is logged in, then, fade it in over 0.3s. 

With the title, as much as I don't like the flicker, we're best of displaying '@username's feed' rather than risk having a blank title for people on slow connections since people will be relying on that element for page context.

roadscape:
The flickering of votes is harder to prevent (because it relies on a heavy `get_state` call) but it's also the least jarring of the effects. The other flickering ("My Feed", side menu, login status) seems like it is completely preventable because we have immediate access to client login status (via localstorage). We need to figure out why those elements end up waiting for the heavy call to finish rather than processing immediately (and potentially before the browser's first render).

bnchdrff:
i've made two changes:

1) to reduce the disorienting flicker in the sidebar, the server-side render excludes the "new to steemit?" element. after login is completed on the client side, either that block or the "links" block is displayed with a fade-in animation

2) i localized the strings for "my feed" / "username's feed"

gl2748:
# Before
before

# After
after_ssr


# After with cache disabled
after_cache_disabled

PR #2662, opened 2018-03-14
allow existing user record during create_user

bnchdrff:
refs https://steemit.slack.com/archives/C5ADW8RQQ/p1521030862000166

### DIFF
--- a/src/server/api/general.js
+++ b/src/server/api/general.js
@@ -290,41 +290,35 @@ export default function useGeneralApi(app) {
         try {
             if (!emailRegex.test(email.toLowerCase()))
                 throw new Error('not valid email: ' + email);
-            const existingUser = yield findUser({
+            let user = yield findUser({
                 email: esc(email),
                 name: esc(name),
             });
-            if (existingUser) {
-                this.body = JSON.stringify({
-                    success: false,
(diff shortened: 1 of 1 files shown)


src/server/api/general.js
```diff
@@ -290,41 +290,35 @@ export default function useGeneralApi(app) {
         try {
             if (!emailRegex.test(email.toLowerCase()))
                 throw new Error('not valid email: ' + email);
-            const existingUser = yield findUser({
+            let user = yield findUser({
```
jnordberg: I think we should skip the lookup entirely, unless we are sure that the existing user instance won’t lead to an unrecoverable account

bnchdrff: i'll change it to an upsert

bnchdrff: actually should we remove any rows with matching usernames, and then create a new row for this approved user?

bnchdrff: should we remove rows w/ matching usernames or emails, or just usernames?

jnordberg: I was thinking we just add more users, or is there an unique index?

bnchdrff: no unique index; i changed it to just make a new user record (see commit "just make a new user record" below;))

PR #2670, opened 2018-03-17
do a phishing test using an in-repo list

### LINKED ISSUE #2616: Add blacklist for phishing sites (replace link with warning)
Lately we have a lot of scam sites that are used to steal people passwords, by impersonating Steemit, Steemconnect or just luring people to log in with their Steem credentials.

We need to be able to put list of such domains on a blacklist and then replace links to such sites with a warning and do not allow people to follow such links etc.

For example:
`[click here](http://therealsteemit.com/enteryourpassword)`
adding 'therealsteemit.com' to the blacklist should get rid of above link
plus variants like:
`https://therealsteemit.com/`
`http://signup.therealsteemit.com/`
etc.

bnchdrff:
closes #2616


Smittyvb:
You should add sleemit.com, and șteemit.com to the phishing list.

gl2748:
@Smittyvb thanks I will :-)

PR #2675, opened 2018-03-19
add Japanese language

roadscape:
Replaces #1916 by @inoue-a.


roadscape:
Far as I can tell, the only thing missing here is the language switch option itself.

PR #2676, opened 2018-03-19
in post lists, link comments/replies to parent, not root

### LINKED ISSUE #2580: clicking a comment from the replies page does not scroll down to the anchor
The URL has an anchor on it, and the comment is highlighted, but the user must scroll down to find it.  The page should scroll directly to it.

roadscape:
Resolves #2580

Currently, clicking a child comment on users comments/replies tabs will take you to the OP with the specific comment highlighted. This does not work well when there are a lot of comments.

The new behavior is to link to the direct parent, making the relevant comment much easier to spot. (It's still highlighted). Also, the comments count/icon links directly to the comment, but jumps to the comments section.

### DIFF
--- a/src/app/components/cards/PostSummary.jsx
+++ b/src/app/components/cards/PostSummary.jsx
@@ -103,27 +103,28 @@ class PostSummary extends React.Component {
         const archived = content.get('cashout_time') === '1969-12-31T23:59:59'; // TODO: audit after HF17. #1259
         const full_power = content.get('percent_steem_dollars') === 0;
 
-        let title_link_url;
-        let title_text = p.title;
-        let comments_link;
+        let post_url;
+        let title_text;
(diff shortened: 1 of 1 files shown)


roadscape:
(Also renamed `title_link_url` to `post_url` and `comments_link` to `comments_url` to be more semantic.)

gl2748:
# Before
Visiting a comment for **the first time**, from the user's `my-comments` page did not auto-magically scroll to that comment in the comment feed.
before1
Visiting a comment for **the second time**, from the user's `my-comments` page **did** auto-magically scroll to that comment in the comment feed.
before2

# After
Regardless of first or subsequent navigation to a specific comment from the user's `my-comments` page the comment is loaded by itself along with the direct parent.
after1

PR #2685, opened 2018-03-21
2569 bandwidth error modal

### LINKED ISSUE #2569: New bandwidth error modal UI with 'Buy Steem' button
As a user, when I perform an action that fails due to a lack of bandwidth, I need to see actionable items that help me resolve the issue, so I can continue interacting on Steemit. 

----

When a user interaction fails due to a lack of bandwidth, let's display a modal window which explains the issue clearly and provide buttons, prompting them to buy Steem or power up to Steem Power.

_Notes:_ 

- This will only be displayed to users when they're prevented from doing an interaction / transaction. 
- We'll continue to research better medium term options around solving bandwidth issues.

gl2748:
closes #2569 

Some things to note:
- Looks like blocktrades regex is failing to acknowledge usernames that include a `-` when reading query params.
- The error that is thrown from the blockchain does not detail the origination of the transaction, therefore the modal cannot (as implemented) include a reference to the particular type of transaction that failed because the user is out of bandwidth (e.g. Vote, Transfer, etc...). Therefore the modal copy simply reads 'Your Transaction Failed to Process'
- It would be nice to include translations of all this additional copy... @originated thoughts on that?
- @pkattera At this stage - I have not implemented the updated modal styles, I think this perhaps is a seperate task, estimated at a 1 - would be nice to spend some time adding a transition effect there.

modal


src/app/components/modules/Modals.jsx
```diff
@@ -66,6 +85,15 @@ class Modals extends React.Component {
               })
             : [];
 
+        const buySteemPower = e => {
+            if (e && e.preventDefault) e.preventDefault();
+            const new_window = window.open();
+            new_window.opener = null;
+            new_window.location =
+                'https://blocktrades.us/?input_coin_type=eth&output_coin_type=steem&receive_address=' +
```
pkattera: Right now, it points to buy 'Steem' from Blocktrades instead of 'Steem Power'. Looks like we need to use `coin_type=steem_power`.

screen shot 2018-03-21 at 3 00 34 pm

gl2748: Ok thanks.

gl2748:
in dev, error message is different,
per last commit - 
```
err key at current check: Transaction broadcast error. err string at current check RPCError: plugin exception:Account: test-safari bandwidth limit exeeded. Please wait to transact or power up STEEM.
app.7ac47ba0d148d489a54e.js:1 Transaction Error:Transaction broadcast error.
app.7ac47ba0d148d489a54e.js:1 [Violation] 'setTimeout' handler took 77ms
```
Need to fix.

gl2748:
TODO, note in above error on Dev - `bandwidth limit exeeded`, `exeeded != exceeded`!

pkattera:
This PR looks good to me. If it passes a technical review, let's get this into production asap.

gl2748:
This is all good to go. Waiting until AWS is back up and running.

PR #2703, opened 2018-03-30
community - Show STEEM breakdown on pending payout. Revert paid payout display.

eonwarped:
Per feedback in #2460  and referenced in #2429 , seems prudent not to show this inaccurate breakdown in the historical payout, so that's done in this pull request. At the very least they'll see it in their historical payouts. Not ideal, but also not inaccurate.

In addition, with the sbd_print_rate changes, I felt we should also show this in the pending payout. This at least seems to be strictly better for the current situation.

I've added some unit tests to capture the changes made here as well.

This is what it looks like:

payoutsplitchange

You'll note that the right paren is sort of clipped, not sure how to fix that.


roadscape:
Thanks for the contribution, @eonwarped! The PR review continues in #2746.

PR #2705, opened 2018-04-02
filter non-string tags

### LINKED ISSUE #2693: A blog post by me corrupts all the feeds of my followers and my blog
Hello,

We are developing an application on Steem blockchain. The app accidentally posted through Steemit via Steem Connect v2 then everything related with my posts disappeared.

Here is my account: https://steemit.com/@ikidnapmyself
And it returns "Sorry! This page doesn't exist."

Here is the post that making trouble: https://steemd.com/dngo/@ikidnapmyself/my-awesome-title

Any follower of mine can not reach their feeds and my page too.

Just in case; everything work properly except the post. I am able to post, edit etc. And I tried to delete my post using Postman with comment_delete scope to broadcast the operation, it did not work.
```
{
    "error": "unauthorized_client",
    "error_description": "This access_token allow you to broadcast transaction only for the account @ikidnapmyself"
}
```
 I also with SteemJS GitHub example. There is a reply on the post so I can not delete it.
```
{
  "cause": {},
  "isOperational": true,
  "code": -32000,
  "data": {
    "code": 10,
    "name": "assert_exception",
    "message": "Assert Exception",
    "stack": [
      {
        "context": {
          "level": "error",
          "file": "steem_evaluator.cpp",
          "line": 467,
          "method": "do_apply",
          "hostname": "",
          "timestamp": "2018-03-25T02:12:27"
        },
        "format": "comment.children == 0: Cannot delete a comment with replies.",
        "data": {}
      },
```

Thanks in advance!

kirkins:
Fixes #2693

Filtering out non-string based tags.

Examples of pages that break.

https://steemit.com/@ikidnapmyself

https://steemit.com/dngo/@ikidnapmyself/my-awesome-title

### DIFF
--- a/src/app/components/elements/TagList.jsx
+++ b/src/app/components/elements/TagList.jsx
@@ -14,6 +14,7 @@ export default ({ post, horizontal, single }) => {
 
     const json = post.json_metadata;
     let tags = [];
+    tags = tags.filter(tag => typeof tag === 'string');
 
     try {
         if (typeof json == 'object') {
@@ -32,10 +33,15 @@ export default ({ post, horizontal, single }) => {
     // Category should always be first.
     tags.unshift(post.category);
 
-    // Uniqueness filter.
-    tags = tags.filter(
-        (value, index, self) => value && self.indexOf(value) === index
-    );
+    tags = tags
+        .filter(
+            // Uniqueness filter.
+            (value, index, self) => value && self.indexOf(value) === index
+        )
+        .filter(
+            // Filter non-strings
+            tag => typeof tag === 'string'
+        );
 
     if (horizontal) {
         // show it as a dropdown in Preview
--- a/src/app/components/pages/TagsIndex.jsx
+++ b/src/app/components/pages/TagsIndex.jsx
@@ -55,6 +55,8 @@ export default class TagsIndex extends React.Component {
         //console.log('-- TagsIndex.render -->', tagsAll.toJS());
         const { order } = this.state;
         let tags = tagsAll;
+        //filter out non-strings
+        tags = tags.filter(tag => typeof tag === 'string');
 
         const rows = tags
             .filter(
--- a/src/app/utils/StateFunctions.js
+++ b/src/app/utils/StateFunctions.js
@@ -163,6 +163,7 @@ export function contentStats(content) {
     // Combine tags+category to check nsfw status
     const json = content.get('json_metadata');
     let tags = [];
+    tags.push(content.get('category'));
     try {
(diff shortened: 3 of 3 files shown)


src/app/components/elements/TagList.jsx
```diff
@@ -14,6 +14,7 @@ export default ({ post, horizontal, single }) => {
 
     const json = post.json_metadata;
     let tags = [];
+    tags = tags.filter(tag => typeof tag === 'string');
```
bnchdrff: `tags` is empty at this point, so this filter is not necessary here

src/app/components/pages/TagsIndex.jsx
```diff
@@ -55,6 +55,8 @@ export default class TagsIndex extends React.Component {
         //console.log('-- TagsIndex.render -->', tagsAll.toJS());
         const { order } = this.state;
         let tags = tagsAll;
+        //filter out non-strings
+        tags = tags.filter(tag => typeof tag === 'string');
```
bnchdrff: this inline funciton is reused a few times -- even though it's simple (for now) it should be a named function

src/app/components/elements/TagList.jsx
kirkins: yes, this line is kind of pointless.

src/app/components/elements/TagList.jsx
```diff
@@ -32,10 +32,15 @@ export default ({ post, horizontal, single }) => {
...
 
-    // Uniqueness filter.
-    tags = tags.filter(
-        (value, index, self) => value && self.indexOf(value) === index
-    );
+    tags = tags
+        .filter(
+            // Uniqueness filter.
+            (value, index, self) => value && self.indexOf(value) === index
+        )
```
kirkins: I guess I should put the uniqueness filter into the named function and use that here.

roadscape:
Hi @kirkins, it looks like the tags page is broken (https://github.com/steemit/condenser/issues/2741), and I noticed this PR was the last to modify it. Could you please check if regression is due to this PR?

bnchdrff:
i just confirmed with a local test that this PR caused the tags page to break.

the tags page receives a `Map` of tag objects, but the `filterTags` function expects an array of tags. we should revert this work for now.

kirkins:
Sorry about that, just assumed everything was ok when the tests passed.

Seems the roll back is live and the broken tag problem is still fixed.

sneak:
Please write tests for the tags page (that fail under this commit) as well as a failing test for the user/post page (that goes green under this commit).  I don't like relying on institutional knowledge and public-facing testing.

PR #2734, opened 2018-04-16
Community - Fix "Account not found" but when using usernames with role suffixes

### LINKED ISSUE #2593: Unable to change any settings/profile images since before Christmas
Account: @mobbs

I've been stuck with my christmas theme for months now. any attempt to change simply gives me the error 'Server returned error' - no other information. 

Timcliff and I have tried numerous variations; Chrome, Explorer, Firefox on Windows. Tried on another device; android phone, all of these with and without VPN. Attempted to change and clear all details in the boxes, various types of URLs to the images - all simply 'server returned error'

The same URLs work on both a friends account and my second account. It seems it's specifically my account that holds an issue. It's not limited to the URLs, *any* change comes with 'server returned error'
image

eonwarped:
Fixes #2593  assuming it has to do with a role-suffix on the user name when doing the login flow.

Description of Problem:
When using eonwarped/active and corresponding password to login while performing a transaction operation such as powering up or transfering funds, or in the case of the referenced issue, changing the account profile, you would get an error like this:

erroraccountnotfound

The reason was because of the suffix role, and it simply needs to be stripped. That's what's in this pull request.

### DIFF
--- a/src/app/redux/AuthSaga.js
+++ b/src/app/redux/AuthSaga.js
@@ -132,6 +132,11 @@ export function* findSigningKey({ opType, username, password }) {
     username = username || currentUsername;
     if (!username) return null;
 
+    if (username.indexOf('/') > -1) {
+        // "alice/active" will login only with Alices active key
+        username = username.split('/')[0];
+    }
+
     const private_keys =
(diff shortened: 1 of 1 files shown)


roadscape:
Thank you for the fix!

PR #2736, opened 2018-04-17
use latest feed price for wallet value

### LINKED ISSUE #2205: Eliminate 3 day averaging of estimated steem account value in wallet view
Motivation: whenever there's an increase in the price of Steem, it's likely to attract new buyers. But after they buy, they experience buyer's remorse because condenser reports a 3 day average of the price which is naturally lower than the actual price. Please replace this with a more timely value estimate, or remove it entirely. I strongly favor using a more accurate price versus removal, by the way.

originated:
Fixes #2205 - (update) I'm seeing a ci/circleci: run-blackboxtest error looking at it.

### DIFF
--- a/src/app/components/modules/UserWallet.jsx
+++ b/src/app/components/modules/UserWallet.jsx
@@ -26,6 +26,7 @@ import {
 import * as transactionActions from 'app/redux/TransactionReducer';
 import * as globalActions from 'app/redux/GlobalReducer';
 import DropdownMenu from 'app/components/elements/DropdownMenu';
+import * as userActions from 'app/redux/UserReducer';
 
 const assetPrecision = 1000;
 
@@ -760,11 +761,17 @@ export default connect(
     // mapStateToProps
     (state, ownProps) => {
         let price_per_steem = undefined;
-        const feed_price = state.global.get('feed_price');
-        if (feed_price && feed_price.has('base') && feed_price.has('quote')) {
-            const { base, quote } = feed_price.toJS();
-            if (/ SBD$/.test(base) && / STEEM$/.test(quote))
-                price_per_steem = parseFloat(base.split(' ')[0]);
+        let steem_api_price = state.user.get('steem_api_price');
+        if (steem_api_price) {
+            price_per_steem = steem_api_price;
+        } else {
+            // remove else conditional once tested
+            const feed_price = state.global.get('feed_price');
+            if (feed_price && feed_price.has('base') && feed_price.has('quote')) {
+                const { base, quote } = feed_price.toJS();
+                if (/ SBD$/.test(base) && / STEEM$/.test(quote))
+                    price_per_steem = parseFloat(base.split(' ')[0]);
+            }
         }
         const savings_withdraws = state.user.get('savings_withdraws');
         const gprops = state.global.get('props');
--- a/src/app/components/modules/UserWallet.jsx
+++ b/src/app/components/modules/UserWallet.jsx
@@ -26,7 +26,6 @@ import {
 import * as transactionActions from 'app/redux/TransactionReducer';
 import * as globalActions from 'app/redux/GlobalReducer';
 import DropdownMenu from 'app/components/elements/DropdownMenu';
-import * as userActions from 'app/redux/UserReducer';
 
 const assetPrecision = 1000;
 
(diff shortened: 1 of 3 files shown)


originated:
getRecentPrice function updated to utilize last element looks right.  Something seems off with Redux getting the new value in the wallet

src/app/components/modules/UserWallet.jsx
```diff
@@ -26,6 +26,7 @@ import {
 } from 'app/client_config';
 import * as transactionActions from 'app/redux/TransactionReducer';
 import * as globalActions from 'app/redux/GlobalReducer';
+import * as userActions from 'app/redux/UserReducer';
```
roadscape: looks unused

src/app/components/modules/UserWallet.jsx
```diff
@@ -765,11 +766,17 @@ export default connect(
...
     (state, ownProps) => {
         let price_per_steem = undefined;
-        const feed_price = state.global.get('feed_price');
-        if (feed_price && feed_price.has('base') && feed_price.has('quote')) {
-            const { base, quote } = feed_price.toJS();
-            if (/ SBD$/.test(base) && / STEEM$/.test(quote))
-                price_per_steem = parseFloat(base.split(' ')[0]);
+        let steem_api_price = state.user.get('steem_api_price');
+        if (steem_api_price) {
+            price_per_steem = steem_api_price;
```
roadscape: probably needs to do base/quote here

src/app/redux/UserReducer.js
```diff
@@ -68,6 +69,9 @@ export default function reducer(state = defaultState, action) {
             });
         }
 
+        case SET_PRICE_FEED:
+            return state.set('steemApiPrice', payload);
```
roadscape: typo? in other places this is `steem_api_price`

src/app/redux/UserSaga.js
```diff
@@ -67,6 +68,11 @@ export function* watchRemoveHighSecurityKeys() {
     yield* takeLatest('@@router/LOCATION_CHANGE', removeHighSecurityKeys);
 }
 
+function* getRecentPrice() {
+    const feed = yield call([api, api.getFeedHistoryAysnc]);
```
roadscape: typo: `Aysnc`

src/app/redux/UserSaga.js
```diff
@@ -67,6 +68,11 @@ export function* watchRemoveHighSecurityKeys() {
...
+    if (feed) yield put(userActions.setPriceFeed({price: feed[feed.length - 1]}));
```
roadscape: `getFeedHistoryAsync` returns an object with the feed under the `price_history` key

roadscape:
http://localhost:8080/@test25/transfers
0.509 STEEM POWER
dev: est value = $1.22 (~2.40/STEEM)
prod: est value = $1.27 (~2.50/STEEM)

Seems to be working now. 2.50 is the rolling median, and ~2.40 is the last median.

src/app/redux/UserSaga.js
```diff
@@ -67,6 +68,15 @@ export function* watchRemoveHighSecurityKeys() {
     yield* takeLatest('@@router/LOCATION_CHANGE', removeHighSecurityKeys);
 }
 
+function* getLatestFeedPrice() {
+    const history = yield call([api, api.getFeedHistoryAsync]);
```
gl2748: I think putting this in a try catch would be nice ...

src/app/redux/UserSaga.js
```diff
@@ -67,6 +68,19 @@ export function* watchRemoveHighSecurityKeys() {
     yield* takeLatest('@@router/LOCATION_CHANGE', removeHighSecurityKeys);
 }
 
+function* getLatestFeedPrice() {
+    try {
+        const history = yield call([api, api.getFeedHistoryAsync]);
+        if (history) {
```
gl2748: We don't need this if statement anymore, because if the `yield call` fails, the catch will get it...

roadscape: Is it possible `yield call` can return blank without throwing an exception?

gl2748: Cannot tell - where is this endpoint documented?

roadscape: I'm not sure it's documented anywhere; https://github.com/steemit/steem/blob/bb8ca476f33c08434274f2d76333fcb74cfa58b8/libraries/plugins/apis/database_api/database_api.cpp#L158 is where the API is defined and https://github.com/steemit/steem/blob/62b48877d9f731c3fe00ef818e3324a0a3de3e63/libraries/chain/database.cpp#L3082 is where the structure is maintained

roadscape:
Rebased on `master`.

PR #2746, opened 2018-04-19
show STEEM breakdown on pending payout, revert paid payout display

roadscape:
Replaces #2703 by @eonwarped:
> Per feedback in #2460  and referenced in #2429 , seems prudent not to show this inaccurate breakdown in the historical payout, so that's done in this pull request. At the very least they'll see it in their historical payouts. Not ideal, but also not inaccurate.
> 
> In addition, with the sbd_print_rate changes, I felt we should also show this in the pending payout. This at least seems to be strictly better for the current situation.
> 
> I've added some unit tests to capture the changes made here as well.
> 
> This is what it looks like:
> 
> payoutsplitchange
> 
> You'll note that the right paren is sort of clipped, not sure how to fix that.
> 
>


roadscape:
(I meant to commit the other const's as a separate PR but the husky hook insisted on adding my unstaged changes.) The implementation looks good. It looks like the print rate has gone back to 100% so to test we need to stick in a mock value. Main remaining concern is text overflow.

eonwarped:
Yes, there is a test covering both cases and it should look the same now that the print rate is back to normal.

 Is there anything I need to do or this is something you will take over?

roadscape:
We'll take it from here, thanks!

PR #2759, opened 2018-04-25
Remove un-needed logging

bnchdrff:


bnchdrff:
@gl2748 i removed the bad part of that function - see https://github.com/steemit/condenser/pull/2759/files#diff-10be560a8e29acae0b34682a14dde0c0 -- now it only logs to overseer

PR #2761, opened 2018-04-26
2599 transaction modal

### LINKED ISSUE #2599: transaction confirmation modal
When I do a transfer to another user, it shows a 'standard' login page. It would be very useful if this page showed a summary of what was about to be signed, such as "sending X STEEM tokens to @user; memo: xyz" (or something like that). It also seems unnecessary to have the "Sign up. Get STEEM" option at the bottom, since a user who is already logged in and sending tokens already has an account.

image

gl2748:
closes #2599 

confirm_transaction_001


gl2748:
- [x] make the submit button say 'next'.
- [ ] and next on the confirmation modal rather than 'ok'
- [ ] try... to detect, make it 'confirm'...
- [x] Disable the username field.
- [x] Make title of signin-to-transfer-modal 'Sign to complete transfer.'
- [ ] 'Figure out what remember me for this session' does in the context of a transfer.

gl2748:
screenshot from 2018-04-30 17-34-46
screenshot from 2018-04-30 17-34-55

Issue #2767, opened 2018-04-29
Login/Signup buttons overflowing when search bar is expanded


Nicknyr:
When the search bar at the top is clicked so that it expands it causes the sign up button to overlap the log in link and both overflow like so.

I can fix this, just want to make sure no one is already on it.

![screenshot from 2018-04-29 01-07-25](https://user-images.githubusercontent.com/8207697/39403435-2a7977ba-4b4a-11e8-9116-0b72a444f4d0.png)

gl2748:
closed by #2772

PR #2768, opened 2018-04-29
Fixes issue #2767 - Login/Signup buttons overflowing

Nicknyr:
Adjusted the flexbox column sizes in the header so that there is no overflowing of the signup/login links when the search bar in the header is expanded.


roadscape:
Hi @Nicknyr thanks for the contribution! The review continues in #2772.

PR #2769, opened 2018-04-30
add default payout options

### LINKED ISSUE #2512: adjustments to reward options
1. add ability to decline payouts on all *comments* via option on the Settings page
   - no dropdown next to comments, just a reminder it will be declined if the option is set
2. temporarily disable "power up 100%" reward option
   - because when SBD > 1, it's causing users to lose out on rewards
3. remove the "upvote post" checkbox
   - it was needed only until the 30-min reverse auction was implemented

roadscape:
Closes #2512 

 - removed autovote checkbox
 - move payoutType selection to Settings page, and split into Blog/Comment options
 - post editor does not show payout option unless non-default is selected.


roadscape:
Here's what it looks like when non-default (50%/50%) option selected:

image

And for declined:

image

When the default is chosen, nothing is shown:

image

---

Settings page:

image

roadscape:
After this passes @pkattera's UI review, need one more round of functional testing for posts and (especially) comments.

free999enigma:
is just an opinion but this updates looks more like an "extra unnecessary step" yeh the "removed autovote checkbox" it's nice but still people will upvote themself

roadscape:
Tested working for blogs and comments.

PR #2772, opened 2018-04-30
fix login/signup buttons overflowing

roadscape:
Replaces #2768 by @Nicknyr:

> ### Fixes issue #2767 - Login/Signup buttons overflowing
>
> Adjusted the flexbox column sizes in the header so that there is no overflowing of the signup/login links when the search bar in the header is expanded.


roadscape:
this branch
image

prod
image

gl2748:
Still seeing this on some screens.
screenshot from 2018-05-01 14-45-06

I think the issue is that there's not enough real estate to the right of the centered sort options (trending, new etc...) on 'Medium' screens. Therefore propose disabling the expandable search on medium, and having that feature only for large...

PR #2774, opened 2018-05-02
ExplorePost: remove steemdata.com from options

crokkon:
phist

phist.steemdata.com was discontinued together with the SteemData.com MongoDB instance on May 1st 2018:
https://steemit.com/steemdata/@furion/steemdata-is-shutting-down-on-may-1st


roadscape:
Looks good, thank you for the fix!

PR #2780, opened 2018-05-06
remove title from reddit logo SVG to show localized `share_on_reddit` string

crokkon:
The title in the SVG is shown instead of the parent link title if present.

share_reddit

The localized `share_on_reddit` string on the parent link is not shown:
https://github.com/steemit/condenser/blob/0f7e5a2f8316f40a342e46e5d7e9d2a991b1cb75/src/app/components/cards/PostFull.jsx#L313


roadscape:
Thanks for the contribution, moved to local branch at https://github.com/steemit/condenser/pull/2806

Issue #332, opened 2016-09-23
Further SEO Page Title Refinements


mcifani:
- [x] Ensure all titles are always first letter capitalized (consistency and user readability) 

- [x] Further tweaks to page titles with no category/tag selected

PR #2790, opened 2018-05-14
reduce server-side fs reads, refs #332

gl2748:
closes https://github.com/steemit/steempunks/issues/332

With the changes in this PR and following profiling commands

I am seeing:
```
232.76 kb/s avg before.
253.1 kb/s avg after.

8.73 % increase in kb/s.

685.3 req/100s before
899 req/100s after

31.24% increase in requests/s
```

Per:
https://gist.github.com/gl2748/5ee5b145ef2f168895cce74a3466f321

This on the parent process as reported by:
```
yarn run build && SDC_DATABASE_URL="mysql://root@localhost/steemit_dev" NODE_ENV=production node --prof --log_timer_events lib/server/index.js
```
So not too bad. I have a hunch that fs reads for translations might also be slowing things down server side.

### See below.

### DIFF
--- a/src/server/app_render.jsx
+++ b/src/server/app_render.jsx
@@ -6,31 +6,19 @@ import { serverRender } from '../shared/UniversalRender';
 import models from 'db/models';
 import secureRandom from 'secure-random';
 import ErrorPage from 'server/server-error';
-import fs from 'fs';
 import { determineViewMode } from '../app/utils/Links';
+import { getSupportedLocales } from './utils/misc';
 
 const path = require('path');
 const ROOT = path.join(__dirname, '../..');
 const DB_RECONNECT_TIMEOUT =
     process.env.NODE_ENV === 'development' ? 1000 * 60 * 60 : 1000 * 60 * 10;
 
-function getSupportedLocales() {
-    const locales = [];
-    const files = fs.readdirSync(path.join(ROOT, 'src/app/locales'));
-    for (const filename of files) {
-        const match_res = filename.match(/(\w+)\.json?$/);
-        if (match_res) locales.push(match_res[1]);
-    }
-    return locales;
-}
-
 const supportedLocales = getSupportedLocales();
 
-async function appRender(ctx) {
+async function appRender(ctx, locales = false, resolvedAssets = false) {
     ctx.state.requestTimer.startTimer('appRender_ms');
-
     const store = {};
-
     // This is the part of SSR where we make session-specific changes:
     try {
         let userPreferences = {};
@@ -48,6 +36,7 @@ async function appRender(ctx) {
         if (!userPreferences.locale) {
             let locale = ctx.getLocaleFromHeader();
             if (locale) locale = locale.substring(0, 2);
+            const supportedLocales = locales ? locales : getSupportedLocales();
             const localeIsSupported = supportedLocales.find(l => l === locale);
             if (!localeIsSupported) locale = 'en';
             userPreferences.locale = locale;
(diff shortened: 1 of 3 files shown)


src/server/app_render.jsx
```diff
@@ -88,20 +74,11 @@ async function appRender(ctx) {
...
-        const assets_filename =
-            ROOT +
-            (process.env.NODE_ENV === 'production'
-                ? '/tmp/webpack-stats-prod.json'
-                : '/tmp/webpack-stats-dev.json');
-        const assets = require(assets_filename);
-
         // Don't cache assets name on dev
+        // TODO: Investigate removing this.
         if (process.env.NODE_ENV === 'development') {
```
bnchdrff: what happens when `process.env.NODE_ENV === 'development'`? it seems like the assets data required at startup will be removed but it will not be restored.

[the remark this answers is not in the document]
gl2748: I think originally it might be to get some-kind of hot-reloading of assets for local dev. I think it can be removed as we do not typically change the assets while developing locally and requiring a server-restart is not as onerous as perhaps it once was...

https://nodejs.org/api/modules.html#modules_caching

src/server/server.js
```diff
@@ -78,6 +78,15 @@ if (env === 'development') {
     );
 }
 
+if (process.env.NODE_ENV === 'production') {
+    const resolvedAssets = require(path.join(
```
bnchdrff: define resolvedAssets outside of this conditional block so that it's false rather than undefined in the surrounding block where it's used later, to reduce ambiguity

webpack/base.config.js
```diff
@@ -59,6 +59,7 @@ module.exports = {
             {test: /\.(jpe?g|png)/, use: 'url-loader?limit=4096'},
             {test: /\.json$/, use: 'json-loader'},
             {test: /\.js$|\.jsx$/, exclude: /node_modules/, use: 'babel-loader'},
+            // All svgs are loaded inline...
```
bnchdrff: put this in another pr where we address the removal of svgs from the webpack stats output

src/server/server.js
```diff
@@ -78,6 +78,15 @@ if (env === 'development') {
...
+        __dirname,
+        '../..',
+        '/tmp/webpack-stats-prod.json'
+    ));
+    const supportedLocales = getSupportedLocales();
```
bnchdrff: same thing here -- if supportedLocales is defined as false and then conditionally set to the output of getSupportedLocales then we can scrap the conditional mess below

src/server/app_render.jsx
```diff
@@ -88,20 +75,14 @@ async function appRender(ctx) {
             userPreferences,
             offchain,
         });
-
-        // Assets name are found in `webpack-stats` file
```
bnchdrff: please keep these & other comments

gl2748: Done.

src/server/app_render.jsx
```diff
@@ -88,20 +75,15 @@ async function appRender(ctx) {
...
-                ? '/tmp/webpack-stats-prod.json'
-                : '/tmp/webpack-stats-dev.json');
-        const assets = require(assets_filename);
-
-        // Don't cache assets name on dev
-        if (process.env.NODE_ENV === 'development') {
+        let assets;
+        if (!resolvedAssets) {
+            // Assets name are found in `webpack-stats` file
+            const assets_filename = ROOT + '/tmp/webpack-stats-dev.json';
```
roadscape: Don't we need to keep the `process.env.NODE_ENV === 'production'` switch?

roadscape: Oh, I see. `webpack-stats-prod.json` is resolved in `src/server/server.js` for `production`.

I guess the dev logic is here because it allows hot reloading?

roadscape:
Changes LGTM, ready for testing

PR #2795, opened 2018-05-15
updated DMCA blacklist

Gandalf-the-Grey:


Smittyvb:
Why was [this post](https://busy.org/@taslimabegum/can-t-crack-fl-studio-12-5-on-windows-10) added to the DMCA blacklist? Although the post is about piracy, it doesn't give any instructions on how to do it, nor does it have any copyrighted material, as the post is completely original content.

PR #2813, opened 2018-05-24
2253 remove rfc again

gl2748:
This PR removes React-Foundation-Components dependencies from everything except the `Voting.jsx` component. It is part of a succession of PRs.

refs #2253 

Outside of task but nice to have:
- [x]  Remove React-Foundation-Components from Voting.jsx  - see: https://github.com/steemit/condenser/compare/2253-remove-rfc-again-part-2?expand=1

PR #2817, opened 2018-05-25
2253 remove rfc again part 2

### LINKED ISSUE #2804: rewrite voting widget
rewrite voting widget so that it does not depend on react-foundation-components, and remove react-foundation-components

see #2214 and #2253

gl2748:
Closes #2804 

This removes all the React-Foundation-Components from the `Voting.jsx` component. 

It plays on top of the PR here: #2813


gl2748:
This now has tests to cover the changes to the Voting Component necessitated by the removal of RFC.

gl2748:
Refs: #2804 - This PR includes removal of React-Foundation-Components from the Voting widget.

roadscape:
Tested working. All previous issues resolved, though I notice that the weight is saved on drag finish rather than on submit; this is a change in behavior, is it more intuitive? Also there is a change to slider threshold const for testing which should be reverted.

PR #2821, opened 2018-05-26
fix typo in tos

gl2748:


src/app/components/pages/Privacy.jsx
```diff
@@ -514,7 +514,7 @@ class Privacy extends React.Component {
                         <span className="c1" />
                     </p>
                     <p className="c2">
-                        <span className="c6">Personal Information</span>
+                        <h4 className="c6">Personal Information</h4>
```
bnchdrff: what are all these funny class names?

gl2748: Copy and pasted from the google-doc generated html.

bnchdrff: what is wrong with the TOS currently that this PR fixes? is this a style change?

bnchdrff: ah ic. ok. this makes more sense now :)

PR #2828, opened 2018-05-30
distinguish disabled witnesses

roadscape:
Replaces PR #2228 by @netuoso:

> ### Issue
> There is no distinction between a disabled and active witness on the witness page.
> 
> ### Solution
> Add a logic check and change the CSS for the name of the witness.
> 
> ### Summary
> If there are disabled witnesses in the top 50, users are unaware because the Condenser witness page does not show any distinction. This pull request contains code that will perform a basic check on the witness price feed and signing key. If the price feed is blank or the signing key is the disabled key, the witness name will be grey and strikethrough.
> 
> ### Screenshots
> - Without CSS
> screen shot 2017-12-26 at 11 08 24 pm
> - With CSS
> screen shot 2017-12-26 at 11 08 12 pm
> 
> ### Extra Notes
> This pull request should also address https://github.com/steemit/condenser/issues/1610 by adding a `forceUpdate` after applying the witness vote. Sometimes a simple `setState` does not properly trigger the `ComponentShouldUpdate`.

### DIFF
--- a/src/app/components/pages/Witnesses.jsx
+++ b/src/app/components/pages/Witnesses.jsx
@@ -98,7 +98,6 @@ class Witnesses extends React.Component {
             const signingKey = item.get('signing_key');
             const isDisabled = signingKey == DISABLED_SIGNING_KEY;
             const lastBlock = item.get('last_confirmed_block_num');
-            const priceFeed = item.get('sbd_exchange_rate');
             const classUp =
                 'Voting__button Voting__button-up' +
                 (myVote === true ? ' Voting__button--upvoted' : '');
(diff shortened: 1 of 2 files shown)


corkyk:
Will the force_update solve the insta-unvote issue where writeins are immediately removed when a user (nearly always) instinctively clicks their newly appeared name or does need to be another PR?

netuoso:
@corkyk That was my intention and in practice it seemed to have worked. I noticed the vote was going thru but the update didn't occur, and forceUpdate just tries to push that through.

I haven't been feeling great lately but I am getting back on the ball and seeing a lot of activity on issues I have been meaning to get around to. Thanks for keeping it rolling.

Issue #2785, opened 2018-05-10
Fix typo on transfers screen


pkattera:
The text shouldn't reference a 'Steemit account' because technically, there are no Steemit accounts, just Steem accounts.

![screen shot 2018-05-10 at 2 02 40 pm](https://user-images.githubusercontent.com/1121139/39894159-d44fa750-545a-11e8-8608-ead46df32e32.png)

PR #2836, opened 2018-06-02
Change Steemit on transfer page to Steem in all languages

kirkins:
Hi just looking through your issues for easy things an outsider could do.

I found #2785, just changing Steemit to Steem, was being loaded dynamically from %(APP_NAME)s  so it just changed that to hard-coded "Steem".


roadscape:
@kirkins could you please check what happened to `src/app/locales/ja.json`? It's a very large diff.

kirkins:
@roadscape I was changing the indentation because it didn't match the others. I changed it back to make it easier.

roadscape:
Merged to local PR #2861

PR #2838, opened 2018-06-02
Make Links sidebar use translations if availible

kirkins:
I noticed the sidebar doesn't look to the translations file to see if the english can be replaced.

"my blog" already existed but the other 2 links didn't.

I also so other reference to "my_X" but without an `_` so I re-wrote for consistency.

I will make another PR to add some French translations for things that show on the front page.


roadscape:
Merged to local PR #2861

PR #2839, opened 2018-06-02
French for front-page and few other translations

kirkins:
This commit combined with #2838 will ensure that everything on the front page is translated to french.

Screen-shot of just running this branch below, with #2838 the other 2 links in the right sidebar and the title "Links" will be translated too.

screenshot from 2018-06-02 18-50-09


roadscape:
Merged to local PR #2861

Issue #2180, opened 2017-12-19
hive integration


roadscape:
Hivemind implements a condenser_api endpoint to make integration easy in the short term. In the medium term we'll want to deprecate the old APIs and leverage hive's improvements. This issue concerns only the initial integration.

#### Prerequisites

1. ~switch to appbase apis (condenser_api namespace) -- #2070~ N/A: handled by jussi
2. hive aws environment (or local for dev) -- https://github.com/steemit/hivemind/issues/30
3. routing options:
    a. jussi-based: overwrite old routes, call by call. requires regex-based excludes.
    b. integrate hivemind client lib (e.g. fork steem-js and use endpoint hive.steemit.com)
      - allows client to specify API version essentially

#### APIs

Begin by replacing the following calls to steem-js with hive-js:
```
    condenser_api.get_followers
    condenser_api.get_following
    condenser_api.get_follow_count
```

These are all straightforward pagination methods used in `src/app/redux/FetchDataSaga.js`:
```
    condenser_api.get_discussions_by_trending
    condenser_api.get_discussions_by_hot
    condenser_api.get_discussions_by_promoted
    condenser_api.get_discussions_by_created
    condenser_api.get_discussions_by_blog
    condenser_api.get_discussions_by_feed
    condenser_api.get_discussions_by_comments
    condenser_api.get_replies_by_last_update
```

`get_content` we need to be careful with because the returned state from hive may be a few seconds old. There are a few usages in condenser's codebase where we assume it returns live or intra-block state. Be sure to check `src/app/redux/TransactionSaga.js` where it calls getContent immediately after a comment or vote operation.
```
    condenser_api.get_content
```

`get_state` is a monolith we ultimately need to remove from condenser (https://github.com/steemit/condenser/issues/2244). Hive recreates all `get_state` responses (excluding data that does not appear to be used at all), with the exception of a few paths, namely `/@account/transfers` and `/~witnesses`. `/tags` endpoint is mostly compatible. Condenser will need to branch on the path argument of this call: transfers and witnesses requests go to steem-js; everything else goes to hive.
```
    condenser_api.get_state
```

#### Pre-appbase testing

With appbase, all calls should be directed to `https://hive-host/`. For testing condenser pre-appbase, use `https://hive-host/legacy` as the endpoint.

https://github.com/steemit/condenser/compare/2180-hive-integration

relativityboy:
We are not going to be forking steem-js. We'll be extending it to support hitting hive for relevant calls.

[...]

PR #2844, opened 2018-06-04
hive prep

roadscape:
Part of #2180 hive integration tasks


roadscape:
Will re-evaluate this approach, closing the PR but leaving the branch for now.

PR #2846, opened 2018-06-05
2180 hive integration

Pecking:


roadscape:
Hi @Pecking, this is an old branch/strategy which will not be used. Hive integration will mostly be done at the upstream/api level.

Issue #2847, opened 2018-06-05
'Voting weight is too small' error


roadscape:
With the next hardfork, this condition will be going away. Until then, it would be nice to clarify this message. Just trimming it to `Voting weight is too small, please accumulate more voting power or steem power.` would be a good improvement.

`RPCError: Assert Exception:info->abs_rshares > STEEM_VOTE_DUST_THRESHOLD || vote_weight == 0: Voting weight is too small, please accumulate more voting power or steem power. DISMISS`

netuoso:
This appears to be done and merged

PR #2858, opened 2018-06-10
Community - Clarify voting weight too small error

eonwarped:
#2847 Voting weight too small fix.

errorpic

Very quick fix to change error message.


roadscape:
Thank you! Review continues in local branch at #2860

PR #2859, opened 2018-06-11
Community - Fix witness voting not resolving properly, add feedback

eonwarped:
This addresses #1610, while also adding the spinner for in-progress votes.

For now the logging is still there, but I'll strip it out, the main point is explain what I saw:

-- prebroadcast Saga it tweaks the account state to remove the witness vote.
-- broadcast of the transaction , at some point it seems to re-fetch the account state, and often it undoes the state setting in the prebroadcast phase

So I've moved the state setting to the accepted phase of the transaction saga (post transaction broadcast).

That was quite an annoying bug.

(If you want a quicker patch, you can ignore the tweaks from the spinner I was adding and go directly to just moving the setting of the witness votes on the account to the 'accepted' phase).


roadscape:
Thanks for this @eonwarped! Review continues at #2863

PR #2863, opened 2018-06-11
Community - Fix witness voting not resolving properly, add feedback

roadscape:
Replaces PR #2859 by @eonwarped:
> This addresses #1610, while also adding the spinner for in-progress votes.
> 
> For now the logging is still there, but I'll strip it out, the main point is explain what I saw:
> 
> -- prebroadcast Saga it tweaks the account state to remove the witness vote.
> -- broadcast of the transaction , at some point it seems to re-fetch the account state, and often it undoes the state setting in the prebroadcast phase
> 
> So I've moved the state setting to the accepted phase of the transaction saga (post transaction broadcast).
> 
> That was quite an annoying bug.
> 
> (If you want a quicker patch, you can ignore the tweaks from the spinner I was adding and go directly to just moving the setting of the witness votes on the account to the 'accepted' phase).


gl2748:
Note:  in the video included with this issue, the votes disappear on refresh due to SSR. This is not part of this PR. Rather this just adds a 'vote in progress' indicator to clarify when exactly a vote has been successfully submitted.

# before:
vote_before

# after:
voterz

eonwarped:
Oh, I should have included a video , but I did also reproduce the originally referenced issue, where after unvoting, the icon remained checked (without refreshing). You should be able to reproduce as well by testing the unvoting (the problem doesn't seem to happen with voting, or at least not as often.)

eonwarped:
And I also forgot to mention that it affects the write-in votes section as well, both voting and unvoting.

Issue #2719, opened 2018-04-11
The translate@steemit.com "email us" link in the FAQ doesn't work


TimCliff:
The "email us" link should link to `translate@steemit.com`, but it does not.

![image](https://user-images.githubusercontent.com/20735105/38589628-b516c628-3cf1-11e8-8b0a-47eb23241a42.png)

free999enigma:
is that email still in use? i sent an email no reply after 1-2 months...

roadscape:
Solved by removing the `mailto` link.

PR #2865, opened 2018-06-12
fix mail link in FAQ

originated:
Fix for #2719


src/shared/HtmlReady.js
```diff
@@ -144,7 +144,11 @@ function link(state, child) {
         if (state.mutate) {
             // If this link is not relative, http, https, or steem -- add https.
             if (!/^((#)|(\/(?!\/))|(((steem|https?):)?\/\/))/.test(url)) {
-                child.setAttribute('href', 'https://' + url);
+                if (/(mailto:)/.test(url)) {
```
roadscape: Should be a stricter check, otherwise `mailto` anywhere in the URL may conflict with this logic. Can probably just add it to `(steem|https?)`. Also need to verify sanitizeConfig handles `mailto`.

bnchdrff: agreed. needs a `^` in that regex. also please add tests to `HtmlReady.test.js` - ideally can you add a failing test in one commit and then the fix for the test in a commit after that?

bnchdrff: also, is it possible to abuse this somehow? should we validate the contents of mailto links?

Issue #2802, opened 2018-05-17
Implement a NSP replacement.


gl2748:
NSP was acquired by NPM, we should either use the new flavor of NSP - https://docs.npmjs.com/getting-started/running-a-security-audit and integrate it with our CI or look at other services, like Snyk.

PR #2877, opened 2018-06-15
nsp w/ yarn preprocessor

roadscape:
#2802


roadscape:
We should use `yarn audit` instead https://github.com/yarnpkg/yarn/issues/5808

PR #2879, opened 2018-06-15
Community: Advanced Settings for setting payout option

eonwarped:
In addition to defaults set on the profile, allow on posts/comments to set payout options on the fly.

Moves location from float-right to near advanced settings. Here's on a main post edit:

advpayout_post

Here's on a post reply:

advpayout_postreply

Here's on a comment reply: 

advpayout_comment

And here's the modal:

advpayout_modal

The screenshots above were done on the same post and saved the settings independently of each other. Refreshing the page preserves them.

This is a precursor to dropping in beneficiary functionality in #2700 . (You'll see from commit history I added both and stripped out beneficiary here)


gl2748:
PR with complete commit history is in review here:
https://github.com/steemit/condenser/pull/2880

gl2748:
So here is my feedback.
1. great PR, thanks. I like that you split this off from the work you have done to extend the features available to a user creating a new post. Cannot thank you enough.

2. Can we get the advanced settings modal to look like this:
screenshot from 2018-06-15 14-41-14
basically:

- Tweak copy in the title to communicate this only effects the current post.
- Remove word 'Default' from the payout options dropdown, as this is not dynamically updated and may in fact not be the default as configured by the user
- Line that states: ``Current default for your posts: ${user_default_payout_option}`` 
- Link to the `/@username/settings` that reads: "Update defaults in user settings."

A final nice-to-have would be for the payout options dropdown's initially selected value to be the user's default.

3. It looks like that because the trigger for the advanced settings modal is a [string](https://github.com/steemit/condenser/pull/2879/files#diff-4998476ba94a25b3b766df53be6535c6R51) (the formId) it's throwing a proptype warning, because the modal expects a Bool:
screenshot from 2018-06-15 14-35-02

eonwarped:
Done.

gl2748:
@eonwarped thanks! I've added this work to the shadow PR and it's all good to go. Awesome work and excited to get the beneficiaries payout options in next.

PR #2880, opened 2018-06-15
adv setting - payout (@eonwarped)

gl2748:
Inside org version of this PR:
https://github.com/steemit/condenser/pull/2879


src/app/components/modules/Modals.jsx
```diff
@@ -39,6 +41,7 @@ class Modals extends React.Component {
         show_bandwidth_error_modal: PropTypes.bool,
         show_signup_modal: PropTypes.bool,
         show_promote_post_modal: PropTypes.bool,
+        show_post_advanced_settings_modal: PropTypes.string,
```
gl2748: Are we sure this should be a string?

src/app/components/modules/Modals.jsx
```diff
@@ -30,6 +31,7 @@ class Modals extends React.Component {
         show_transfer_modal: false,
         show_confirm_modal: false,
         show_login_modal: false,
+        show_post_advanced_settings_modal: '',
```
gl2748: Why not bool?

gl2748: Ah I see because it is the formId that is saved here.

eonwarped: Ah yes, that's odd, I remember putting in a // formId comment there, but maybe it's somewhere else

gl2748:
Closing this, while the OG PR is still in play.

roadscape:
Looking good, need @pkattera to sign-off

pkattera:
I made some minor copy + CSS changes and also added English strings to the translation files to avoid errors when viewing the new strings from other languages.

roadscape:
Unexpected behavior:

1. In settings, both options set to 'Decline Payout'
2. Go to `/submit.html`, press Reset. "Rewards: Decline Payout" shown as expected.
3. Click 'Advanced Settings', shows `Default: 50% SBD / 50% SP` unexpectedly.

eonwarped:
@roadscape  good catch. I sent a PR to address these.

roadscape:
This is looking great, just waiting on feedback if we want this exposed on comments.

roadscape:
@eonwarped could you please make the following changes:
 - for comments, do not expose advanced settings
 - when a non-default comment payout option is selected in settings, display the current setting and link to Settings page for adjustments. otherwise do not render anything.

We may need to expose these options later (perhaps through a power user option), but the vast majority of the value of the Advanced Settings dialog is on top-level posts. It's important to keep the comments interface as simple as possible for users who have not changed defaults, and the changes requested above should accomplish this.

eonwarped:
Done, linked PR via reference.

roadscape:
LGTM, merged. Thanks for this contribution @eonwarped!

rebuildman:
I made an update on my capacitor, now I get the following error, commenting on any post, which makes it impossible to comment, Transaction broadcast error.
  "RPCError: Assert Exception: beneficiaries.size (): Must specify at least one beneficiary"

benefiarios
 
error
 
How can I fix it?

eonwarped:
@rebuildman This was a much earlier version of the advanced setting module with beneficiary. You can find my newest changes here: https://github.com/steemit/condenser/pull/2700

rebuildman:
@eonwarped  thank you !

PR #2886, opened 2018-06-17
Community - Save comment and root voting bar values independently (#2884)

### LINKED ISSUE #2884: Voting bar no longer saves previous vote strength independently between root posts and comments.
This is a pretty big usability issue for those who like to vote up root posts at 100% and comment posts closer to 5%. Please fix ASAP.

I got confirmation from @TimCliff that it happens to him as well, so I'm not going to include screenshots. The reproducibility steps are obvious (vote on a root post, then a comment, then a root post)

Thanks!

Smittyvb:
Stores comment vote weights with a different localStorage key. Closes #2884.

### DIFF
--- a/src/app/components/elements/Voting.jsx
+++ b/src/app/components/elements/Voting.jsx
@@ -110,7 +110,7 @@ class Voting extends React.Component {
                 weight !== 0
             ) {
                 const saved_weight = localStorage.getItem(
-                    'voteWeight' + (up ? '' : 'Down') + '-' + username
+                    'voteWeight' + (up ? '' : 'Down') + (is_comment ? '-comment' : '') + '-' + username
                 );
                 const castToNegative = up ? 1 : -1;
                 weight = Number(saved_weight) * castToNegative;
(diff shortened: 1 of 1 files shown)


src/app/components/elements/Voting.jsx
```diff
@@ -141,23 +141,23 @@ class Voting extends React.Component {
...
         this.syncSliderWeight = () => {
-            const { username, net_vesting_shares } = this.props;
+            const { username, net_vesting_shares, is_comment } = this.props;
             if (net_vesting_shares > VOTE_WEIGHT_DROPDOWN_THRESHOLD) {
                 const sliderWeightUp = Number(
-                    localStorage.getItem('voteWeight' + '-' + username)
+                    localStorage.getItem('voteWeight' + (is_comment ? '-comment' : '') + '-' + username)
                 );
                 const sliderWeightDown = Number(
-                    localStorage.getItem('voteWeight' + 'Down' + '-' + username)
```
gl2748: We want to keep 'Down' here.

bnchdrff: `Down` is still there - they just got rid of the unnecessary string concatenation. let's use template literals for these though!

roadscape:
Seems to be addressed in https://github.com/steemit/condenser/pull/2887/files

PR #819, opened 2016-12-08
Split Blog tab into Blog and Resteemed tabs

thebitcoiner:
For issue #616

Needs API change from PR https://github.com/steemit/steem/pull/678

Post: https://steemit.com/steemit/@bitcoiner/resteemed-tab-has-been-developed-with-steemd-backend-change

Issue #616, opened 2016-11-13
Split Blog tab into Blog and Resteemed tabs - $50 SBD Bounty Offered (details inside)


TimCliff:
- In the profile section, there should be a "Blog" and "Resteemed" tab.
- The "Blog" tab should only show the posts that the user authored.
- The "Resteemed" tab should show the posts that the user Resteemed.
- There should be a link to both tabs in the profile menu.
- The order on the profile page and the order in the profile menu should be the same.
- Both tabs should display the correct information when viewing data for the current user, and when browsing other users' profiles.
- The code style and format should match with the existing Steemit.com code-base as much as possible.

This feature request will be tied to a developer bounty which will be posted to Steemit.com

TimCliff:
Here is the link to the bounty:
https://steemit.com/bounty/@timcliff/steemit-com-developer-bounty-split-blog-into-blog-and-resteemed-usd50-sbd-steem-sbd-author-payout-from-this-post

thebitcoiner:
I would like to take this challenge!

TimCliff:
Linking to steem issue: https://github.com/steemit/steem/issues/581

TimCliff:
Closing. Based on the comment from @valzav in the PR: _Rather than splitting Blog into two tabs we decided to add some more filters to Blog page so users would be able to filter posts by author (including blog's owner) and by tag._

PR #2897, opened 2018-06-20
Community - Filter Resteemed Posts from Individual User Profile Feed

eonwarped:
Referenced in #616 and #819, I decided to take a look at a possible angle for supporting resteem filtering in a way that would be useful for a good chunk of cases and without relying on backend filtering.

The way it is done is that in the fetch data saga, it will post-filter the data, and fetch the next portion, until the batch size is completely filled, or until an imposed max-batch-limit is reached (currently set to 10, for a total of 200 posts fetched at most for such a run).

Data fetched while in the "hide resteem" mode will essentially be thrown out, and when toggling between the modes, this PR throws out the blogs and starts from scratch again. 

Hmm... as I'm writing this I'm thinking about whether I should even just retain the data that I fetch so that when toggling to and from resteem it doesn't need to start from scratch, and that would simplify the current PR by a lot. Not sure I'm even saving much in terms of memory / GC.... I'll explore that a bit.

Here's where I put the link for now, would need a mock to guide where to place it, if we want it. Most likely needs to be pulled-right or something:

hideresteemlink


eonwarped:
As I'm looking at the other, it sounds like it would be better to even introduce a filter predicate passed to the fetchData saga that can do the same behavior, retain all data, but fetch until the desired amount of filtered items is reached (or max-fetch-batch is hit).

This way it can support generic filtering like tag-based filtering.

Then again, it also seems like it would be obsolete with app-base, but I was itching to have something immediate. I understand if it the changes don't really jive with what's coming.

roadscape:
@eonwarped would you mind adding tests for this feature? Otherwise it's looking good to go.

eonwarped:
Added reducer tests for the new additions.

roadscape:
Thank @eonwarped! Continues in local branch @ https://github.com/steemit/condenser/pull/3011

Issue #1072, opened 2017-01-28
Factor promotion SBD into a post's trending ranking


pfunks:
Post promotion should provide an extra ranking boost not only in the promotion tab but also in any trending ranking (per-tag and overall) on steemit.com. I don't have the numbers but I would guess there are more visitors to the overall trending page than the promoted page. Adding the promotion SBD amount to the post's potential payout for ranking purposes would give a lot of extra value to promoting posts.

For example, [Joseph's post here](https://steemit.com/zdash/@joseph/an-update-on-zdash-a-zcash-alternative) has been promoted with 21.00 SBD.

![image](https://cloud.githubusercontent.com/assets/22420130/22396011/991c06ea-e513-11e6-8c40-f5088b178a34.png)

But it does not rank above posts with more potential payouts that are less than the total promotion SBD+potential payout of his post, in specific tag trending lists or in the main trending list.

![image](https://cloud.githubusercontent.com/assets/22420130/22395982/0ade2f70-e513-11e6-8b62-dcfcc3fac47c.png)

![image](https://cloud.githubusercontent.com/assets/22420130/22396032/28aec0d6-e514-11e6-98c6-a04bca07c444.png)

adcpm:
I agree on this but I think it should be handled by the blockchain, i created a similar issue few months ago https://github.com/steemit/steem/issues/497

iamsmooth:
i dont like the idea 

i think some promoted posts should be displayed on the trending page, but separately identified the way most web sites with paid placement do

Could be 2-3 top-ranked promoted posts shown at the top, or interspersed, but in either case with some sort of "Promoted" tag or box label.

liondani:
when you type something in google you see first the ads 
it could be only top 1-3 promoted  post on trending page not more
so there will be a competition wich one gets in top 3 promoted!
And i would like when someone burns SBD for promotion that the post get a time boost and ends later! I would burn a bunch of SBD for some promoted post before they expire 
(get paid)

clayop:
I think the idea is good but agree with @adcpm. It needs to be blockchain-level feature.

roadscape:
It will most likely not be a blockchain feature (UI concerns generally don't belong in the node). One idea is to have a promoted post at the top of every "page" (i.e. every 20 posts), ranked by promotion cost.

```
promoted #1
posts 1-20
promoted #2
posts 21-40
promoted #3
posts 41-60
```

liondani:
[...]

PR #2908, opened 2018-06-24
Community - Place Promoted Posts within Hot and Trending

eonwarped:
This directly addresses comments in #1072 which suggests that it is desirable to show promoted posts interspersed with trending ones.

Here's a snapshot of what it looks like:

promotedandtrending

The change is fairly small, involving:
1. Fetching of the state of the promoted page in addition to that of trending and hot.
2. Interleaving the promoted posts with the others.
3. Style changes to mark promoted posts.

Added a few saga tests as well. It does appear it was pending some infrastructure changes, but wanted to see what you thought about this.


eonwarped:
Hmm... looks like I will need to add changes to only highlight the selected posts instead of every promoted post.

noisy:
I wrote about this few months ago as a comment to [Steemit Roadmap 2018: Community Input Requested
](https://steemit.com/roadmap2018/@steemitblog/steemit-roadmap-2018-community-input-requested#@noisy/re-steemitblog-steemit-roadmap-2018-community-input-requested-20171116t203908596z)

> ## Facts
> * no one is looking into "promoted tab"
> * "promoting tab" is not effective, therefore people are looking for more effective solutions
> 
> ## Solution?
> * Promoted content should be mixed by default into trending page
>   * fact: artificially promoted content is already mixed into trending, by voting bots, but right now it is much harder to check how many of paid votes some article got
> * starting from 2nd or 3rd spot on trending, every 10-15 entry on trending should be filled by promoted content
> * with only a few spots on trending page, people will get proportional number of displays (impressions) in comparison to amount of SBD spent on promotion
>   * this mean, that if @fknmayhem would spend 50SBD on promotion, and @steevc would spend 100 SBD on promotion, then article of @steevc will appear twice as often on trending page as article of @fknmayhem
> * even with spending just 1 SBD you will have a chance to be displayed on trending page. If you will have some luck and your content will be good, some whale will vote for you!

IMO it is important... that amount spent on promotion would not translate directly to rank on trending page, but to probability of being displayed on the top of trending page.

eonwarped:
Yes, I do like the idea of a probabilistic selection system, but something like that would need backing infrastructure (hivemind?). However, I don't think that this would hamper the current trending landscape. I'm not aware of how the promoted values decay in value though, so not sure how long a post would be stuck there with the current proposal.... In any case, if it's desired, this would allow steemit to put it in easily.

roadscape:
Shelving this for now, post lists will be undergoing significant changes w/ communities rollout, and such a feature could be more easily implemented on the new backend APIs.

PR #2940, opened 2018-07-16
Community - fix bugs concerning payout type in advanced settings

eonwarped:
Fixes two things:
1. Advanced Settings shows correct default chosen.
2. Advanced Settings correctly chooses default based on post or reply. Constant added for post form IDs.


eonwarped:
Addresses bugs in #2880

PR #2941, opened 2018-07-16
if no locally stored vote weight values, prefer default

### LINKED ISSUE #2926: "RPCError: Assert Exception:o.weight != 0: Vote weight cannot be 0." - Can't vote at 100% (w/o changing it before clicking). Can vote 98%.
See gif. My VP was good, 98 worked. 

On a sidenote, the entire slider mechanism is bad UI IMO. You would do good to get rid of it, or change it so it doesn't just pop up when the vote button is clicked. Instead of clicking a post to vote, I  now need anywhere between 2 to 10 mouse operations, considering how fine tuning of the VP is not possible.

gl2748:
closes #2926


src/app/components/elements/Voting.jsx
```diff
@@ -117,7 +117,9 @@ class Voting extends React.Component {
                         (is_comment ? '-comment' : '')
                 );
                 const castToNegative = up ? 1 : -1;
-                weight = Number(saved_weight) * castToNegative;
+                if (saved_weight !== null) {
```
roadscape: Why is saved weight being read on submit? Shouldn't it be read on mount/activation then saved on submit?

gl2748: not sure what difference that would make?

roadscape: I don't understand why it's being read here in addition to `syncSliderWeight` (which gets called on mount and dropdown onshow). The logic looks redundant and wouldn't have it already been read by this point?

PR #2947, opened 2018-07-17
Use steemscript to format canonical URLs

### LINKED ISSUE #336: Proposal: canonical links and json_metadata
If you'd rather I take this to steemit.com for a discussion, I'm more than happy to start up a post. I figured a simple agreement here though may be easy enough. 

I'm proposing a simple change to the way that steemit.com generates it's canonical links, to establish a standard use case of SEO for non-steemit.com websites that are powered by the steem blockchain. The change within this codebase would be simple:

> By default on each comment, steemit.com generates the canonical link meta tag as normal. However, if within the `json_metadata` field the `canonical` property is set, it will use that value in place of the steemit.com domain.

This would become a best practice for anyone developing a website displaying user generated content from the steem blockchain.

For example, I created a post in with the tag #steemws, titled `steem.ws - The public steem api cluster`. By default, when viewing the post on steemit.com, the canonical link included into the content would be:

`https://steemit.com/steemws/@jesta/steem-ws-the-public-steem-api-cluster`

However, if when this post was created, I set the `canonical` field within `json_metadata` to `http://jesta.us`, when viewing the post on steemit.com the canonical link would now be:

`http://jesta.us/steemws/@jesta/steem-ws-the-public-steem-api-cluster`

This will let content creators specify the origin of their content and control a bit of it's SEO. Bots  could also use this to avoid accidentally flagging original content as plagiarism. This concept could also be further expanded to other SEO criteria, but I feel the canonical link is the most important of the bunch, especially with this content technically being available for any website. 

I am not asking for steemit.com to allow for the addition or modification of this data, and by default, all canonical links should default to steemit.com. However, if using an external tool (such as a CMS build on steem for running a custom website) I'd expect the tool to automatically append this data to any posts created from that website. 

Is this a reasonable request? If so, I'll put forth the effort and issue a pull request.

Smittyvb:
Closes #336. Now canonical URLs are formatted according to [steemscript](https://github.com/bonuschain/steemscript) rules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "speakingurl": "9.0.0",
     "sqlite3": "4.0.1",
     "statsd-client": "0.4.2",
+    "steemscript": "1.0.1",
     "store": "1.3.20",
     "style-loader": "0.18.2",
     "svg-inline-loader": "0.8.0",
--- a/src/app/utils/ExtractMeta.js
+++ b/src/app/utils/ExtractMeta.js
@@ -1,6 +1,7 @@
 import extractContent from 'app/utils/ExtractContent';
 import { objAccessor } from 'app/utils/Accessors';
 import normalizeProfile from 'app/utils/NormalizeProfile';
+import Apps from 'steemscript/apps.json';
 
 const site_desc =
     'Steemit is a social media platform where everyone gets paid for creating and curating content. It leverages a robust digital points system (Steem) for digital rewards.';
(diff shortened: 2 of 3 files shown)


roadscape:
I notice there is an entry which will generate non-unique canonical URLs:

```
"zappl": {
    "name": "Zappl",
    "homepage": "https://zappl.com",
    "url_scheme": "https://zappl.com/zappl/a21/lochintar-bar-lal"
  },
```

Also, there is one other approach: allowing an explicit `canonical_url` in `json_metadata`. It seems like an important use case but at the very least we would need to validate that the URL is valid.

Is there potential for abuse with either strategy we need to consider?

roadscape:
@Smittyvb let's use the standard name. However we still need to check if we can allow custom canonical urls.

Smittyvb:
Perhaps this could be merged for now, while you evaluate user-generated canonical URLs? That way, less posts will get incorrectly classified as being made by Steemit.

roadscape:
@Smittyvb if you could clean up `canonicalUrl` generation logic to be a bit more readable, we'll merge the PR. Thanks!

Smittyvb:
Logic should be better now.

Issue #2950, opened 2018-07-20
Avoid Unintentional Edits


inertia186:
_As an author creating a new article, I want to make sure steemit.com in fact creates a new post instead of editing an old post, so I don't accidentally edit a post from 5 months ago._

The `author` named `@silviabeneforti` created a post 5 months ago with the `title`: `Artist at work! ^_^`

Today, the same `author` created a post with the `title`: `Artist at work ^_^`

For both posts, the `permlink` became `artist-at-work-`.  The `author` intended for these two operations to represent two different posts, but to her surprise, the newer post edited the older post.

Original post op: https://steemd.com/tx/1e952370459176338ef7e10e4e0db95ed235c328
Unententional edit op: https://steemd.com/tx/d5f187ad9b7be7fb006d1c228764a9c0c84acb44

PR #2952, opened 2018-07-21
Community - Fix specific permlink collision on create (#2950)

eonwarped:
A potential fix for #2950 by moving the character stripping logic to before the uniqueness check. It appears the parts behind will not add more invalid characters, but may need some auditing.


roadscape:
Thank you @eonwarped! Local PR: #3286

PR #2973, opened 2018-08-04
Community - Add message warning that transactions could be delayed (#2900)

### LINKED ISSUE #2900: Community - Wallet page history is delayed to show transactions
After completing a wallet transaction, it often takes a minute or so for the transaction to show up in the History section. This can be very alarming/confusing for users, and possibly even dangerous - if they are led to believe that their transaction did not go through.

Ideally the wallet history should instantly show the results of any successfully completed transaction.

If this cannot be done, then the next best thing would be to include a warning message in the history that the results may be delayed by X amount of time.

Smittyvb:
Adds Transactions will not show until they are confirmed on the blockchain, which may take a few minutes.
Uses i18n key ``transfer_jsx.transactions_make_take_a_few_minutes``.
Closes #2900.

### DIFF
--- a/src/app/components/modules/UserWallet.jsx
+++ b/src/app/components/modules/UserWallet.jsx
@@ -440,9 +440,7 @@ class UserWallet extends React.Component {
         // set dynamic secondary wallet values
         const sbdInterest = this.props.sbd_interest / 100;
         const sbdMessage = (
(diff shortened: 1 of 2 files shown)


roadscape:
Review continues in local PR #2977.

PR #2989, opened 2018-08-22
filter out certain users in a steem api wrapper

bnchdrff:
closes SD-310

new file `src/app/utils/GDPRUserList.js` stores usernames

404 user-related paths
prevent comments & their children from rendering

### DIFF
--- a/src/app/utils/GDPRUserList.js
+++ b/src/app/utils/GDPRUserList.js
@@ -1,0 +1,7 @@
+const list = `
+bnchdrff
+`
+    .trim()
+    .split('\n');
+
+export default list;
--- a/src/app/utils/GDPRUserList.js
+++ b/src/app/utils/GDPRUserList.js
@@ -1,5 +1,4 @@
 const list = `
-bnchdrff
 `
     .trim()
     .split('\n');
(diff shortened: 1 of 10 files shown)


Smittyvb:
IMO, posts/comments that have replies should still allow those replies to
be shown. Otherwise, users can effectively delete replies to their content,
which is undesirable.
The GDPRed post could show as being by "[deleted]", with a post body of
"[deleted]", when they have replies.

bnchdrff:
@Smittyvb very good point; for now we're going to stick with this method to keep it simple, but that doesn't mean we can't change it down the road.

src/app/utils/GDPRUserList.js
```diff
@@ -0,0 +1,6 @@
+const list = `
```
gl2748: Would be nice to add a comment to indicate that this expects a list separated by new-lines... at least until some names are added?

bnchdrff: this is a convention used in all other user lists in this codebase

src/app/utils/steemApi.js
```diff
@@ -0,0 +1,11 @@
+import { api } from '@steemit/steem-js';
+
+import stateCleaner from 'app/redux/stateCleaner';
+
```
gl2748: It might be nice to wrap `getAccount` in the state cleaner too, such that when a GDPR user attempts to login it also fails ... because I imagine this will be one method a user might check that we have satisfied their request?

PR #2993, opened 2018-08-24
NaN as an array

gl2748:
If it was strictly typed, `formatDecimal` would return a type of `Array` because later on `.join()` is called on it.


gl2748:
@relativityboy can you drop your commits from this branch? I don't think we want to commit this to master.

PR #2995, opened 2018-08-27
Community - Use steemscript to format canonical URLs

### LINKED ISSUE #336: Proposal: canonical links and json_metadata

roadscape:
Replaces #2947 by @Smittyvb:

> Closes #336. Now canonical URLs are formatted according to [steemscript](https://github.com/bonuschain/steemscript) rules.


roadscape:
Steemscript diff shows that v1.0.2 tag on github does not include a version bump commit: 
https://github.com/bonuschain/steemscript/compare/v1.0.0...v1.0.2

`master` reports as 1.0.2 but includes more changes to review:
https://github.com/bonuschain/steemscript/compare/v1.0.0...master

roadscape:
@Smittyvb could you reach out to the maintainers wrt the discrepancy above?

Smittyvb:
It appears the version on NPM is the real 1.0.2. The master branch of steemscript has changes, but they have not yet been released.

roadscape:
@Smittyvb thank you, one last thing -- would you mind putting the logic into its own method and adding a couple simple tests? Example tests can be found in `condenser/src/app/utils/`.

roadscape:
Excellent, thanks for the contrib @Smittyvb!

gansbrest:
I've been reading through a bunch of PRs and it seems like the current solution is only applicable to "apps", meaning platform aggregators that generate content for other users (like Medium for example).

How about regular use case, where I have a blog on my own domain example.com where I publish posts and distribute them via different means (plugins) to other platforms like steemit?

Each post on example.com would have simple blog urls like example.com/my-post-url (note, there are no users or categories) which suppose to be canonical url on external platforms like steemit.

I'm sure there are a ton of blogs like that that would like to distribute their content and use proper canonical links. 

Have you consider those? What's the approach we should take for regular blogs/sites? Should we submit PR for apps.js steemscript for each blog (would get out of hands pretty quickly)?

Please advice.

gansbrest-hbo:
@Smittyvb could you elaborate a bit more please (show example). Is this json_metadata.canonical suppose to be part of post data, or is that common json_metadata field that stores everything.

It seems like it makes sense to pass canonical url along with the post itself in addition to "account" level.

Also when you say set canonical to blog post URL, do you mean specific post or some common base url?

gansbrest:
@Smittyvb could you elaborate a bit more please (show example). Is this json_metadata.canonical suppose to be part of post data, or is that common json_metadata field that stores everything.

It seems like it makes sense to pass canonical url along with the post itself in addition to "account" level, so individual post could overwrite defaults on the account, but I'm not sure how post are currently stored on steemit honestly..

Also when you say set canonical to blog post URL, do you mean specific post or some common base url?

Smittyvb:
@gansbrest

> json_metadata 

Steem posts can have additional data beyond the body and title. For instance, tags. This is called the JSON metadata. You can put whatever data you want there, and it won't usually get displayed. Some things that are stored here:
- Tags
- App used to create the post
- The post format
- Images in the post

This could also store a custom canonical URL for the post. Currently you can't easily set custom values for the JSON metadata. It might be added to Steemit in the future. But if you use the CLI, or use an API, you can set the JSON metadata. 

> "account" level

That's a good idea, but it would require an additional request to load account level data for each post request, which we want to avoid. With unlimited post editing, just setting a canonical URL for each individual post should work.

> Also when you say set canonical to blog post URL, do you mean specific post or some common base url?

Blog post URL.

> show example

So when you make a post on your Example.com blog, you'd post the post on your Example.com blog, with a URL of ``http://example.com/a-post``. Then, you'd post it on the blockchain, with the ``canonical`` key of the JSON metadata being ``http://example.com/a-post``.

gansbrest:
Thanks a lot for the clarifications @Smittyvb 

Biggest question I have at the moment - if I add `canonical: [link]` property to post metadata (somehow, cli or api) would it override default canonical link tag on the Steemit.com website? 

In you previous post you mentioned that it's not supported, by that you meant there is no easy way to access it through UI or it this post `json_metadata.canonical` property is not used for <link> tag generation on steemit?

Could you please provide super simple example of the api / cli call that updates metadata with canonical property if you have time. I'm sure it would be super useful for a lot of people.

talhasch:
the pr you referred doesn't explains clearly usage of this case. lets wait until this type of usage matured.

PR #3012, opened 2018-09-19
set default expiration for market orders to 30 days henceforth, close…

### LINKED ISSUE #3004: Set internal market order expiration to 30 days
With HF20, internal market orders will need to expire after 30 days, otherwise all market orders will be rejected.

gl2748:
Closes #3004


roadscape:
Tested working

PR #3023, opened 2018-09-24
Testnet asset symbols

gl2748:
WIP, Getting condenser operations to work on the testnet where `STEEM -> TESTS` and `SBD -> TBD`.


roadscape:
This is fairly incomplete and could be simplified, closing for now and leaving the branch as reference.

Issue #3026, opened 2018-09-25
The out of RC message is a bit misleading while RC is being recalculated.


lukestokes:
My suggestion is to add something temporary like so:

![image](https://user-images.githubusercontent.com/344748/46029925-59551b80-c0ba-11e8-8b8c-4ac03cd45ea2.png)

People who buy Steem Power thinking it will fix their account may be disappointed to later find out the system was not at equilibrium when they were told to power up.

PR #3027, opened 2018-09-25
Adding a _temporary_ message to let users know the RC system is updating.

lukestokes:
This temporary fix does not go through the normal
language translation approach (en.json, etc) but if there
is interest to do that, we can make that change as well.
The current message is misleading because RC are not
yet being properly calculated.

References #3026


roadscape:
Thank you, continued in #3030

PR #3031, opened 2018-09-25
Hf20 revert warnings

gl2748:
Reverts day-of hf20 changes


therealwolf42:
Why should this be removed completely? I think that's actually quite helpful, especially if there are multiple colours. Red is quite in the eye and should be used for important notifications (as the HF20), but for example right now - it's more of an info notification and I think that's actually quite neat to have something like that.

jredbeard:
Closing in favor of #3044

PR #3032, opened 2018-09-26
Width of blog content & Witness vote removal confirmation modal

quochuy:
Previously, the mismatch in width makes it difficult to know in advance if certain markdown will mess up the layout once the post is created.

Setting PostFull max-width to 52rem to match the width of ".ReplyEditor form".

Also added a confirmation modal upon removing a witness vote on the witnesses page.


quochuy:
Will separate the two changes in different PR

PR #3049, opened 2018-10-03
Community -- Minor tweaks, cleanup and  localization for hide resteem change

eonwarped:
Very small changes to remove an unused case and also set up the text for localization.


roadscape:
Continued in https://github.com/steemit/condenser/pull/3072, thanks!

PR #3052, opened 2018-10-05
Remove banner for HF20 stable

pkattera:


KINGdotNET:
Would be nice to see the "**Announcement Banner**" in a management console for admin, maybe in configuration file where we can easily update.

Issue #3051, opened 2018-10-03
Followed users page displays incorrect follow state when transaction is rejected


ausbitbank:
When I unfollow users too quickly using the buttons [here](https://steemit.com/@ausbitbank/followed) I get denied due to the 1 custom json operation per block rule. 

The error is displayed, the transaction is denied, but the button still updates to "follow", instead of remaining as "unfollow".

I've uploaded a short clip of this behavior here https://youtu.be/41D5jCY2DUA

I assume (but have not tested) that this issue also affects https://steemit.com/@user/followers as well as https://steemit.com/@user/followed

Software Used : Firefox 63.0b11 (64-bit) on Windows 10

pkattera:
@ausbitbank We appreciate the video demonstration.
I can confirm this is a bug for me too on Mac / Chrome.

gryter:
Same if you try to follow (it's the same operation) but goes back to normal after refresh. 

![image](https://user-images.githubusercontent.com/31012685/46507290-cdec3080-c805-11e8-9c9b-582bd8622287.png) 
Pictured is three accounts, error on middle, restored on refresh, other two went through and are accurate.

gadrian78:
I'm not sure if the situations are comparable code-wise, but if they are, would a solution similar to how witness voting page operates be good? I tried to break the witness voting page the same way by unapproving/approving two witnesses within 3 seconds and there was no error and no issue with how the votes were displayed.

Issue #2996, opened 2018-08-29
Follow/Unfollow Feature Issue - Display Of Follow/Unfollow Button Not Working Properly


zorankrstin:
I have submitted an issue regarding follow/unfollow feature and how follow status is displayed on Steemit.

URL: [https://steemit.com/steem/@zorank/follow-feature-sometimes-does-not-work-for](https://steemit.com/steem/@zorank/follow-feature-sometimes-does-not-work-for)

Please pay your attention to post Update 2, where I think I might have found the reason, which is causing this issue.

PR #3054, opened 2018-10-07
Community - Fix Follow Status Mismatch

eonwarped:
Directly addresses  #3051 and possibly will address #2996 as well.

Don't update the following state until the transaction has been confirmed.

Tested locally and appears to resolve the given issues.


roadscape:
Thanks @eonwarped, review continues in #3091

Issue #2968, opened 2018-08-03
Claim rewards frequently mirrors the memo of another transaction


themarkymark-steem:
Claim rewards should never have memos, yet frequently it will grab a memo from another transaction in the wallet.

The wallet page is the most important page on Steemit for 99% of the users. It should be reliable, and trustworthy.

roadscape:
Addressed in #3090 but needs to move to https://github.com/steemit/wallet

Issue #2967, opened 2018-08-03
Transactions show out of order on wallet page


themarkymark-steem:
Transactions should show up in reverse chronological order.  Many times you will find transactions out of order.  Typically a refresh will fix this, **if** the page loads properly.

The wallet page is the most important page on Steemit for 99% of the users.  It should be reliable, and trustworthy.

roadscape:
Addressed in #3090 but needs to move to https://github.com/steemit/wallet

PR #3061, opened 2018-10-14
Community - Fix wallet page so that transfer history is not corrupted

### LINKED ISSUE #2970: Duplicate transactions in the wallet page
Frequently when visiting a wallet page, mine or another user I will see a recent transaction show up, then another copy of the transaction show up a few lines down out of chronological order and as a duplicate similar transaction.  This happens often and typically for newer transactions in my experience.

The wallet page is the most important page on Steemit for 99% of the users. It should be reliable, and trustworthy.

eonwarped:
This should resolve #2970 #2968 #2967.

The problem is that mergeDeep when given two parallel lists will merge each entry by index, and this causes some weird behavior when the two transfer history lists aren't exactly matching up. Instead, the transfer history should be replacing the old transfer history. This change attempts to do just that.

One problem is that I was unable to reproduce the linked errors predictably, so I will try to see if there's a sequence that can reproduce. For now, just wanted to show the proposed fix.

I believe this should only happen in RECEIVE_STATE, and not in the other pieces of code that merge accounts (since they are used on account fetch calls that do not contain transfer history).

### DIFF
--- a/src/app/redux/GlobalReducer.js
+++ b/src/app/redux/GlobalReducer.js
@@ -63,20 +63,9 @@ const transformAccount = account =>
  */
 
 const mergeAccounts = (state, account) => {
-    const accountName = account.get('name');
-    let newState = state.updateIn(['accounts', accountName], Map(), a =>
+    return state.updateIn(['accounts', account.get('name')], Map(), a =>
         a.mergeDeep(account)
     );
-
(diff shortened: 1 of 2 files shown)


roadscape:
Thanks @eonwarped, continues in #3090

PR #3063, opened 2018-10-15
Community - Match width of editor preview and actual post

quochuy:
Changing the width of editor preview to 50rem like PostFull

### DIFF
--- a/src/app/components/cards/PostFull.scss
+++ b/src/app/components/cards/PostFull.scss
@@ -8,7 +8,7 @@
   padding-bottom: 1rem;
   padding-top: 2rem;
   margin: 0 auto;
-  max-width: 50rem;
+  max-width: 52rem;
   @include themify($themes) {
       border-bottom: themed('border');
   }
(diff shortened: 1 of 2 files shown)


roadscape:
@quochuy did you consider changing the editor to 50em instead? Need to check if 52em for PostFull causes any display issues, and constructing the preview is much less invasive

quochuy:
@roadscape no I haven't tried the other way around as I feel that more space is better anyway.

I would even make it wider as 52rem renders to only 832px and I feel that there are too much empty spaces on either sides but that's another issue.

roadscape:
@quochuy it's possible that a lot of posts are formatted to look best at the current 50em width. For this reason I think we should limit the changes to just the editor preview width. If rendered content needs to be wider let's treat it as a separate issue -- there are other concerns such as long lines can be bad for readability.

quochuy:
@roadscape Fair enough, agreed. Will update the PR

quochuy:
PR updated.

roadscape:
Thanks! Moved to local PR #3297

PR #3066, opened 2018-10-18
Update banner to Ned's Steem Needs You post

pkattera:


Smittyvb:
IMO this doesn't really deserve to have a site wide banner. We should reserve these banners to times when the website/blockchain is having issues. Getting on exchanges is good, but it doesn't really deserve a banner like this.

Smittyvb:
Now that Steem has won, this doesn't matter anymore.

PR #3083, opened 2018-11-03
Add Twitch Support to condenser

wehmoen:
With this pull request I want to add support for Twitch.TV livestreams and VODs to condenser.

Links like
`https://twitch.tv/ninja` or `https://www.twitch.tv/ninja`
for channels
and
`https://twitch.tv/videos/330831726`or `https://www.twitch.tv/videos/330831726`

will be converted to an iframe. It will allow more livestreamers to share their streams on steem.

image
image


algo-coder:
Definitely needed !
Would be great if Steemit could merge this ! The sooner the better...
Should bring a whole lot of streamers in here !

skzap:
Very good idea!

cardboard-git:
Absolutely great update.

roadscape:
Thanks @wehmoen, review continues in #3088

Issue #3093, opened 2018-11-13
Community - Integrate Steem Keychain


eonwarped:
Would like to use the Steem Keychain (https://github.com/MattyIce/steem-keychain) in interacting with Steemit.com (and other condensers) as an option. Wanted to open up the discussion to see how this might look and if it would be accepted as a change.

Smittyvb:
Looks like Keychain is missing a few needed operations:
- Create limit order (and create_limit_order2 would be nice)
- Claim reward balance
- Change account authorities (change password)
- Withdraw from savings
- Power up
- Power down
- Profile editing

eonwarped:
Ah I should have mentioned, I was toying with opening up all operations in the keychain with the following PR on steem keychain:  https://github.com/MattyIce/steem-keychain/pull/51

inertia186:
There's no mention of Steem Keychain if the extension hasn't been installed.

![image](https://user-images.githubusercontent.com/494368/58522833-e800c780-8176-11e9-9ed5-978b403ffd7e.png)

(when the extension *is* installed:)

![image](https://user-images.githubusercontent.com/494368/58522878-0070e200-8177-11e9-98ce-c133e5fb1ecd.png)

I think it'd be a good idea to mention that users can install the extension if they want to.

PR #3095, opened 2018-11-14
Community - Integrate with STEEM Keychain

eonwarped:
Addresses #3093.

Preliminary look at how it would work. Note that this requires modifications to the steem keychain, which is proposed in https://github.com/MattyIce/steem-keychain/pull/51

Likely could use some tests, but will wait for preliminary feedback.

One oddity: Steem keychain requires the operations to be JSON-serializable to use the chrome extension message passing feature, which is why some Buffers got removed. They don't seem to be critical, but I might be missing something.

Also missing is the 'feature flag' functionality, but I don't know if that is in use much?


relativityboy:
Without looking @ the code -

* It should have feature flag support. 
* condenser will want some way to validate that the plugin _is well behaved_, since supporting it could be construed as an endorsement of it. RN the only thing that comes to mind is some kind of checksum, version-lock etc.

roadscape:
`steem_keychain` *could* refer to an implementation-agnostic API interface. Are there any other wallet browser extensions?

eonwarped:
There appears to be a "handshake" function you can call to see if the keychain implements it, but I don't really see that adding much in terms of verifying anything.  I wonder if it is possible to guarantee that the extension is well behaved... it seems that other potentially malicious extensions can inject themselves between the condenser and the keychain interface to trick any kind of challenge-response... In the end it will rely on the user installing trustworthy keychains that conforms to the assumed interface.

At the moment I am not aware of any other wallet browser extensions. I was thinking about placing the gates in one utility file just to group it together, but until there are others, I figure it's not strictly necessary. And other wallets can follow the same interface defined by the current steem_keychain, or create an adapter for it. But yes, right now there is an implicit assumption that the window.steem_keychain conforms to that in the linked repo.

Smittyvb:
The absence of a password field on login could be confusing to some users. It would be good if a message appeared where the password field would be, informing users that Keychain was being used.

Smittyvb:
Also: If broadcasting via Keychain fails (as it will with older versions), then Condenser should prompt for a password, instead of popping up an error.

eonwarped:
> The absence of a password field on login could be confusing to some users. It would be good if a message appeared where the password field would be, informing users that Keychain was being used.
> Also: If broadcasting via Keychain fails (as it will with older versions), then Condenser should prompt for a password, instead of popping up an error.

Good ideas, I will look into it. As far as older versions, it may be sufficient to detect the presence of a particular feature (signBuffer, broadcast). Not perfect, but I think workable. Otherwise, I'm not sure how easy it is to do a more general fallback with current structure.

eonwarped:
Now relies on https://github.com/MattyIce/steem-keychain/pull/53/files . Addressed most comments above.

eonwarped:
Better UserSaga diff: (messed up due to indented block) 
[usersagadiff.pdf](https://github.com/steemit/condenser/files/2609538/usersagadiff.pdf)

(mostly formatting diffs until you get to  "const bufSha = hash.sha256(JSON.stringify(challenge, null, 0));".  I should have disabled the auto-formatting right away in the diff probably....

src/app/redux/UserSaga.js
```diff
@@ -407,13 +467,32 @@ function* promptTosAcceptance(username) {
...
-            [api, api.signedCallAsync],
-            'conveyor.get_feature_flags',
-            { account: username },
-            username,
-            posting_private
-        );
+        let flags;
+        if (hasCompatibleKeychain()) {
+            flags = yield new Promise((resolve, reject) => {
+                window.steem_keychain.requestSignedCall(
```
pauliusuza: Just tested this in my browser console.

This is a significant change in UX as it will trigger a popup and interrupt loading instead of making a background call.
Also currently the keychain extension fails to use higher order keys, so if I have "active" key set, this function will fail because it requires "posting" key, with no way to recover it.

eonwarped: Ah, the feature flag fetching is already triggered by a background call earlier (a fork), so I don't believe it is the case that it interrupts loading.

Hmm... as for the higher order key, could be a keychain behavior to allow for active key to sign instead of a posting. But I'm not sure. Would that work?

src/app/redux/UserSaga.js
```diff
@@ -212,186 +228,235 @@ function* usernamePasswordLogin2({
...
-            .toString();
-        private_keys = fromJS({
-            posting_private: isRole('posting', () =>
-                PrivateKey.fromSeed(username + 'posting' + password)
-            ),
-            active_private: isRole('active', () =>
-                PrivateKey.fromSeed(username + 'active' + password)
-            ),
-            memo_private: PrivateKey.fromSeed(username + 'memo' + password),
+    if (!useKeychain) {
```
eonwarped: A better visual diff here:

[usersagadiff.pdf](https://github.com/steemit/condenser/files/2728887/usersagadiff.pdf)

eonwarped:
Added better user saga visual diff to comment in UserSaga file.

Now has ability to toggle between using keychain and not using keychain,.

Screenshots:

When arriving: 
image
 
When toggling checkbox:
image

eonwarped:
For anyone following along, uploadImage flow needs more changes. Just added it, but needs a change to keychain as well, filed in https://github.com/MattyIce/steem-keychain/pull/68

roadscape:
@eonwarped with the wallet split this PR now has conflicts. Could you bring it up to date? We really want to add this feature. Thank you!

src/app/redux/UserSaga.js
eonwarped: Better diff using ?w=1 in query param.

https://github.com/steemit/condenser/pull/3095/files?w=1#diff-9aad451f6da41ec1ad5dc40338a06da2

eonwarped:
@roadscape Now up to date.

One of the diffs is large but is better with the query param ?w=1,  and i've added a link as a comment to help with that.

roadscape:
Local branch: `eonwarped-eonwarped_steemkeychain`

roadscape:
I can sign in using the Keychain but every attempted broadcast results in `There was an error broadcasting this transaction, please try again. DISMISS`. I don't see a call being made to broadcast from condenser. This error message appears to originate from the plugin itself. I'm currently using the latest version.

I'm not sure if voting goes through [background.js#L136](https://github.com/MattyIce/steem-keychain/blob/03f41f5a53b97f1253f3b9724140bbae22e71445/js/background.js#L136) or [background.js#L360](https://github.com/MattyIce/steem-keychain/blob/03f41f5a53b97f1253f3b9724140bbae22e71445/js/background.js#L360).

roadscape:
Looking at the plugin's console I see `RPCError: Unauthorized: Verification failed (Invalid signature)`. That's a much more useful message. It turns out the keys I had saved in the plugin a while back were no longer valid. Now it works 👍

roadscape:
Moved to local PR #3382 👍

PR #3089, opened 2018-11-09
Community: Enable support for start timestamp on YouTube and Vimeo embeds

roadscape:
From @netuoso #3084:

> Closes #3087
> 
> ### Issue
> Condenser does not support timestamps for Youtube/Vimeo embeds
> 
> ### Description
> When a user copies a timestamped link from Youtube/Vimeo, condenser ignores this data because of a simple regex match on the url that does not take timestamp into account. This pull request contains working logic for parsing the timestamp field from Youtube/Vimeo and properly delivering it in the URL. Users simply need to copy/paste a timestamp link to use this feature.
> 
> ### Working Example
> https://condenser.steemliberator.com/music/@netuoso/greydon-square-late-night-munchies-low-technology-youtube
> 
> ### Files Changed
> - `app/components/cards/MarkdownViewer.jsx`
> - `app/components/elements/YoutubePreview.jsx`
> - `app/utils/Links.js`
> - `shared/HtmlReady.js`

PR #3096, opened 2018-11-14
Community - Add support for timestamped embeds

netuoso:
Continuation of https://github.com/steemit/condenser/pull/3089 with twitch support merged in and tested.


netuoso:
Ok for reducing complexity I am pushing a commit to only support `t=seconds` format. Although youtube itself does parse hours/minutes/seconds. I am not sure how common they actually are.

netuoso:
@relativityboy ok reduced support to seconds only for each.

eg youtube: https://www.youtube.com/watch?v=QASbw8_0meM&t=5h
eg vimeo: https://vimeo.com/1084537#t=60s

also added more tests for variations of the youtube link and for the twitch embeds as well.

netuoso:
So with the team slimming down, is there any point for continuing with community PR such as this one? I believe it was fairly well peer-reviewed already and tests were added. It is passing CI and works well on mspsteem.com

Should I close the PR/issue or just let it get orphaned in the backlog?

Just trying to decipher the state of things and determine if this is just noise for you guys.

src/shared/HtmlReady.test.js
```diff
@@ -205,7 +205,52 @@ describe('htmlready', () => {
         const testString =
             '<html><p>before text https://vimeo.com/193628816/ after text</p></html>';
         const htmlified =
-            '<html xmlns="http://www.w3.org/1999/xhtml"><p>before text ~~~ embed:193628816 vimeo ~~~ after text</p></html>';
+            '<html xmlns="http://www.w3.org/1999/xhtml"><p>before text ~~~ embed:193628816 vimeo 0 ~~~ after text</p></html>';
```
roadscape: Start must to be optional, it's important the old tests continue to work as-is

roadscape:
Hi @netuoso -- this is looking solid, main thing that I noticed was that old tests were modified to include start time and that is concerning unless we can be sure there are no side-effects.

netuoso:
@roadscape please recheck

roadscape:
Excellent, thanks! Moved to local PR #3288

PR #3111, opened 2018-12-03
Cap image size to save on bandwidth costs

rjungemann:
Caps blog post image size and should finish the work to dramatically reduce bandwidth costs.

The work has been deployed `imageproxy` side, so we cap images by default, but clients can override the default if they need higher res. However, any blog post image on `condenser` will still be cached at the uncapped size, so this branch makes sure that the URLs are formatted such that they're rendered and cached at the capped size.

This change is easily reversible in case we wish to roll back this behavior later, by simply reverting the commit and re-deploying `condenser`.

### DIFF
--- a/src/app/utils/ProxifyUrl.js
+++ b/src/app/utils/ProxifyUrl.js
@@ -11,6 +11,7 @@
 const rProxyDomain = /^http(s)?:\/\/steemit(dev|stage)?images.com\//g;
 const rProxyDomainsDimensions = /http(s)?:\/\/steemit(dev|stage)?images.com\/([0-9]+x[0-9]+)\//g;
 const NATURAL_SIZE = '0x0/';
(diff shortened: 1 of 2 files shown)


src/app/utils/ProxifyUrl.js
```diff
@@ -11,6 +11,7 @@
 const rProxyDomain = /^http(s)?:\/\/steemit(dev|stage)?images.com\//g;
 const rProxyDomainsDimensions = /http(s)?:\/\/steemit(dev|stage)?images.com\/([0-9]+x[0-9]+)\//g;
 const NATURAL_SIZE = '0x0/';
+const CAPPED_SIZE = '640x0/';
```
rjungemann: `import * as config from 'config';` was giving me trouble. Possibly because of a difference between server-side vs. client-side rendering? Anyone have any tips on that?

PR #3114, opened 2018-12-05
Pinned Posts

rjungemann:


config/default.json
```diff
@@ -75,5 +75,6 @@
   "chain_id": "0000000000000000000000000000000000000000000000000000000000000000",
   "address_prefix": "STM",
   "conveyor_posting_wif": false,
-  "conveyor_username": false
+  "conveyor_username": false,
+  "pinned_posts_url": "https://raw.githubusercontent.com/steemit/condenser-pinned/master/pinned-posts.json"
```
jredbeard: We might want this to default to false - thoughts?

src/server/server-html.jsx
```diff
@@ -40,14 +40,6 @@ export default function ServerHTML({ body, assets, locale, title, meta }) {
                                     content={m.content}
                                 />
                             );
-                        if (m.name && m.content)
```
roadscape: Any idea what's with this? Has it been returning dupe meta's this whole time?

src/app/redux/FetchDataSaga.js
```diff
@@ -39,7 +39,6 @@ export function* fetchState(location_change_action) {
         const username = m[1];
         yield fork(fetchFollowCount, username);
         yield fork(loadFollows, 'getFollowersAsync', username, 'blog');
-        yield fork(loadFollows, 'getFollowingAsync', username, 'blog');
```
roadscape: We need to keep this line

src/app/components/cards/PostsList.jsx
```diff
@@ -147,6 +147,7 @@ class PostsList extends React.Component {
     render() {
         const {
             posts,
+            showPinned,
```
roadscape: `showPinned` is set but never read

src/app/components/cards/PostSummary.jsx
```diff
@@ -383,8 +383,10 @@ export default connect(
         return {
             post,
             content,
-            pending_payout,
-            total_payout,
+            pending_payout: pending_payout
+                ? pending_payout.toString()
+                : pending_payout,
+            total_payout: total_payout ? total_payout.toString() : total_payout,
```
roadscape: What is the purpose of this change? IIRC, both these props, if set, are strings already.

src/server/server.js
```diff
@@ -279,6 +280,10 @@ if (env === 'production') {
 if (env !== 'test') {
     const appRender = require('./app_render');
     app.use(function*() {
+        // Load the pinned posts and store them on the ctx for later use. Since
+        // we're inside a generator, we can't `await` here, so we pass a promise
+        // so `src/server/app_render.jsx` can `await` on it.
+        this.pinnedPostsPromise = pinnedPosts();
```
jredbeard: nice workaround

config/default.json
rjungemann: It'll get overridden by the env variable, but yeah, it's probably safer to not try and show pinned posts at all, if the env variable isn't set. I'll make the change.

config/custom-environment-variables.json
```diff
@@ -70,5 +70,6 @@
   "chain_id": "SDC_CHAIN_ID",
   "address_prefix": "SDC_ADDRESS_PREFIX",
   "conveyor_posting_wif": "CONVEYOR_POSTING_WIF",
-  "conveyor_username": "CONVEYOR_USERNAME"
+  "conveyor_username": "CONVEYOR_USERNAME",
+  "pinned_posts_url": "CONVEYOR_PINNED_POSTS_URL"
```
roadscape: Could we remove `CONVEYOR` from the env var since it doesn't seem to rely on the conveyor service at all?

jredbeard: agree, let's just call it PINNED_POSTS_URL - sg @rjungemann ?

src/app/redux/FetchDataSaga.js
rjungemann: Why do we need to do this twice?

src/server/server-html.jsx
rjungemann: This was removed because we check it twice (why?)

src/app/components/cards/PostSummary.jsx
rjungemann: It depends where the post comes from. The pinned post I get from `steem-js` has these fields as numbers.

config/custom-environment-variables.json
rjungemann: Oops I changed this and I think I rolled it back. Sorry

src/app/redux/FetchDataSaga.js
jredbeard: at a glance I thought it was the same thing also, but it is getFollowers and getFollowing

src/app/components/cards/PostSummary.jsx
roadscape: Gotcha

src/app/redux/FetchDataSaga.js
rjungemann: OH MAN. Thank you. Rolling back the change

src/app/components/cards/PostsList.jsx
rjungemann: Fixed

roadscape:
Tested working

PR #3118, opened 2018-12-06
Add initial Google AdSense code for testing

rjungemann:


Smittyvb:
This doesn't actually put any ads anywhere. Where do we plan to place ads?

jredbeard:
> This doesn't actually put any ads anywhere. Where do we plan to place ads?

This is just an exploratory excercise for testing. The 'where' or even the 'if' is not decided.

src/server/server.js
```diff
@@ -274,7 +282,39 @@ if (env === 'production') {
...
+            yield next;
+        } else {
+            // If user is signed out, enable ads.
+            let policy = this.response.header['content-security-policy']
+                .split(/;\s+/)
+                .map(el => {
+                    if (el.startsWith('script-src')) {
+                        const oldScriptSrc = el.replace(/^script-src/, '');
+                        `script-src 'unsafe-inline' 'unsafe-eval' data: http: https: ${
+                            scriptSrc
```
jredbeard: should this be `${oldScriptSrc}`?

rjungemann: Fixed

PR #3125, opened 2018-12-14
Community - Clarify that pending payout is estimated

### LINKED ISSUE #3115: Add *Estimated* to pending payout, because now it's really off.
From 
image

to
image

Smittyvb:
Closes #3115. Changes "Pending Payout" to "Estimated Pending Payout".


roadscape:
Any way to make it more palatable? The dialog box is already a bit unwieldy:
image

roadscape:
Or what about just "Estimated Payout" instead of "Pending Payout"?

roadscape:
This dropdown is going to need an overhaul if we add much more to it, and for now the original issue does not seem all that relevant because SBD has been much closer to the peg since the last HF. Closing for now.

PR #3127, opened 2018-12-14
Hopefully fix AdSense script tags

rjungemann:


Smittyvb:
We shouldn't have ca-pub-9368037717385698 hard-coded, this should be a config variable.

jredbeard:
@Smittyvb agree - we will fix this in the near future

PR #3130, opened 2018-12-17
Fix active login

rjungemann:
This differentiates initial login and active login, so it will not break the permissions page and the market page.

### DIFF
--- a/src/app/redux/UserSaga.js
+++ b/src/app/redux/UserSaga.js
@@ -140,6 +140,11 @@ function* usernamePasswordLogin2({
     operationType /*high security*/,
     afterLoginRedirectToWelcome,
 }) {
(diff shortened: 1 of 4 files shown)


src/app/redux/UserSaga.js
```diff
@@ -140,6 +140,11 @@ function* usernamePasswordLogin2({
     operationType /*high security*/,
     afterLoginRedirectToWelcome,
 }) {
+    const user = yield select(state => state.user);
+    const loginType = user.get('login_type');
+    const justLoggedIn = loginType === 'basic';
+    console.log('Login type:', loginType, 'Just logged in?', justLoggedIn);
```
roadscape: Can we remove this line?

roadscape:
LGTM, tested working.

PR #3132, opened 2018-12-18
Community - esteem:// protocol support

feruzm:
This will sanitize and allow links with `esteem://`

### DIFF
--- a/src/app/utils/SanitizeConfig.js
+++ b/src/app/utils/SanitizeConfig.js
@@ -89,7 +89,7 @@ export default ({
         // title is only set in the case of an external link warning
         a: ['href', 'rel', 'title'],
     },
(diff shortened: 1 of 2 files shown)


roadscape:
@feruzm is this still relevant? Can eSteem leverage the `steem://` protocol or is there some fundamental different?

feruzm:
@roadscape yes, we have support for both `steem://` and `esteem://`. This PR adds `esteem://` because if user has Vessel and eSteem on same desktop machine it doesn't know what app to open and there is no way to choose default as well. But for mobile devices it is bit different story because user is given option to open with app of their choice.

roadscape:
Hi @feruzm is this still relevant?

roadscape:
Local PR: #3323

PR #3139, opened 2018-12-20
Community - Update power down message for HF20

fbslo:
Edited power down message from "That amount is locked up and not available to power down until the delegation is removed and a full reward period has passed." to "That amount is locked up and not available to power down until the delegation is removed and 5 days has passed.", because full reward period is 7 days and delegations are returned in 5 days since hard fork 20.


fbslo:
Updated by @crokkon - https://github.com/steemit/condenser/pull/3144

roadscape:
Thanks for catching this. Now that these features have moved to the wallet, I've made your change here: https://github.com/steemit/wallet/pull/26

PR #3146, opened 2019-01-01
Community - Update SMT estimated launch year to 2019

Smittyvb:
I'm guessing that a hardfork won't be coming within the next 14 hours, so this updates the estimated SMT launch time in the FAQ from 2018 to 2019.


roadscape:
Local branch at #3273

KINGdotNET:
We are still waiting for SMT launch/release while steem-engine already running.
Why not dropping the SMT and embracing steem-engine produced by the community? We already know steem-engine is working. Just a thought.

roadscape:
Why not both? ;)

PR #3151, opened 2019-01-03
Google AdSense Initial

rjungemann:


src/app/components/pages/PostsIndex.jsx
```diff
@@ -260,7 +261,13 @@ class PostsIndex extends React.Component {
                         )
                     )}
                     <Notices notices={this.props.notices} />
+                    <GoogleAd
+                        name="sidebar-1"
+                        slot="9855000063"
```
Smittyvb: This should be in a config file.

Smittyvb:
I tried this code on my local Condenser instance, but all I see is an empty ``ins`` element instead of an ad. Why?
image

wehmoen:
Google takes a while till it really shows ads on a page.

PR #3156, opened 2019-01-08
Community - Return '451 Unavailable For Legal Reasons' for users on black lists

### LINKED ISSUE #3155: Standardize behavior when handling content or users deemed to be posting illegal content
Currently, when a user is on the GDPRUserList.jsx blocklist, Condenser renders a "Page Not Found" error message. This is misleading because in other places, Condenser responds to a blocked user/content request with a 451 response.

Condenser should standardize its response to blocked/illegal content by utilizing the 451 status code message throughout the app where required. The point of the 451 status code is not to say whether data exists or doesn't exist, just that the requests to it are being denied.

netuoso:
Closes #3155 

This PR intercepts requests made for users in the GDPRUserList.jsx blocklist and responds to requests with a 451 status code instead of a 404 status code.

These changes result in a prevention of any aspect of the page load process. Changes were also made to the PostJson and UserJson parser to prevent returning data for blocked users.

This implementation is live on https://mspsteem.com/ and can be tested by visiting the following links:
`https://mspsteem.com/@thedarkoverlord`
`https://mspsteem.com/thedarkoverlord/@thedarkoverlord/thedarkoverlord-is-now-on-steemit`


jredbeard:
The blocking of the JSON endpoints for things that are on the blacklists is a welcome bug fix, and I don't disagree with using an HTTP status code of 451 for that. The message for the GDPR user list entries must say 'Content unavailable' though instead of what's in the PR. For DMCA the message is currently correct.

We do need a separate ToS users list, but, the behavior should be the same as what the GDPR users list does.

So, in order to accept this, the additions in `server.js` would need to be taken out and the message for the json endpoints would need to be adjusted.

Thanks for your PR @netuoso

roadscape:
@netuoso are you able to make the suggested changes?

netuoso:
@roadscape recheck please

roadscape:
Thank you @netuoso! Local PR: #3298

PR #3157, opened 2019-01-09
Add retrying of steem-js calls

rjungemann:


roadscape:
The [default options](https://www.npmjs.com/package/retry#retrytimeoutsoptions) look reasonable except `minTimeout` is 1000ms. Since this affects page load time we may want to go as low as 50ms. In practice, a 2nd retry is going to be exceedingly rare.

PR #3173, opened 2019-01-15
GDPR 1197

nynhex:


Smittyvb:
This is a list of usernames. Post links should go in the DMCA list.

Smittyvb:
Your first commit was correct, your second commit moved it to the wrong file.

jredbeard:
The correct fix for this is to just have the username in the GDPR list instead of a URL.

nynhex:
@Smittyvb @jredbeard fixed

PR #3215, opened 2019-02-21
Show login warning

rjungemann:


Smittyvb:
So is there going to be a separate condenser instance for the wallet, that only has wallet functionality?

PR #3217, opened 2019-02-22
Fix login error flickering after showing login warning

rjungemann:


rjungemann:
NOTE: Removes some logic that blocks active/owner login, so we should make sure that logic stays

PR #3270, opened 2019-04-16
Update GPT Slot Ids

JonathanPorta:
- Need the right ids in these slots.


JonathanPorta:
There will be more slot updates in the future and we can make these changes at that time.

PR #3266, opened 2019-04-12
Feature: Allow User to Hide a Pinned Post

JonathanPorta:
- Ensure that a user can hide pinned posts.
- Fix a bug causing the pinned post icon to be highlighted constantly.
- Clean up a couple of style issues caught by the linter.

### Shortcomings
The values that determine whether to hide the pinned post and whether to highlight the pinned icon are saved in the browser's local storage. Therefore, these changes do not persist to other devices.

> image
> Small Screen (Expanded) - Dismiss Icon + Highlighted Pin Icon (First Visit)

>image
> Small Screen (Expanded) - Pinned Post Dismissed

>image
> Small Screen (Expanded) - Dismiss Icon + Highlighted Pin Icon (Subsequent Visit)

Replaces https://github.com/steemit/condenser/pull/3265 - Branch Naming Issue

PR #3271, opened 2019-04-16
Fix Server Side Rendering Exception

JonathanPorta:
Fixes a bug introduced via https://github.com/steemit/condenser/pull/3266 where an exception will be thrown when performing a server side render due to `localStorage` not being defined.

This favors the user experience for users that have chosen to hide a pinned post. For users that have not taken action to hide a pinned post they will see the pinned post gently load at the top of the posts lists. @rjungemann and I felt this was a good compromise to make the user experience better for users having chosen to close a post.


roadscape:
Excellent Thanks!

PR #3292, opened 2019-04-26
Community - Fix comment error for authors with special characters

### LINKED ISSUE #3291: Cannot comment in some cases "Invalid permlink character: ."
User reports:

"I'm getting some errors when trying to write a comment, first this one:"

```
RPCError: Assert Exception:false: Invalid permlink character: .
```

"This specific comment I cannot reply to: https://steemit.com/movies/@bil.prag/bil-prag-re-lordnigel-re-themarkymark-what-is-your-favorite-movie-and-why-is-it-top-gun-20190426t023259842z"

The issue is here:

https://github.com/steemit/condenser/blob/acbfd36ef07791a0f206023f7ef56ad09e8c4069/src/app/redux/TransactionSaga.js#L468 

parent author has a period in it. The substitution code was moved in #3286 and did not account for other characters to would be stripped from names.

eonwarped:
In the permlink creation code, restore substitution code at the end as well, as account name can also have invalid permlink characters.

Should resolve #3291

Untested, but will verify shortly.

### DIFF
--- a/src/app/redux/TransactionSaga.js
+++ b/src/app/redux/TransactionSaga.js
@@ -467,6 +467,8 @@ export function* createPermlink(title, author, parent_author, parent_permlink) {
         parent_permlink = parent_permlink.replace(/(-\d{8}t\d{9}z)/g, '');
         permlink = `re-${parent_author}-${parent_permlink}-${timeStr}`;
     }
+    // only letters numbers and dashes shall survive
+    permlink = permlink.toLowerCase().replace(/[^a-z0-9-]+/g, '');
     if (permlink.length > 255) {
         // STEEMIT_MAX_PERMLINK_LENGTH
         permlink = permlink.substring(permlink.length - 255, permlink.length);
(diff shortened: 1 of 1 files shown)


src/app/redux/TransactionSaga.js
```diff
@@ -467,6 +467,8 @@ export function* createPermlink(title, author, parent_author, parent_permlink) {
         parent_permlink = parent_permlink.replace(/(-\d{8}t\d{9}z)/g, '');
         permlink = `re-${parent_author}-${parent_permlink}-${timeStr}`;
     }
+    // only letters numbers and dashes shall survive
+    permlink = permlink.toLowerCase().replace(/[^a-z0-9-]+/g, '');
```
roadscape: If the only issue here is that parent can have a period in the name, how about we remove this line and just replace out any `.` of the `parent_author` in the `else` branch above? This keeps it minimal and explicit.

eonwarped: Yes I can do that. I think I would still remove all of them though. Are there any other characters allowed in the account name? I was thinking dashes, but now I see that dashes are included in the permlink list.

roadscape: Period is the only character allowed in account name that's not allowed in permlink :)

eonwarped: Ok. will change. One moment

eonwarped:
@roadscape  done, thanks!

roadscape:
Thanks! Moved to #3293

PR #3297, opened 2019-04-30
Community - Match width of editor preview and actual post

roadscape:
From @quochuy #3063:
> Changing the width of editor preview to 50rem like PostFull

### DIFF
--- a/src/app/components/elements/ReplyEditor.scss
+++ b/src/app/components/elements/ReplyEditor.scss
@@ -87,7 +87,7 @@
 }
 
 .ReplyEditor .Preview {
-  max-width: 50rem;
+  max-width: 42rem;
   margin-left: auto;
   margin-right: auto;
 }
--- a/src/app/components/elements/ReplyEditor.scss
+++ b/src/app/components/elements/ReplyEditor.scss
@@ -86,6 +86,12 @@
   opacity: 0;
 }
 
+.ReplyEditor .Preview {
+  max-width: 50rem;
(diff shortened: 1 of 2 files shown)


roadscape:
Had to change max-width to `42rem` because PostFull limits to `40rem` and the preview adds `1rem` on each side.

quochuy:
@roadscape I noticed the width has been changed from 50rem to 42rem which is very small as opposed to the 50rem of the PostFull width.

src/app/components/elements/ReplyEditor.scss
```diff
@@ -87,7 +87,7 @@
 }
 
 .ReplyEditor .Preview {
-  max-width: 50rem;
+  max-width: 42rem;
```
quochuy: Why this drastic reduction of the max-width value? PostFull was at 50rem.

quochuy: Unless you wanted to make it 52rem and remove my margins and replace them with a padding of 1em on either side?

quochuy: OK, gotcha.
42.125rem is actually better as it would render to exactly 640px

quochuy: Screen Shot 2019-05-03 at 7 34 27 pm

quochuy: I opened a new PR

PR #3309, opened 2019-05-03
Setting the max-width of editor preview to 42.125rem

quochuy:
Setting the max-width of editor preview to 42.125rem so that it renders to exactly 640px like the post width

Screen Shot 2019-05-03 at 7 35 43 pm


roadscape:
Local PR: #3322

PR #3323, opened 2019-05-08
Community - esteem:// protocol support

roadscape:
From @feruzm #3132:

> This will sanitize and allow links with `esteem://`


src/shared/HtmlReady.js
```diff
@@ -142,8 +142,8 @@ function link(state, child) {
     if (url) {
         state.links.add(url);
         if (state.mutate) {
-            // If this link is not relative, http, https, or steem -- add https.
-            if (!/^((#)|(\/(?!\/))|(((steem|https?):)?\/\/))/.test(url)) {
+            // If this link is not relative, http, https, steem or esteem -- add https.
+            if (!/^((#)|(\/(?!\/))|(((steem|esteem|https?):)?\/\/))/.test(url)) {
```
JonathanPorta: The comment references `http` but the regex doesn't match `http`. Is that intentional?

roadscape: @JonathanPorta `http` will match `/https?/` :)

PR #3329, opened 2019-05-09
Bug: Fix excessive header padding

JonathanPorta:
Introduced by https://github.com/steemit/condenser/pull/3318

Before:
image

After:
image

### DIFF
--- a/src/server/server-html.jsx
+++ b/src/server/server-html.jsx
@@ -251,6 +251,8 @@ export default function ServerHTML({
 
                                 googletag.cmd.push(function() {
                                   googletag.pubads().disableInitialLoad();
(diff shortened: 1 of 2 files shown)


src/server/server-html.jsx
```diff
@@ -251,6 +251,8 @@ export default function ServerHTML({
 
                                 googletag.cmd.push(function() {
                                   googletag.pubads().disableInitialLoad();
+                                  googletag.pubads().setTargeting("edition", ["new-york"]);
```
JonathanPorta: FWIW, this was included with the previous code. I removed it when I was troubleshooting the prebid code.

Issue #3281, opened 2019-04-23
Content near top of screen is sometimes obscured


JonathanPorta:
Observed Behavior:
Some pages load with their top-most content obscured by the header/ad bar.

Expected Behavior:
Any content at the top of a page should be visible below the header/ad bar.

Reproduction:
1. Load a user's profile such as: https://steemit.com/@justinw
2. Perform several non-cache-busting page reloads until the content is obscured (e.g. ctrl/cmd+r)

Additional Info:
1. Performing a cache-busting page reload seems to always result in the proper layout with no obscured content. (e.g. ctrl/cmd+shift+r)

Screenshots:
![image](https://user-images.githubusercontent.com/1451007/56618428-d8261e80-65df-11e9-97fa-c24fd59bba6c.png)

bala41288:
Wondering why this issue was closed. I see that this is still an issue. It is inconsistent but definitely happening.

JonathanPorta:
@bala41288 I closed a version of this issue that I accidentally opened against the wallet repo. Planning on fixing this soon. :)

lukestokes:
Just noticed this on [my own post here](https://steemit.com/dac/@lukestokes/working-with-the-dac-it-s-not-personal).

![image](https://user-images.githubusercontent.com/344748/57094371-68e6b380-6cde-11e9-8ef6-91aa86d63dad.png)

Strangely, it looks okay in private mode:

![image](https://user-images.githubusercontent.com/344748/57094449-8e73bd00-6cde-11e9-9833-2f02a7c861b6.png)

quochuy:
I managed to reproduce this consistently, it happens if you land on any page for the first time. The page loads without the ad banner and when the banner appears it widen the header block without pushing the actual content block down.

Subsequent navigation to another page fixes it as the banner is already rendered but if you reload the page again the issue will occur again.

quochuy:
Here is a fix proposal which needs help with testing as I can't get ads to work on my dev environment:
https://github.com/steemit/condenser/pull/3377

quochuy:
@roadscape has this been deployed to steemit.com ?
The issue is still happening.

roadscape:
@quochuy it has been deployed but there seems to be a lingering issue which appears to be documented in #3385

quochuy:
@roadscape when you have time, could you please test my fix attempt https://github.com/steemit/condenser/pull/3377 or let me know how to setup my local Condenser so that it displays ads.

[...]

PR #3330, opened 2019-05-09
Bug: Fix Sometimes Obscured Content

JonathanPorta:
This is meant to fix https://github.com/steemit/condenser/issues/3281.

I had some trouble reproducing this consistently during this work. When I originally wrote the issue I had no problems reproducing. I think the delayed rendering (and sometimes no rendering) of the banner ad is the root issue. Therefore, it's likely that code introduced in https://github.com/steemit/condenser/pull/3270 likely made this work a little better.

### DIFF
--- a/src/app/components/modules/Header/index.jsx
+++ b/src/app/components/modules/Header/index.jsx
@@ -283,7 +283,18 @@ class Header extends React.Component {
                     )}
                     {/* If announcement is shown, ad will not render unless it's in a parent div! */}
                     <div>
-                        <GptAd type="Basic" slotName="top_nav" />
+                        {this.props.gptEnabled ? (
+                            // Setting a fixed height here allows the content to
+                            // render below the header without the header later
+                            // expanding and blocking the content. GPT hides the ad
+                            // div when no ad shows so this should be ok in that
+                            // situation. If GPT is disabled, this div doesn't render.
+                            <GptAd
+                                type="Basic"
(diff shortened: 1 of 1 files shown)


src/app/components/modules/Header/index.jsx
```diff
@@ -283,7 +283,18 @@ class Header extends React.Component {
                     )}
                     {/* If announcement is shown, ad will not render unless it's in a parent div! */}
                     <div>
-                        <GptAd type="Basic" slotName="top_nav" />
+                        {this.props.gptEnabled ? (
```
JonathanPorta: This comment block details the two changes I made here. I felt it was important to explain this since it is usually not a good practice to use inline styles.

quochuy: Having the height set to the final 90px height fixes the issue, however, if the ad server decides there is no ad to display or if ads are disabled in the Condenser config, we are stuck with a large header and a large empty space.

Around line 34
```this.gptadshown = event => {
            // This makes sure that the sticky header doesn't overlap the welcome splash.
            this.forceUpdate();
        };```
Maybe here, you can add a timeout of few seconds, then check if the DFP iframe has loaded or contains anything? Or maybe the 'slotRenderEnded' from googletag.pubads() might give some details on whether the content is empty or not

JonathanPorta:
Closing in favor of https://github.com/steemit/condenser/pull/3379

PR #3333, opened 2019-05-10
Fixes #3179 - Replaces multiple <br> in post body with a single one

### LINKED ISSUE #3179: collapse excessive whitespace
Even though there are endless ways to abuse whitespace, we should consider handling simple cases like these:
https://steemit.com/actifit/@abusereports/abusereports-re-abusereportsabusereports-re-abusereportsabusereports-re-abusereportsabusereports-re-mitradoactifit-mitrado-20190118t233353131z

quochuy:
Screen Shot 2019-05-10 at 7 39 41 pm

### DIFF
--- a/src/app/components/cards/PostFull.jsx
+++ b/src/app/components/cards/PostFull.jsx
@@ -256,6 +256,7 @@ class PostFull extends React.Component {
             document.title = title + ' — ' + APP_NAME;
 
         let content_body = content.body;
+        content_body = content_body.replace(/(<br\s*\/?>){3,}/gi, '<br>');
         const url = `/${category}/@${author}/${permlink}`;
         const bDMCAStop = DMCAList.includes(url);
         const bIllegalContentUser = userIllegalContent.includes(content.author);


roadscape:
Hi @quochuy, I'm not sure it's a good idea to do such manipulation here outside of HtmlRender or Sanitizer. While it would resolve cases of spammy `<br>`s, I'm not sure how prevalent this is and it does not address `<br />`s nor other vectors such as putting spaces in between. This issue could potentially be mitigated by setting some reasonably high max-height on the content body and having overflow scrollbars. Tempted to close the issue as wontfix for now, unless it's still a common problem. Have you seen it in the wild?

quochuy:
@roadscape

True, I don’t know the condenser in depth yet so missed that. 

I haven’t seen those in a while. Covering all cases would be a bit tricky.

Maybe check the height of a Loren ipsum text that is about the size of a block size and set it as max-height with scroll bars like you said.

roadscape:
Hi @quochuy, let's put this one on the shelf for now. I'll close the original issue with an update. If this becomes an issue again let's look at optional scrollbars. But with better moderation features it will be easy to just hide these comments in the future.

PR #3335, opened 2019-05-10
Fixes #1819 - Overwriting pending payout amount

### LINKED ISSUE #1819: 0.01 post/comment potential payout rounding error
Based on discussions with people who are familiar with the steemd logic, steemd will not pay out post/comment rewards when the potential payout is 0.01. It only starts to pay out at 0.02. This can lead to poor user experience (especially for brand new users) when their first experience with the payout process is a 0.01 post or comment that doesn't get paid after the 7 days. Even though it is only a very small amount of money, for new users this may lead to a bad impression of the site overall, because it feels like it is either broken or cheating them. To better improve the UX, 0.01 pending payments should be shown as 0.00.

quochuy:
Set pending payout to 0 if it is less than the minimum amount of 0.02.

Not sure this is the best way, would better if this minimum amount could be stored in a shared class or config file but I'm not too familiar with React.


roadscape:
What if instead of changing the payout amount, the dropdown displays a notice when `0.02 > pending_payout > 0.00`? E.g. "Note: payout must reach $0.02 for payout"

quochuy:
Need to see if it fits in the space

economicstudio:
@roadscape I like your new idea. As I explained in #1819 just making it zero artificially may lead to another UX problem

ps. I really appreciate @quochuy's many PRs these days!

quochuy:
Screen Shot 2019-06-19 at 3 26 01 pm

src/app/components/elements/Voting.jsx
```diff
@@ -398,7 +398,14 @@ class Voting extends React.Component {
...
         const payoutItems = [];
 
+        const minimumAmountForPayout = 0.02;
+        let warnZeroPayout = '';
+        if (pending_payout > 0 && pending_payout < minimumAmountForPayout) {
+            warnZeroPayout = tt('voting_jsx.must_reached_minimum_payout');
+        }
+
         if (cashout_active) {
+            const payoutDate =  <span>{tt('voting_jsx.payout')} {' '} <TimeAgoWrapper date={cashout_time} /></span>;
```
quochuy: This creates two nested SPAN containers but that's the only way I've found to concatenate "Payout " before the TimeAgo...

quochuy:
@roadscape implemented the suggestion

src/app/components/elements/FormattedAsset.jsx
```diff
@@ -8,6 +8,11 @@ const FormattedAsset = ({ amount, asset, classname }) => {
     if (amount && typeof amount === 'string') {
         amount = parsePayoutAmount(amount);
     }
+
+    const minimumAmountForPayout = 0.02;
+    if (amount < minimumAmountForPayout) {
+      amount = 0.00;
```
roadscape: The posted screenshot shows a pending payout of $0.01 still showing as 1 cent, which seems to contradict the logic here. Can we remove the amount reset?

quochuy: ah yea, will do that

quochuy:
@roadscape done.

Screen Shot 2019-06-20 at 8 28 52 am

I've noticed a little bug. The dropdown on a post uses this from the Theme Light:
font-family: "Source Sans Pro", "Helvetica Neue", Helvetica, Arial, sans-serif;

However on /created it uses this from .article:
font-family: helvetica, sans-serif;

Which is a larger font making the text slightly overflow. Do we want to fix this in this PR?

roadscape:
@quochuy wrt the differing font -- great catch, and yes, ideal to fix in this PR since we're pushing the limits of the dropdown

quochuy:
@roadscape in https://github.com/steemit/condenser/pull/3335/commits/a530f1f315146899ebf5ad0e9a2a5405591eaa2e

I'm proposing to unify the font-family, so we don't end up with too many different fonts.
`helvetica, sans-serif;` is a fallback in  `"Source Sans Pro", "Helvetica Neue", Helvetica, Arial, sans-serif;` that is used by the Theme Light so it makes sense to me unify these.

roadscape:
Perfect, thanks! Local PR: #3412

PR #3359, opened 2019-05-20
Community - fix non-existing post 404

### LINKED ISSUE #3340: non-existing posts do not result in 404
This seems to be new behavior:

https://steemit.com/dontexist/@abc/this-post-doesnt-exist
image

roadscape:
From @economicstudio #3354:

> Fix #3340
> 
> non-existing post 404
> 
> 
> 
> The bug was due to `if (!dis) return null;` not rendering 404 page.
> 
> ps. Eventually, it might be better to be handled by `RootRoute.js` or somewhere centrally, but that may need some substantial change, so I rather fixed the existing `Post.jsx` only.

### DIFF
--- a/src/app/components/pages/Post.jsx
+++ b/src/app/components/pages/Post.jsx
@@ -1,22 +1,19 @@
 import React from 'react';
-// import ReactMarkdown from 'react-markdown';
 import PropTypes from 'prop-types';
 import Comment from 'app/components/cards/Comment';
 import PostFull from 'app/components/cards/PostFull';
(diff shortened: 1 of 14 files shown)


src/app/components/pages/Post.jsx
```diff
@@ -63,7 +65,47 @@ class Post extends React.Component {
         }
         const dis = content.get(post);
```
JonathanPorta: It's kind of gross that this happens in `render`. But I know that was already like that. We should clean it up later.

PR #3369, opened 2019-05-21
Fixes #3362 - Only show reply link to logged in users

### LINKED ISSUE #3362: Unable to cancel a reply on mobile (Brave)
When you try to reply to a post or comment, the cancel button doesn't work at all, the textarea or post/cancel buttons won't disappear. I even tried multiples, none of them would close.

This seems to work fine on desktop using the Brave browser. But on the mobile version it doesn't work at all, even when I turn the shields off. I have also tried in Chrome but it seems to work fine for that on desktop and mobile. Which is strange because Brave is chromium.

quochuy:
… under posts and comments


roadscape:
Local PR: #3399

PR #3376, opened 2019-05-22
Community - Replicate GitHub's image upload behaviour

### LINKED ISSUE #2832: When uploading image, you can't just drop it - text cursor must stay where you want the image to be posted. When it finishes, it jumps to end of text.
This is a very minor nuisance, but a nuisance none the less.
Also it's a meta bug, as I took notice at another less than friendly behavior, where let's say in GitHub where you drop it as where it will be inserted in the editor, no matter where your cursor is the moment it finishes uploading. As soon as you drop it it crates a text string placeholder that gets replaced automatically when the file is online, and doesn't move the cursor, even if at that moment the user is actively typing elsewhere as I am doing right now to demonstrate this behavior:

Here it is on steemitstage:

As can be seen from GitHub it can be improved a lot, but having it go to right after where the image string was posted would be sufficiently friendly IMO. The way it jumps to the bottom of the document is quite off putting.

quochuy:
I can't get the image upload to work on my dev container so can someone test this please?


roadscape:
Seems to be a bug.. no error though, and nothing in console:
Nov-20-2019 10-59-37

quochuy:
will check thanks

roadscape:
Merged to dev. Thanks @quochuy!

PR #3377, opened 2019-05-22
Fixes #3281 - Replacing 'slotRenderEnded' with 'impressionViewable' …

### LINKED ISSUE #3281: Content near top of screen is sometimes obscured

quochuy:
…to resize the header only when the ad is visible.

I can't seem to get ads to work on my dev environment so help testing this would be appreciated.

I wanted to test this solution:
https://stackoverflow.com/questions/35025046/the-exact-moment-of-gpt-viewability-vs-slotrenderended-event-fired

### DIFF
--- a/src/app/components/elements/GptAd.jsx
+++ b/src/app/components/elements/GptAd.jsx
@@ -22,7 +22,7 @@ class GptAd extends Component {
                     googletag.pubads().refresh([slot]);
                     googletag
                         .pubads()
-                        .addEventListener('slotRenderEnded', event => {
+                        .addEventListener('impressionViewable', event => {
                             window.dispatchEvent(new Event('gptadshown'));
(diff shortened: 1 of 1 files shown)


roadscape:
Moved to local PR #3396 for further testing

PR #3382, opened 2019-05-28
Community - Integrate with STEEM Keychain

roadscape:
From @eonwarped #3095:

> Addresses #3093.
> 
> Preliminary look at how it would work. Note that this requires modifications to the steem keychain, which is proposed in https://github.com/MattyIce/steem-keychain/pull/51
> 
> Likely could use some tests, but will wait for preliminary feedback.
> 
> One oddity: Steem keychain requires the operations to be JSON-serializable to use the chrome extension message passing feature, which is why some Buffers got removed. They don't seem to be critical, but I might be missing something.
> 
> Also missing is the 'feature flag' functionality, but I don't know if that is in use much?


JonathanPorta:
Worked for me! :+1:

src/app/redux/UserSaga.js
```diff
@@ -418,13 +499,32 @@ function* promptTosAcceptance(username) {
...
-            { account: username },
-            username,
-            posting_private
-        );
+        let flags;
+        if (!posting_private && hasCompatibleKeychain()) {
+            flags = yield new Promise((resolve, reject) => {
+                window.steem_keychain.requestSignedCall(
+                    username,
+                    'conveyor.get_feature_flags',
```
eonwarped: @roadscape FYI-- here's where feature flag fetch is, and I believe you should see it in testing since the window pops up twice, one for this call.

roadscape:
Merged. Thanks @eonwarped!!

PR #3393, opened 2019-06-05
Fix border issues with comments

quochuy:
- remove depth indicators as not needed now that comments layers are clearly defined
- new comment classes .downvoted and .revealed to have more precise CSS selectors
- hide the voting block from comments when collapsed as the block is too narrow, it breaks the layout. It does not make sense to vote/downvote without reading a comment anyway.
- tweaked and added more precise CSS selectors (for precedence) to fix the border issues with children comments of downvoted comments

### DIFF
--- a/src/app/components/App.scss
+++ b/src/app/components/App.scss
@@ -173,7 +173,7 @@
         }
     }
 }
-.downvoted {
+.downvoted .downvoted {
     opacity: 0.5;
     -webkit-filter: grayscale(1); // image
     filter: gray; // image grayscale
--- a/src/app/components/App.scss
+++ b/src/app/components/App.scss
@@ -173,7 +173,7 @@
         }
     }
 }
-.downvoted .downvoted {
+.downvoted {
     opacity: 0.5;
     -webkit-filter: grayscale(1); // image
     filter: gray; // image grayscale
--- a/src/app/components/cards/Comment.jsx
+++ b/src/app/components/cards/Comment.jsx
@@ -372,18 +372,15 @@ class CommentImpl extends React.Component {
         const commentClasses = ['hentry'];
         commentClasses.push('Comment');
         commentClasses.push(this.props.root ? 'root' : 'reply');
+        if (this.state.collapsed) commentClasses.push('collapsed');
 
+        let innerCommentClass = 'Comment__block';
         if (ignore || gray) {
-            commentClasses.push('downvoted');
+            innerCommentClass += ' downvoted clearfix';
             if (!hide_body) {
-                commentClasses.push('revealed');
+                innerCommentClass += ' revealed';
             }
         }
-        if (hide_body || this.state.collapsed) {
-            commentClasses.push('collapsed');
-        }
-
-        let innerCommentClass = ignore || gray ? 'downvoted clearfix' : '';
         if (this.state.highlight) innerCommentClass += ' highlighted';
 
         //console.log(comment);
--- a/src/app/components/cards/Comment.jsx
+++ b/src/app/components/cards/Comment.jsx
@@ -372,7 +372,16 @@ class CommentImpl extends React.Component {
         const commentClasses = ['hentry'];
         commentClasses.push('Comment');
         commentClasses.push(this.props.root ? 'root' : 'reply');
-        if (hide_body || this.state.collapsed) commentClasses.push('collapsed');
+
+        if (ignore || gray) {
+            commentClasses.push('downvoted');
+            if (!hide_body) {
+                commentClasses.push('revealed');
+            }
+        }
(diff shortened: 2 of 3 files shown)


src/app/components/App.scss
```diff
@@ -173,7 +173,7 @@
         }
     }
 }
-.downvoted {
+.downvoted .downvoted {
```
quochuy: As I've added .downvoted to the .Comment container it will also change the opacity of the children comments. This targets just the downvoted comment not the children.

src/app/components/cards/Comment.jsx
```diff
@@ -372,7 +372,16 @@ class CommentImpl extends React.Component {
...
         commentClasses.push('Comment');
         commentClasses.push(this.props.root ? 'root' : 'reply');
-        if (hide_body || this.state.collapsed) commentClasses.push('collapsed');
+
+        if (ignore || gray) {
+            commentClasses.push('downvoted');
+            if (!hide_body) {
+                commentClasses.push('revealed');
+            }
+        }
```
quochuy: Downvoted comments behave like a collapsed comment but they are not collapsed, as you can see there is a `[-]` link on the right to actually collapse them. 
Adding two new classes for downvoted comments to make it easier to precisely select downvoted/revealed comments.

src/app/components/cards/Comment.scss
```diff
@@ -43,7 +43,10 @@
   }
 }
 
-.Comment.collapsed {
+.Comment.collapsed:not(.downvoted),
+.Comment.collapsed.downvoted.revealed .Comment__body,
+.Comment.downvoted.revealed .Comment.collapsed:not(.downvoted),
+.Comment.downvoted .downvoted {
```
quochuy: There was an issue with CSS selector precedence, so added more precise selectors

src/app/components/cards/Comment.scss
```diff
@@ -56,6 +59,7 @@
     .Voting {
       margin-left: 1rem;
       border-right: none;
+      display: none;
```
quochuy: In long nested comments, the width is too small and the vote/downvote buttons and the count break the layout and make it two lines when collapsed. Does it make sense to vote/downvote a comment without reading it anyway?

src/app/components/cards/Comment.scss
```diff
@@ -64,12 +68,16 @@
     }
   }
   .Comment__body {
-    padding: 0;
-    border: none;
+    @include themify($themes) {
+      padding: 0;
+      border: none;
+    }
   }
```
quochuy: Needed to add the theme here even if it is not really themified to create a more precise CSS selector for precedence over the non-collapsed comments

src/app/components/cards/Comment.scss
```diff
@@ -105,7 +113,8 @@
...
 }
 
-.Comment__body {
+.Comment .Comment__body,
+.Comment.downvoted.revealed:not(.collapsed) .Comment__body {
```
quochuy: Again for precedence

src/app/components/cards/Comment.scss
roadscape: Let's leave it for now and check how to best address it later

src/app/components/App.scss
roadscape: I'm not sure I understand what the intent is -- is this fixing a bug or changing behavior?

src/app/components/cards/Comment.jsx
```diff
@@ -405,25 +414,13 @@ class CommentImpl extends React.Component {
...
-        }
-
         return (
             <div
                 className={commentClasses.join(' ')}
                 id={anchor_link}
                 itemScope
                 itemType="http://schema.org/comment"
             >
-                {depth_indicator}
```
roadscape: Would you please remove the relevant css definitions as well?

src/app/components/App.scss
quochuy: When a comment is downvoted, the DOM looks like this:
Screen Shot 2019-06-06 at 9 46 03 am

The .Comment div has a child div with a class .downvoted which sets its opacity to 0.5.

In order to fix the original issue for this PR, I had to add the .downvoted class to the .Comment div but by doing so, the 0.5 opacity was also applied to all nested comments under it. So making `.downvoted .downvoted` allowed me to target just the first comment, not its children (replies).

quochuy: But I tested again, although it works in most situation, it does not fix the original issue when we have multiple nested downvoted comments.

Will see, if there is another way...

quochuy:
@roadscape PR updated

roadscape:
Excellent, merged!

PR #3390, opened 2019-06-04
Update Slots

JonathanPorta:
These are not your father's ads.

PR #3396, opened 2019-06-07
Community - fix overlapping header

### LINKED ISSUE #3281: Content near top of screen is sometimes obscured

roadscape:
From @quochuy #3377:

> Fixes #3281 - Replacing 'slotRenderEnded' with 'impressionViewable' to resize the header only when the ad is visible.
>
> I can't seem to get ads to work on my dev environment so help testing this would be appreciated.
>
> I wanted to test this solution:
> https://stackoverflow.com/questions/35025046/the-exact-moment-of-gpt-viewability-vs-slotrenderended-event-fired


roadscape:
Hi @quochuy, not sure if you're available right now but the branch `quochuy-header_height_with_ads` is now deployed to https://steemitdev.com for testing.

quochuy:
@roadscape thank you. https://steemitdev.com/ is getting a 503 error.

roadscape:
@quochuy ah the environment goes down for a few hours every day. But I'll get some feedback on #3390 as it may end up resolving the original issue.

roadscape:
Ok, it sounds like this change has already been tried and didn't resolve the issue. The merge of #3390 would require re-evaluating the approach, so let's close this for now. It's possible we may have to ditch the floating header.

PR #3398, opened 2019-06-10
Community - Fix keychain on login.html

scot-nitrous:
Before, when on Login.html, if the username is there and password is blank, clicking on the 'use keychain button' makes the Submit button disabled. This resolves that particular issue.


scot-nitrous:
Oops. I logged in as a different user.  (from @eonwarped)

roadscape:
Local PR:  #3406

PR #3406, opened 2019-06-18
Community - Fix keychain on login.html

roadscape:
From @eonwarped #3398:

> Before, when on Login.html, if the username is there and password is blank, clicking on the 'use keychain button' makes the Submit button disabled. This resolves that particular issue.


roadscape:
🎉 Thanks @eonwarped!

PR #3394, opened 2019-06-06
Community - add border around comment blocks

roadscape:
From @quochuy #3332:

> Fixes #3195
>
> Local screenshot:
> Screen Shot 2019-05-10 at 9 58 15 am

(note, screenshot is slightly outdated)

PR #3408, opened 2019-06-18
Comment separation - Mobile layout

quochuy:
Ref: https://github.com/steemit/condenser/pull/3394

I've fixed the layout on mobile screens.
I've also implemented a suggestion to add a dotted border on the left of each reply block like the old Reddit design, this increase visibility of parent/child connections. Also, since, the border adds to that visibility, we can reduce the 20px left margin to just 10px which gives more space for nested comments.

Mobile:
Screen Shot 2019-06-19 at 8 11 09 am

Desktop (updated):
Screen Shot 2019-06-19 at 8 39 01 am


src/app/components/cards/Comment.scss
```diff
@@ -184,17 +189,35 @@
 
 @media screen and (max-width: 39.9375em) {
   .root {
-    .Comment__header, .Comment__footer, .Comment__body, .Comment__replies {
+    .Comment__block .Comment__header,
+    .Comment__block .Comment__footer,
+    .Comment__block .Comment__body,
+    .Comment__replies {
       margin-left: 0;
```
roadscape: This looks to be just a dupe now

src/app/components/cards/Comment.scss
```diff
@@ -166,6 +170,7 @@
 .Comment__replies {
   margin-top: 1.4rem;
   margin-left: 62px;
+  border-left: 1px dotted #eee;
```
roadscape: The dotted border looks great on mobile, but on desktop the alignment with avatar is awkward. I think on desktop we need to either remove the border, or add padding between border and avatar

src/app/components/cards/Comment.scss
quochuy: @roadscape I wasn't sure if the margin-left needed to also be outside of the  @include themify($themes) block or not in case the theme classes are not added

src/app/components/cards/Comment.scss
quochuy: @roadscape I did added margin around avatar. The screenshot is not up-to-date.
Here is the new one. 

Screen Shot 2019-06-19 at 8 36 44 am

Also themiefied the color

roadscape: 👍 👍

src/app/components/cards/Comment.scss
quochuy: @roadscape I removed them

roadscape:
Thanks @quochuy!

PR #3412, opened 2019-06-21
Community - warn when below $0.02 threshold

### LINKED ISSUE #1819: 0.01 post/comment potential payout rounding error

roadscape:
From @quochuy #3335:

> Fixes #1819
>
> ...
>
>


roadscape:
Thanks @quochuy!

PR #3414, opened 2019-06-26
Override theme colors for modals as they don't support theme yet.

quochuy:
Screen Shot 2019-06-26 at 4 28 22 pm

Maybe at some point we need to apply themes to all modals as white is quite harsh after coming from a dark page.


roadscape:
You're awesome. Local PR: #3415

PR #3417, opened 2019-06-28
community - Allow the use of the json_metadata 'canonical_link' to set canonical links

drov0:
This is a followup of #336. 

Long story short, the current implementation involves a JSON file which does app (from json_metadata) => cannonical link via a list Which works great for dapps like dtube or busy which have a static domain but doesn't work at all for steempress.

If you are unaware about the project, steempress is a wordpress plugin to allow wordpress powered websites to use steem (cross-posting and using steem as a comment section)

So basically this logic is added : add a parameter in the json_metadata "canonical_link" and if it's set, use that url as canonical link, otherwise fallback to the app logic.

### DIFF
--- a/src/app/utils/CanonicalLinker.js
+++ b/src/app/utils/CanonicalLinker.js
@@ -1,30 +1,40 @@
 import Apps from 'steemscript/apps.json';
 
 export function makeCanonicalLink(d) {
-    var canonicalUrl = 'https://steemit.com' + d.link;
-    if (
-        d.json_metadata &&
-        d.json_metadata.app &&
-        typeof d.json_metadata.app !== 'string'
-    ) {
-        return canonicalUrl;
-    }
-    const hasAppTemplateData =
-        d.json_metadata &&
-        d.json_metadata.app &&
-        d.category &&
-        d.json_metadata.app.split('/').length === 2;
-    if (hasAppTemplateData) {
-        const app = d.json_metadata.app.split('/')[0];
-        const hasAppData = Apps[app] && Apps[app].url_scheme;
-        if (hasAppData) {
-            canonicalUrl = Apps[app].url_scheme
-                .split('{category}')
-                .join(d.category)
-                .split('{username}')
-                .join(d.author)
-                .split('{permlink}')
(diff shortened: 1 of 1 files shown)


reggaemuffin:
That is a much needed feature for TokenBB too, as we also allow custom domains for different forums.

src/app/utils/CanonicalLinker.js
```diff
@@ -1,30 +1,40 @@
...
-                .join(d.category)
-                .split('{username}')
-                .join(d.author)
-                .split('{permlink}')
-                .join(d.permlink);
+    let canonicalUrl = 'https://steemit.com' + d.link;
+
+    if (d.json_metadata) {
+        if (
+            d.json_metadata.canonical_link &&
```
roadscape: Could we use `canonical_url` instead of `canonical_link`? In this context `link` may imply an object consisting of URL, title, and optional attributes.

src/app/utils/CanonicalLinker.js
```diff
@@ -1,30 +1,40 @@
...
+            typeof d.json_metadata.canonical_link === 'string'
+        ) {
+            const urlTester = new RegExp(
+                /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$/gm
```
roadscape: The regex on this line seems to make the protocol optional and it's something that needs to be explicit. In some other places we use a simpler regex, e.g. `/^https?:\/\//`.

drov0: Good point I'll use that.

src/app/utils/CanonicalLinker.js
drov0: Ah I wasn't aware of that, I used the general convention regarding canonical links as "canonical_url" sounded a bit odd. But alright I'll switch to that.

drov0:
Link to a post to test the feature http://localhost:8080/test/@petanque/test-canonical-link1

roadscape:
Local PR: #3435

PR #3421, opened 2019-07-03
Community - Prevent phishing dialog

quochuy:
- Update the phishing domain blacklist with known domains from Steem Cleaners.
- On hover on a comment, darken its background to reveal images that have white background color.
- Like other social platforms, open external links to an intermediate page to warn the user.

Screen Shot 2019-07-03 at 2 46 01 pm


eonwarped:
Why does this need to be a route? Vs a modal, say.

src/app/locales/fr.json
```diff
@@ -686,5 +688,18 @@
...
+        "phishylink_caution": "Lien caché; tentative de phishing",
+        "phishylink_reveal": "affichier le lien"
+    },
+    "externallink_jsx": {
+        "about_to_leave_steemit": "Vous allez quitter Steemit",
+        "the_link_you_clicked_is_external":
+            "Le lien que vous venez de cliquer va vous emmener vers le site suivant",
+        "we_just_want_to_verify_to_continue":
+            "Nous voulons juste vérifier que vous voulez bien continuer.",
+        "go_back": "Go back",
```
Smittyvb: This should be in French, not English

src/app/components/all.scss
```diff
@@ -58,3 +58,4 @@
 @import "./pages/TagsIndex";
 @import "./pages/Welcome";
 @import "./pages/About";
+@import "./pages/ExternalLink";
```
Smittyvb: File should end in an empty line

src/app/locales/fr.json
quochuy: Ohh I missed the translation! Thanks

quochuy:
> Why does this need to be a route? Vs a modal, say.

@eonwarped 
It's not necessary to be a route indeed. Do you see any disadvantage of using a route?
Also, I'm not very familiar with how the Condenser use generates a new modal, will need to dig more into this.

eonwarped:
Having its own route imo is awkward just from a REST resource point of view, and there's certainly no reason for it to have to make an additional call to steemd (or is it magically handling a '/' route as a "don't actually call Steem API"?)

Hmm the Modal route is a bit clunkier but that could also be remedied. This sort of behavior is much like other warning modals that pop up, like for Transfers, etc.

quochuy:
@eonwarped 
Yea, apparently the Condenser calls getState for every single route change.
Will have to see how I can use Modals instead.

quochuy:
@eonwarped I switched to using a modal:
Screen Shot 2019-07-04 at 4 46 50 pm

quochuy:
As for the comment hover effect.
Mouse out:
Screen Shot 2019-07-04 at 4 52 46 pm

Mouse hovering the comment body:
Screen Shot 2019-07-04 at 4 52 53 pm

eonwarped:
Nice! And looks like the dialog was pretty easy after all.

quochuy:
Yes it was, if you know where to look at. It was a challenge as there was several way to do it and I was looking at the wrong place at the beginning. Then I discovered the one for resteeming.

roadscape:
Excellent work! Local branch for testing: `quochuy-prevent_phishing`

roadscape:
Looking great so far, but the comment highlighting needs appropriate color in night mode. Waiting on https://github.com/steemit/condenser/pull/3434 to be reviewed to wrap up testing.

quochuy:
@roadscape cool, I'll have a look at the comment highlighting in night mode later on today

quochuy:
@roadscape got few minutes free time so there it is

roadscape:
Local PR: #3421

quochuy:
Correction: local PR is https://github.com/steemit/condenser/pull/3440

PR #3433, opened 2019-07-15
Closes #3005: don't hide own comments even if flagged

### LINKED ISSUE #3005: My own comments shouldn't be hidden
If a certain user is logged in, their own comments should never be greyed out/hidden.

quochuy:


roadscape:
Local PR: #3437

PR #3435, opened 2019-07-16
Community - json_metadata canonical_url support

roadscape:
From @drov0 #3417:

> This is a followup of #336. 
> 
> Long story short, the current implementation involves a JSON file which does app (from json_metadata) => cannonical link via a list Which works great for dapps like dtube or busy which have a static domain but doesn't work at all for steempress.
> 
> If you are unaware about the project, steempress is a wordpress plugin to allow wordpress powered websites to use steem (cross-posting and using steem as a comment section)
> 
> So basically this logic is added : add a parameter in the json_metadata "canonical_link" and if it's set, use that url as canonical link, otherwise fallback to the app logic.


roadscape:
Tested, works great

roadscape:
Thanks @drov0!

PR #3437, opened 2019-07-16
Community - don't hide own flagged comments

### LINKED ISSUE #3005: My own comments shouldn't be hidden

roadscape:
From @quochuy #3433:

> Closes #3005: don't hide own comments even if flagged


roadscape:
Thanks @quochuy!

PR #3461, opened 2019-08-14
3453 get state modifications

gl2748:
Adds condition to the fetchState saga, to not make a getState call on those paths that contain '.html', i.e. static pages / routes , that don't rely on dynamic data from the blockchain.

### DIFF
--- a/src/app/redux/FetchDataSaga.js
+++ b/src/app/redux/FetchDataSaga.js
@@ -35,7 +35,6 @@ let is_initial_state = true;
 export function* fetchState(location_change_action) {
     const { pathname } = location_change_action.payload;
     const m = pathname.match(/^\/@([a-z0-9\.-]+)/);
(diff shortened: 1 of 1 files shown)


gl2748:
Worth noting that this could be extended using the routes detailed here:
https://github.com/steemit/hivemind/commit/71e8893fa62ca88a5b8e81f8febaab9d29be1d10

i.e. drop unnecessary / duplicate getState requests for account tabbing etc... @roadscape what do you reckon?

roadscape:
This logic needs to live in `steemApi.js`'s ``getStateAsync`` because `src/app/redux/FetchDataSaga.js` is not the only caller. For instance, with this PR, `/submit.html` still calls `get_state /submit.html` (most likely from `src/shared/UniversalRender.jsx`). And it will probably need to return a dummy state object because condenser may not be ready to handle the `global` key missing or incomplete.

Issue #3462, opened 2019-08-16
get_state observer param


roadscape:
Basing off the `bridge-api` branch, `getStateAsync(url)` in `app/utils/steemApi.js` needs to take an additional parameter, `observer`, and pass it through to `callBridge`.

Every caller of `getStateAsync` will need to provide the currently-logged-in user's name (e.g. `test-safari`) as the value for this param.

 - if a user is not logged in, pass `null`
 - ensure the logic works fine with SSR

roadscape:
merged #3465

PR #3465, opened 2019-08-16
3462 im get state async observer param, refs #3462

gl2748:


gl2748:
Screen Shot 2019-08-16 at 8 33 19 AM

PR #3483, opened 2019-08-30
Closes #3480 - Don't two ads in the comments when there are less than 5

### LINKED ISSUE #3480: When less than 5 comments only display 1 ad in comments section
Current behavior with only one comment:

image (3)

quochuy:


src/app/components/pages/Post.jsx
```diff
@@ -277,7 +277,7 @@ class Post extends React.Component {
                         </div>
                     </div>
                 </div>
-                {this.props.gptEnabled ? (
+                {this.props.gptEnabled && commentCount >= 5 ? (
```
JonathanPorta: @quochuy Would you mind moving this check to line 252? This will hide the ad on top of the comments when there are less than 5 comments instead of the ad at the bottom of the comments. I am asking for this change because the bottom ad makes more money than the top ad so we would be better off showing the ad that makes more money. :stuck_out_tongue_winking_eye:

quochuy: OK no problem.

quochuy: Done

roadscape:
Local PR: #3490

PR #3490, opened 2019-09-05
Community - hide ad when less than 5 comments

### LINKED ISSUE #3480: When less than 5 comments only display 1 ad in comments section

roadscape:
From @quochuy #3483:

> Closes #3480 - Don't two ads in the comments when there are less than 5


roadscape:
Thanks @quochuy!

PR #3491, opened 2019-09-05
Add BSA Tracking Script

JonathanPorta:
- Allows for realtime ad performance monitoring.

image

### DIFF
--- a/src/server/server-html.jsx
+++ b/src/server/server-html.jsx
@@ -191,6 +191,12 @@ export default function ServerHTML({
                         }}
                     />
                 ) : null}
+                {gptEnabled ? (
+                    <script
+                        src="//m.servedby-buysellads.com/monetization.js"
+                        type="text/javascript"
+                    />
+                ) : null}
                 {shouldSeeCookieConsent ? (
                     <script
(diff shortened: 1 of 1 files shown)


bobinson:
@JonathanPorta  - Can you also look into header bidding ? Enabling header bidding is the best possible means to improve monetization.

jredbeard:
@bobinson we already do this through our current advertising partner (bidding between multiple ad networks)... prior to that we were doing it with GPT and prebid.js

src/server/server-html.jsx
```diff
@@ -212,6 +218,19 @@ export default function ServerHTML({
                 {assets.script.map((href, idx) => (
                     <script key={idx} src={href} />
                 ))}
+                {gptEnabled ? (
+                    <script
```
roadscape: async=false?

JonathanPorta: The browser loads scripts synchronously unless you specify `async=true`. BSA was telling me not to add that tag.

Info here if you want to read: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

PR #3515, opened 2019-09-13
disable app-based canon links

roadscape:


Smittyvb:
What  exactly is the issue? Are some apps using a different scheme than apps.json? It seems like we can keep ``json_metadata.canonical_url`` in that case. Also if the app couldn't be detected no canonical URI to would be better than falsely claiming Steemit.

PR #3520, opened 2019-09-16
Community - show beneficiary info

### LINKED ISSUE #3233: Show beneficiary info on the pending payout
Now there are more and more apps that use beneficiary. Notable one is the recent drugwars that charges 35% for the shared battle posting.

Obviously 35% of that will actually be a loss for most users even if they get some "free" voting from fundition account. But this isn't clearly shown on condenser. While there are some alternative UIs to check this information, the official condenser should also have that feature.

roadscape:
From @economicstudio #3310:

> Close #3233 
> 
> Previously, Steemit.com doesn't show any beneficiary information, which may surprise users, especially when beneficiary percentage is high.
> 
> 
> > Now it shows the beneficiary info
> 
> Of course, this PR works well with multiple beneficiares
> 
> > Also works well with multiple beneficiaries.
> 
> 
> I also thought about showing some word like "beneficiary" or so in the pop-up, but I believe that users are already familiar with that from other UIs, so I sticked to the simplest form. For instance, Busy shows as follows:
> 
> 
> > beneficiary info on Busy.org
> 
>


jredbeard:
thanks @economicstudio

PR #3522, opened 2019-09-17
3509 author dropdown

quochuy:


roadscape:
This works really well! Only thing left is AuthorDropdown needs to use `normalizeProfile`.

roadscape:
Minor changes: https://github.com/steemit/condenser/commit/f5233262aa0d478f4d23328ebd527f427c9561e4

Thanks @quochuy!

PR #3552, opened 2019-10-02
Initial video ad support with streamrail

### LINKED ISSUE #3550: Add an initial video ad to condenser
This can be based on an initial (working) test branch here: https://github.com/steemit/condenser/tree/test-video-ad
And this specific commit: https://github.com/steemit/condenser/commit/d764b277559db53c41c94b668dfb8c517944495c

Acceptance criteria:

- [ ] Make a reusable react component to display the div for video ads (similar to existing GPTAd component)
- [ ] Add a CSS file for it
- [ ] Have the API key and Player ID set via a config object
- [ ] Have ad inserted once after the 5th post in a post list
- [ ] Ability to turn on/off video ads with a config var (similar to the existing `shouldDisplayAds` for the existing GPT ads)
- [ ] Remove the CSP changes that were added to the test branch for testing so they can be added in our dev/production environments with the existing environment variables
- [ ] Ads should refresh when navigating to different pages. This may be able to be handled with the StreamRail API, but, it also could be handled other ways just in condenser itself.

Ads will only display on desktop (set in StreamRail console), so, at this point, no need to make them work well with mobile.

The StreamRail API documentation is here: https://partners-api-docs.streamrail.com/

jredbeard:
Closes #3550 

This is a work in process because we are not setup yet with an account with them. It is currently setup to run the demo ad from streamrail.

What this does: when video ads are enabled via env var, a single video ad will display after the 5th post in a post list.

In addition, some unknowns that we'll need to know to finish this:
1. I believe the 'player ID' works like 'tags' with GPT ads, but, we will need to confirm this.
2. We need to know if videos from streamrail will always load from the same host.
3. We need to know if streamrail will have video ads available that fit the same size we use for post lists. If not, what sizes are available and how can we make it look good with those sizes?
4. Is there a function available to refresh ads? If not, we'll have to do something hacky to make it work properly when navigating between posts/post lists.


jredbeard:
Closing, this was usurped by #3634

PR #3555, opened 2019-10-05
Community - Update FAQ

TimCliff:
This PR removes and updates outdated and incorrect information from the condenser FAQ page - mostly related to the latest hardfork. It would be beneficial to do a larger change to add information on all the new stuff that has happened, but at least for now - this at least makes what is there accurate.


roadscape:
Local PR: #3562

PR #3563, opened 2019-10-11
Community - fix clear draft on posting

roadscape:
From @eonwarped #3557:

> Under certain conditions the draft does not clear when posting is done, which can give the impression that a post is being edited if one is not careful.


roadscape:
Thanks @eonwarped!

PR #3574, opened 2019-10-19
Fix blank space at the bottom of a post on nightmode

jredbeard:
When viewing a post in darkmode, the bottom of the page contains whitespace that is not the same background color as darkmode. Adding a 1px `padding-bottom` to the body makes the browser calculate where the body actually ends so it will use the background color all the way to the bottom.

There could be other ways to fix this bug.

Before:

Screen Shot 2019-10-19 at 3 24 44 PM

After:

Screen Shot 2019-10-19 at 3 06 04 PM

### DIFF
--- a/src/app/assets/stylesheets/_themes.scss
+++ b/src/app/assets/stylesheets/_themes.scss
@@ -118,7 +118,8 @@ $themes: (
 /*
  * Implementation of themes
  */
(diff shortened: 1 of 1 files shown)


jredbeard:
Only works on a fresh load the page and not when navigating to one from a posts list - not the proper fix for now

Issue #3575, opened 2019-10-21
ReplyEditor - hide primary category


roadscape:
When posting into a community, ReplyEditor should hide the `hive-xxx` primary tag. It needs to be stored in component state. Ensure that draft save/load functionality is compatible.

roadscape:
#3586 merged

PR #3586, opened 2019-10-22
reply editor community tweaks, refs #3575

gl2748:
- Hides the community account name from the 'Posting to...' area.
- Hides the community account name from appearing as a tag.

Screen Shot 2019-10-22 at 5 10 25 PM


roadscape:
Draft save/load doesn't seem to work -- if you start a draft in a community and navigate away, then go back to editor, the community will be cleared

PR #3595, opened 2019-10-29
Update maximum tag count label in post editor from 5 to 8

ausbitbank:
The tag limit was recently modified, this label was not


roadscape:
Rolled into #3606

PR #3597, opened 2019-10-30
Community - only reset form for non-edit flow

eonwarped:
Upon successful posting, only clear form for non edit flow.


roadscape:
Rolled into #3606

PR #3602, opened 2019-11-14
Little tweaks to sidebar menus

quochuy:


roadscape:
The concept is cool but the sidebars are in a state of flux at the moment.. closing for now, will revisit.

PR #3607, opened 2019-11-14
profile settings

### LINKED ISSUE #3339: setting tab should move back to condenser
I really appreciate this change: https://github.com/steemit/steem/issues/3274

But as you know, now the setting tab should move back to condenser.

Many newbies are actually quite confused as to how they can change profiles. Previously, it requires the active key, which isn't a good idea, but now it only requires the posting key as desired.

So it's time to move the setting tab back to condenser :)

roadscape:
Resolves #3339


roadscape:
Form is now reading and saving `posting_json_metadata` 🎉

roadscape:
TODO: UI nudge when profile empty

PR #3628, opened 2019-11-26
3620 im community language select

### LINKED ISSUE #3620: comms language select
Community settings need a `ISO 639-1` language select field. `CommunitySettings` already has a basic native `<select>` (commented out) and all the related load/save logic is working. However, a native `<select>` is not user friendly for such a large number of options.

We need a form component which functions just as well but provides auto-complete UI and returns a valid `ISO 639-1` code as the value.

gl2748:
Closes #3620 

Screen Shot 2019-11-26 at 2 32 08 PM


gl2748:
Known todos:
- Form submit should validate.
- Default value should be current language setting for community.

PR #3629, opened 2019-11-27
Fix broken styling on the post preview before an ad in the post list

quochuy:
Returning an array allows you to return multiple components without wrapping them inside a <div>


roadscape:
Thanks @quochuy!

PR #3632, opened 2019-12-02
Community - Fix minor error on beneficiary setting error case

eonwarped:
Remove clause that likely never worked. Tested locally to make sure input errors are shown properly as before.


roadscape:
Thanks! Local PR: #3633

PR #3634, opened 2019-12-02
Streamrail Video AD through BSA's optimize

jredbeard:
This adds a video ad after the 5th post in a post list when turned on via environment variable `VIDEO_ADS_ENABLED`. It adds *.streamrail.com as a default source for media streaming. I added this because we just use the default for media sources and do not set them via environment variable as they don't change very often.

These three hosts are required to be added to the `connect-src` CSP directives, but can be added via environment variable: `sdk.streamrail.com api.vidiom.net *.streamrail.net`.

Screen Shot 2019-12-02 at 2 42 08 PM

### DIFF
--- a/src/app/components/cards/PostsList.jsx
+++ b/src/app/components/cards/PostsList.jsx
@@ -281,7 +281,7 @@ class PostsList extends React.Component {
                                 />
                             </li>
 
-                            <div className="articles__content-block--ad text-center">
+                            <div className="articles__content-block--ad video-ad">
                                 <VideoAd id="bsa-zone_1572296522077-3_123456" />
                             </div>
                         </div>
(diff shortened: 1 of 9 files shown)


src/app/components/cards/PostsList.jsx
```diff
@@ -267,6 +268,27 @@ class PostsList extends React.Component {
         const renderSummary = items =>
             items.map((item, i) => {
                 const every = this.props.adSlots.in_feed_1.every;
+                if (this.props.videoAdsEnabled && i === 4) {
+                    return (
```
roadscape: `renderSummary` logic is not very clear; this case needs to be merged into the lower section (where ad is appended to post), otherwise the post at `i==4` will never be rendered

gl2748: We can wrap this in `<center>` if we want it to be centered ...

jredbeard: moved this to an `else if` below - just in case `every` is ever a 4, this is the better way to handle it (because we wouldn't display both a video ad and a regular ad in the same slot)

jredbeard: @roadscape think it would look better centered? I am undecided on that - it depends on a lot on the viewport size that the user has I suppose

PR #3635, opened 2019-12-04
Fixes #3385

quochuy:
Using MutationObserver to detect when the header ad is rendered and resize the headroom.
This should provide a more permanent fix.

PR #3636, opened 2019-12-04
Community - fix ad hiding header

roadscape:
From @quochuy #3635:

> Using MutationObserver to detect when the header ad is rendered and resize the headroom.
This should provide a more permanent fix.


roadscape:
There is an issue w/ SSR - `MutationObserver` only available in browser

gl2748:
## Mobile
Screen Shot 2019-12-17 at 7 53 21 AM

## Desktop
Screen Shot 2019-12-17 at 7 48 18 AM

gl2748:
# before
67FF30E4-5168-4375-BB07-1BF72385CA43
# after
7EEAFD66-92C7-43CF-9C5D-4B2091F36B73

PR #3638, opened 2019-12-06
Fixes to the multiple image uploads

quochuy:
- Fix: If you drop you images by multiple batches, the index number in  temporary didn't increment
- Fix: the latest batch of files replaces the current ones


roadscape:
Thanks @quochuy!

PR #3639, opened 2019-12-06
3627 im unread notifications

### LINKED ISSUE #3627: unread notifications
1. Condenser needs to query for unread notifications and show the unread count on the Header component, in the corner of the logged-in-user's avatar. API: `bridge.unread_notifications(account)`.
2. The notifications page needs a "Mark as read" button which submits a custom_json op with id `notify` and payload `["setLastRead", {"date":"YYYY-MM-DD.."}]`.

gl2748:
closes #3627 
Screen Shot 2019-12-06 at 9 01 35 AM
Screen Shot 2019-12-06 at 9 01 24 AM

TODO:
- [ ] Style Notification 'Dot', e.g. 'Does it need to display the count?, how to handle high figure counts.'
- [ ] Add rules for poll failure, e.g. 'Attempt x times'.
- [ ] Consider hooking into browser notifications.

### DIFF
--- a/src/app/components/pages/UserProfile.jsx
+++ b/src/app/components/pages/UserProfile.jsx
@@ -137,6 +137,7 @@ export default class UserProfile extends React.Component {
                 posts,
                 profile,
                 notifications,
+                isNotificationsLastPage,
             },
         } = this;
 
@@ -176,8 +177,9 @@ export default class UserProfile extends React.Component {
             tab_content = (
                 <NotificationsList
                     username={accountname}
-                    notifications={notifications}
+                    notifications={notifications && notifications.toJS()}
                     loading={fetching}
+                    isLastpage={isNotificationsLastPage}
                 />
             );
         } else if (section === 'settings') {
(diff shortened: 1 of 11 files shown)


src/app/components/pages/UserProfile.jsx
```diff
@@ -176,8 +177,9 @@ export default class UserProfile extends React.Component {
             tab_content = (
                 <NotificationsList
                     username={accountname}
-                    notifications={notifications}
+                    notifications={notifications && notifications.toJS()}
```
roadscape: Most of our components seem to rely on immutablejs; it's a fairly consistent pattern. What's the reason for the switch here, and should we consider moving back to vanilla js as we move forward?

gl2748: Yeah, I wasn't too sure, happy to change this back. I guess vanilla JS is just easier to work with at the component level. I think I read somewhere it's best to keep immutable to the state layer, for example calling toJS() on each component render has performance ramifications, and React's props diffing might not detect changes to immutable objects... But again, happy to change this back if i am mistaken.

PR #3640, opened 2019-12-09
Embedded player for 3Speak

quochuy:


src/shared/HtmlReady.js
```diff
@@ -87,7 +87,10 @@ export default function(html, { mutate = true, hideImages = false } = {}) {
     state.images = new Set();
     state.links = new Set();
     try {
-        const doc = DOMParser.parseFromString(html, 'text/html');
+        const doc = DOMParser.parseFromString(
+            preprocessHtml(html),
```
roadscape: Why is this needed if `embedThreeSpeakNode` is called in `linkifyNode`?

quochuy:
Linkify node only allows transforming plain Url to an embedded player. 3speak however creates posts that contains an linked image, this allows to replace that image with a player:
https://steemit.com/hive-100421/@wehmoen/xrhjxocx

src/shared/HtmlReady.js
quochuy: Linkify node only allows transforming plain Url to an embedded player. 3speak however creates posts that contains an linked image, this allows to replace that image with a player:
https://steemit.com/hive-100421/@wehmoen/xrhjxocx

Additionally, linkify is not being called because the HTML image tag is wrapped inside a CENTER tag, which is why I needed to preprocess the HTML first

roadscape:
Local PR: #3655

PR #3653, opened 2019-12-16
Adding links to posts, comments and payout from profile menu

quochuy:
Screen Shot 2019-12-16 at 7 49 12 pm


roadscape:
Thanks for the PR @quochuy -- I did limit the change to just adding comments for now; 'Payouts' view is low priority, and 'Comments' gets you one click away from 'Posts'. In the spirit of keeping this menu simple, let's see where this gets us.

PR #3655, opened 2019-12-18
Community - Embedded player for 3Speak 

roadscape:
From @quochuy #3640.

> Linkify node only allows transforming plain Url to an embedded player. 3speak however creates posts that contains an linked image, this allows to replace that image with a player:
https://steemit.com/hive-100421/@wehmoen/xrhjxocx
>
> Additionally, linkify only processes plain text nodes, however the MD for images is already transformed into HTML and linkify isn’t called for the IMG tags which is why I needed to preprocess the HTML first.

### DIFF
--- a/src/app/components/cards/MarkdownViewer.jsx
+++ b/src/app/components/cards/MarkdownViewer.jsx
@@ -156,7 +156,7 @@ class MarkdownViewer extends Component {
                 if (type === 'youtube') {
                     sections.push(
                         <YoutubePreview
-                            key={idx++}
+                            key={id}
                             width={w}
                             height={h}
                             youTubeId={id}
@@ -168,15 +168,16 @@ class MarkdownViewer extends Component {
                 } else if (type === 'threespeak') {
                     const url = `https://3speak.online/embed?v=${id}`;
                     sections.push(
-                        <div className="videoWrapper" key={idx++}>
+                        <div className="videoWrapper" key={id}>
                             <iframe
                                 src={url}
                                 width={w}
                                 height={h}
                                 frameBorder="0"
-                                webkitallowfullscreen
-                                mozallowfullscreen
+                                webkitallowfullscreen="true"
+                                mozallowfullscreen="true"
                                 allowFullScreen
+                                title={`ThreeSpeak video ${id}`}
                             />
                         </div>
                     );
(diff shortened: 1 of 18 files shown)


roadscape:
Tested working :+1:

quochuy:
Nice

quochuy:
@roadscape is there anything blocking the release?

src/app/components/cards/MarkdownViewer.jsx
```diff
@@ -199,9 +213,8 @@ class MarkdownViewer extends Component {
                 } else if (type === 'dtube') {
                     const url = `https://emb.d.tube/#!/${id}`;
                     sections.push(
-                        <div className="videoWrapper">
+                        <div className="videoWrapper" key={idx++}>
                             <iframe
```
gl2748: Please use something other than index for the key as this will confuse React's Dom diffing, when for example more than one element has a key of say: 1...
The key ought to be a globally (to the dom) value... in this case url would likely suffice.

src/app/utils/Links.js
```diff
@@ -51,6 +51,8 @@ export default {
     twitch: /https?:\/\/(?:www.)?twitch.tv\/(?:(videos)\/)?([a-zA-Z0-9][\w]{3,24})/i,
     dtube: /https:\/\/(?:emb\.)?(?:d.tube\/\#\!\/(?:v\/)?)([a-zA-Z0-9\-\.\/]*)/,
     dtubeId: /(?:d\.tube\/#!\/(?:v\/)?([a-zA-Z0-9\-\.\/]*))+/,
+    threespeak: /(?:https?:\/\/(?:3speak.online\/watch\?v=)|(?:3speak.online\/embed\?v=))([A-Za-z0-9\_\-\/]+)/i,
+    threespeakImageLink: /<a href="(https?:\/\/3speak.online\/watch\?v=([A-Za-z0-9\_\-\/]+))"><img.*?><\/a>/i,
```
gl2748: Please add a passing test in the test suite for Links. `Links.test.js`

quochuy:
@gl2748 please check the new PR https://github.com/steemit/condenser/pull/3675

roadscape:
Thanks @quochuy!

PR #3662, opened 2020-01-06
Only show Follow button in user profile banner when we are sure.

quochuy:
- Show "Follow" button from server side rendering
- If user is logged on, the frontend will hide the "Follow" button until it gets a definitive result from API.


roadscape:
Thanks @quochuy!

PR #3664, opened 2020-01-07
Allow setting of custom max_accepted_payout values...

quochuy:
...from the advanced settings when creating a post

The "decline payout" has also been moved from payout type to max accepted payout as this is where it should be.

Screen Shot 2020-01-08 at 7 43 10 am

Screen Shot 2020-01-08 at 7 43 16 am

Screen Shot 2020-01-08 at 7 43 26 am


roadscape:
Are these changes compatible with the account-level defaults which are set on the Settings page?

image

PR #3674, opened 2020-01-16
Post template management

quochuy:
I've added a template management feature for blog posts.
From the Advanced Settings panel of the Reply Editor, you can now save your post (title, body, tags), its payout type and beneficiaries settings into a local storage configuration that can be reloaded for future posts of the same type.

To create a new template: in the advanced settings, give a name and click the "Save" button
To load an existing template: in the advanced settings, select a template and click the "Save" button.

Screen Shot 2020-01-15 at 10 36 37 pm

Screen Shot 2020-01-15 at 10 37 02 pm

### DIFF
--- a/src/app/components/cards/PostTemplateSelector.jsx
+++ b/src/app/components/cards/PostTemplateSelector.jsx
@@ -1,0 +1,88 @@
+import React from 'react';
+import tt from 'counterpart';
+
+class PostTemplateSelector extends React.Component {
+    static propTypes = {
+        username: React.PropTypes.string.isRequired,
+        onChange: React.PropTypes.func.isRequired,
+    };
+
+    render() {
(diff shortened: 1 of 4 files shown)


roadscape:
Same as #3659, we should wait until communities are merged to master.

There is an additional concern: more than a template editor, condenser really needs a way to save drafts. And it seem that if a draft could be easily 'copied', this would create a template function. Drafts could be implemented via [conveyor](https://github.com/steemit/conveyor), which allows for authenticated read/write calls.

quochuy:
Would draft be saved locally? I will check what’s involved with the conveyor.
Also draft and templates differ by their TTL. A draft is short live, it lives until it is edited and posted. A templates lives longer, until it gets deleted and can be cloned into a draft to create a new post.

PR #3642, opened 2019-12-09
3614 im notifications modal community

gl2748:
Closes #3614 

Screen Shot 2019-12-09 at 4 28 06 PM

TODO:

- [x] Add moderation log link to `CommunityPaneMobile.jsx`
- [x] Make sure pagination works for the NotificationList.jsx component when showing notifications inside a modal.

Issue #3614, opened 2019-11-18
notification modal on communities home


roadscape:
As a community member, I'd like to be able to access a list of community notifications from the community's home page.

- "moderation log" link under "Leadership" section should bring up a modal
- Make API request to notifications endpoint for the specific community
- Reuse account NotificationsList rendering (refactor if needed)

roadscape:
https://github.com/steemit/condenser/pull/3642 merged

PR #3676, opened 2020-01-17
3525 post notifications

### LINKED ISSUE #3525: show notifs for a post
As a moderator or curious user, I want to see a list of all flags and actions performed on a post.

Leveraging the `bridge.post_notifications` API, add the ability to view notifications relating to a specific post.

gl2748:
Closes #3525 
Screen Shot 2020-01-17 at 5 27 58 AM
Screen Shot 2020-01-17 at 5 27 52 AM
Screen Shot 2020-01-17 at 5 27 43 AM

### DIFF
--- a/src/app/components/cards/PostNotificationsList.jsx
+++ b/src/app/components/cards/PostNotificationsList.jsx
@@ -1,0 +1,120 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import TimeAgoWrapper from 'app/components/elements/TimeAgoWrapper';
+import LoadingIndicator from 'app/components/elements/LoadingIndicator';
+import { actions as fetchDataActions } from 'app/redux/FetchDataSaga';
+import Callout from 'app/components/elements/Callout';
+
+class PostNotificationsList extends React.Component {
+    static propTypes = {
+        notifications: PropTypes.arrayOf(
+            PropTypes.shape({
+                id: PropTypes.number,
+                type: PropTypes.string,
+                score: PropTypes.number,
+                date: PropTypes.date,
+                msg: PropTypes.string,
+                url: PropTypes.url,
+            })
+        ),
+        author: PropTypes.string.isRequired,
+        permlink: PropTypes.string.isRequired,
+        loading: PropTypes.bool,
+        fetchPostNotifications: PropTypes.func.isRequired,
+    };
+
+    static defaultProps = {
+        notifications: [],
+        loading: true,
+    };
+
+    constructor() {
+        super();
+    }
+
+    componentWillMount() {
+        const { author, permlink, fetchPostNotifications } = this.props;
+        fetchPostNotifications(author, permlink);
+    }
+
+    render() {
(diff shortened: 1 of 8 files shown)


gl2748:
This needs to be rebased on `bridge-api-dev` after PR #3642 for #3614 is merged as changes to how pagination works from that work might conflict.

roadscape:
#3642 merged; `bridge-api-dev` merged into this branch without issue. Still needs testing

roadscape:
Tested, working well. @gl2748 could you please add pagination?

gl2748:
It looks like the Post Notifications API doesn't currently support pagination. Currently I think it responds with all of the recent activity on a post/comment.

It's also worth noting that the copy for the notification.msg property is:
`<username> replied to your post`
Suggest 'your post' be changed to the post's author's username.

roadscape:
Hmm.. hive's current signature is `post_notifications(author, permlink, min_score=25, last_id=None, limit=100)` so pagination should be available.

gl2748:
Ah thanks wasn't aware of that!

src/app/components/cards/PostNotificationsList.jsx
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import TimeAgoWrapper from 'app/components/elements/TimeAgoWrapper';
+import LoadingIndicator from 'app/components/elements/LoadingIndicator';
+import { actions as fetchDataActions } from 'app/redux/FetchDataSaga';
+import Callout from 'app/components/elements/Callout';
+
+class PostNotificationsList extends React.Component {
```
roadscape: @gl2748 can we re-use the existing NotificationsList, or at least the bulk of its rendering logic? It still needs quite a bit of polishing and ideally we don't have to modify two copies.

gl2748: @roadscape I've now done this.

roadscape:
Looking good, only thing I see is some strange behavior here: https://steemitdev.com/hive-171487/@test-safari/6dcd9m-did-you-know

1. Click "Activity" on the OP:
image
"Load more" is shown on top.

2. Click "Load more":
image
No new records are shown, but the button disappears. Also, at this point, I cannot close the modal (neither clicking on dark region nor "X" works)

gl2748:
Close Button bug is addressed in another PR, i'll try and dig it up - it's caused by the notification having a higher z-index than the close-button.

Regarding the expected behavior for the `Load more...` button.
On first call to get notifications, there's no way to know if all the notifications have been fetched, so the `load more` button appears by default, once clicked, a second get request is made for more notifications, if this returns an empty list, we can infer that we've got all the notifications and the `load more` button no longer appears.

Perhaps the solution here is to say:

If number of notifications from the first request is less than to the limit of notifications requested, e.g. '50', do not display the `load more...` button... Otherwise display the `load more...` button.

gl2748:
Close button bug addressed here:
https://github.com/steemit/condenser/issues/3723

Issue #3667, opened 2020-01-09
subscription status does not update after login


roadscape:
> When I go to the community and log in straight from the page, my already subscribed page "Subscribe" button is still active.

PostsIndex will need to re-fetch the community with the appropriate `observer` when the current account is updated.

roadscape:
https://github.com/steemit/condenser/pull/3684 merged

PR #3684, opened 2020-01-23
Im xxx improve loading and empty states

gl2748:
Fixes - When loading an index of posts at `/trending/my` for example there is a flash of SSR content indicating that the logged in user has not subscribed to any community:
### Before.
Screen Shot 2020-01-23 at 10 56 33 AM

### Before.
Fixes - When logging out of the site, at `/trending/my` for example, the list of 'My subscriptions' remains visible until the user refreshes the page.
Screen Shot 2020-01-23 at 10 57 49 AM

Fixes - When logging in/out at `/trending/my` , the list of posts from subscribed communities is not updating.


gl2748:
also closes #3667,

Other bugs addressed:
> As an unauthenticated user, visiting the communities exploration page at: /communities and clicking the subscribe button for a community - I am prompted to login, after logging in, the subscribe action fails.

> As an unauthenticated user, when i visit a community page, e.g. /trending/hive-100421 I only see the 'new post' button, as an unauthenticated user this is perhaps not a useful action, perhaps 'subscribe' would be better, with the user being prompted to login / create an account.

src/app/components/modules/LoginForm.jsx
```diff
@@ -406,7 +406,7 @@ class LoginForm extends Component {
                         </button>
                     )}
                 </div>
-                {/*signupLink*/}
+                {signupLink}
```
roadscape: Please revert

src/app/components/modules/Modals.jsx
```diff
@@ -82,11 +90,16 @@ class Modals extends React.Component {
...
         };
-
         return (
             <div>
                 {show_login_modal && (
-                    <Reveal onHide={hideLogin} show={show_login_modal}>
+                    <Reveal
+                        onHide={() => {
+                            hideLogin();
+                            loginBroadcastOperation.errorCallback();
```
roadscape: Just to be clear - this will automatically trigger errorCallback any time the login modal is shown and cancelled? Two concerns: (1) do any callers already handle this on their own, (2) is onHide only for cancel or is it possible to be triggered when the dialog closes because of success?

src/app/components/pages/PostsIndex.jsx
```diff
@@ -106,10 +105,18 @@ class PostsIndex extends React.Component {
...
         } = this.props;
 
+        const status = this.props.status
+            ? this.props.status.getIn([category || '', order])
+            : null;
+        let fetching = (status && status.fetching) || this.props.loading;
+
         let emptyText = '';
-        if (order === 'feed') {
+        if (!process.env.BROWSER) {
```
roadscape: Hmm.. I'm not sure about this, wouldn't it prevent SSR rendering of any post lists?

src/app/redux/CommunitySearchReducer.js
```diff
@@ -0,0 +1,55 @@
+import { Map, List, fromJS } from 'immutable';
```
roadscape: Seems this file may have been included by accident?

src/app/components/pages/PostsIndex.jsx
gl2748: Ah good point, yes, I suggest:
```
if (!process.env.BROWSER && category === 'my') {
  fetching = true;
}
```

src/app/components/modules/Modals.jsx
gl2748: Yes very good points, I'm going to undo this and simply not display a subscribe action pending on the subscribe button when an unauthenticated user clicks it - because a subscribe action isn't pending until they are authenticated.

src/app/components/pages/PostsIndex.jsx
```diff
@@ -152,6 +156,28 @@ class PostsIndex extends React.Component {
...
+            />
+        );
+
+        if (!fetching && !posts.size) {
+            postsIndexDisplay = <Callout>{emptyText}</Callout>;
+        }
+        if (!username && posts.size && category === 'my') {
+            postsIndexDisplay = <Callout>{emptyText}</Callout>;
+        }
+        if (order === 'feed' && !username) {
```
roadscape: This case could be removed; `feed` does not require a user to be logged in

src/app/components/pages/Topics.jsx
```diff
@@ -53,15 +53,18 @@ class Topics extends Component {
             let options = [];
             options.push(opt(null));
 
-            if (username) {
+            if (username && subscriptions) {
```
roadscape: Potentially simplify modified logic

PR #3693, opened 2020-02-03
fix modal text-color on mbile

gl2748:
# Before
Screen Shot 2020-02-03 at 11 41 13 AM

# After
Screen Shot 2020-02-03 at 11 45 59 AM

### DIFF
--- a/src/app/assets/stylesheets/foundation-overrides.scss
+++ b/src/app/assets/stylesheets/foundation-overrides.scss
@@ -121,7 +121,7 @@ tbody tr:nth-child(even) {
   outline: none;
   @include themify($themes) {
     background-color: themed('modalBackgroundColor');
-    color: themed('modalTextColorPrimary');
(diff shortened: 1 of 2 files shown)


src/app/assets/stylesheets/foundation-overrides.scss
```diff
@@ -121,7 +121,7 @@ tbody tr:nth-child(even) {
   outline: none;
   @include themify($themes) {
     background-color: themed('modalBackgroundColor');
-    color: themed('modalTextColorPrimary');
```
roadscape: Is `modalTextColorPrimary` used elsewhere? Wondering if `modalTextColorPrimary` itself needs to be changed

roadscape:
For now, set `modalTextColorPrimary` of nightmode theme to value of `textColorPrimary`; revert changes to `src/app/assets/stylesheets/foundation-overrides.scss`

PR #3694, opened 2020-02-04
allow categories in url with 1 letter, refs #3691

### LINKED ISSUE #3691: URL does not work on beta site
https://beta.steemit.com/c/@don.quijote/posts-votados-por-el-equipo-don-quijote-07-03-2019-23-34-08-03-2019-23-34

This URL throws a 404 and fails to load, but on Steemit.com it loads fine.

gl2748:
Closes #3691


roadscape:
Need to verify this works w/ hive op validation

PR #3707, opened 2020-02-19
Search bar css

quochuy:
- removing duplicated lines of CSS that override one another.
- better centering of text inside input
- white text when hovering the input

Screen Shot 2020-02-20 at 12 18 24 am

Screen Shot 2020-02-20 at 12 18 37 am


roadscape:
Local PR: https://github.com/steemit/condenser/pull/3712/files

PR #3724, opened 2020-02-21
3700 carefully access local storage

### LINKED ISSUE #3700: SPA throws uncaught exception due to accessing localStorage directly
`Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

```
Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:353228)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:743178)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:499992)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:498485)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:473479)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:476072)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:458628)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:459140)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
    at Object.<anonymous> (https://steemit.com/assets/app.7f38b401ec746a4e6664.js:1:406840)
    at e (https://steemit.com/assets/vendor.7f38b401ec746a4e6664.js:1:101)
```

Accessing `window.localStorage` indiscriminately is not advised.  Per spec, reading or writing to it with localStorage disabled in a browser will throw an exception, not a silent failure.  If you don't catch it, your app breaks.

Per https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/ :

```
function isSupported(getStorage) {
  try {
    const key = "__some_random_key_you_are_not_going_to_use__";
    getStorage().setItem(key, key);
    getStorage().removeItem(key);
    return true;
  } catch (e) {
    return false;
  }
}

isSupported(() => localStorage); // => true | false
```

This way you test for supportedness inside of a try/catch and then can branch on it later, so you're not throwing exceptions in the main app.

gl2748:
Closes #3700


roadscape:
The issue indicated that the error occurred when trying to *read* from `localStorage`. These changes only address writing. Also wondering if the `try` check may end up being expensive. Is that the recommended approach?

quochuy:
@roadscape maybe check that localStorage is accessible at a higher level and set a flag, maybe inside a store.

To make it simpler, @gl2748 could also modify AccessLocalStorage() to save a state upon first call instead of doing the try every time

gl2748:
@quochuy good idea - to take it one step further - the best practice here is to persist desired slices of the application store/state with the redux `subscribe` method and to retrieve persisted state *once* from localStorage at application startup. The most succinct description i've found of this is here:

https://egghead.io/lessons/javascript-redux-persisting-the-state-to-the-local-storage

But unfortunately, we're accessing localStorage all over the place and in inconsistent manner... both reading and writing to it in multiple places - which touches on the point made by @roadscape:

> The issue indicated that the error occurred when trying to read from localStorage

Our codebase, failing to be typed as it is, handles this case - when initializing variables from localStorage by accounting for the fact it might be `undefined`. i.e. it treats read operations from localStorage by a user with privacy settings that preclude access to localStorage in the same way it treats a new visitor to the site, who does not have the desired key being retrieved in localStorage yet. While not a particularly good solution (read operations will throw an uncaught error) variables initialized as `undefined` from localStorage read operations are widely accounted for in the codebase ...

PR #3734, opened 2020-02-22
Simple filters for notifications

quochuy:
- localize static texts in notification list
- simple filters for notifications, they just filter the results at frontend side (hide/show)

Screen Shot 2020-02-22 at 10 07 44 pm

Screen Shot 2020-02-22 at 10 59 17 pm

Screen Shot 2020-02-22 at 10 59 27 pm

PR #3743, opened 2020-02-28
Quochuy notification filters

gl2748:
In organization version of: https://github.com/steemit/condenser/pull/3734


gl2748:
@quochuy i agree with this in principle. One change to be made is that the filters appear in a block of code with the following condition:

https://github.com/steemit/condenser/pull/3743/files#diff-83c21b84327f6fc7bb2d1fe315663e50R258

Which means the filters will not appear when there are no unread notifications. I suggest putting the filters outside this condition, as it pertains to the 'mark as read' button alone.

quochuy:
@gl2748
Will have a look soon

Benson0224:
@gl2748  is this PR still valid? if yes please help to merge it, thanks

gl2748:
@Benson0224 yes, still valid, please review.

PR #3755, opened 2020-03-19
Bridge api dev test

leon-tron:


kirkins:
Blacklisting posts about a fork? That seems low...

ScottSallinen:
Front end blacklisting of content is a very inefficient programming strategy. It would be much more efficient to delete the content from the underlying centralized database instead.

PR #3757, opened 2020-03-20
update terms of service 

leojlz007:


lazybookwyrm:
Editing your terms of service so you can censor anyone you want to?

Issue #3937, opened 2025-03-12
Admin and Moderator Privileges Not Appearing


the-gorilla-steem:
The hive-xxxxxx in the URL was used to set admin and moderator permissions for a post.

If the user accesses a post from the community where the state is set, admin actions (e.g. mute, pin, etc.) are visible. 

If the page is then refreshed, these actions disappear because the permissions are forgotten.

PR #3938, opened 2025-03-12
Fixed Bug #3937

the-gorilla-steem:
Fix for bug #3937 

If there's no community information in the state, call the geCommunity function to update the state.

This information is then used to update the page permissions for admins and moderators.

@ety001 - Please can you deploy to steemitdev.com


the-gorilla-steem:
@ety001 - Testing has gone well so can you push to live please.

ety001 (approved):

PR #3984, opened 2026-07-08
feat(cache): server-side stale-while-error caching layer

ety001:
Port wallet's three-tier cache model into condenser's `next` branch:
**Browser (L1, follow-up PR) → Redis (L2, this PR) → Steem RPC (L3).**

This is **PR1 (server-side foundation)** of the cache-layer migration. It targets the #1 user-facing pain point — frequent connection errors and high latency under RPC load — by serving cached/stale data instead of erroring when the Steem node is overloaded.

## What this adds

**`lib/cache/` (new)**
- `redis.ts` — content-cache Redis singleton, namespaced via `REDIS_KEY_PREFIX=condenser` so it never collides with the existing session store (`steem:session:`). All functions no-op when `REDIS_URL` is unset, so behaviour is identical to pre-cache.
- `server-cache.ts` — `withCache(key, ttl, staleTtl, fetcher)` stale-while-error wrapper: serves fresh cache, and on RPC failure returns stale data (degraded) rather than 500-ing.
- `health-monitor.ts` — shared Steem node health state + probe lock (`SET NX`) to prevent cross-instance probe storms.

**`app/api/health/route.ts` (new)** — stale-while-revalidate health probe with in-process single-flight.

## Caching strategy (`lib/steem/client.ts`)
| Data type | Cache? |
|---|---|
| Global data (dynamic-global-properties, communities, community-roles, followers/following lists, single posts) | ✅ cached for all traffic |
| Personalised data (ranked/account posts, profiles, communities list) | ⚠️ **degraded-hit**: cached only when `observer===undefined` (anonymous); logged-in observers bypass Redis to prevent cross-user state leakage |
| Notifications, account lookups, broadcasts | ❌ never cached |

TTLs are **initial values to be tuned** after observing real freshness in prod.

## Write-after-invalidation (`app/api/steem/broadcast/route.ts`)
On successful vote/comment/reblog, drops the affected read-cache prefixes via `cacheDeleteByPrefix` (scoped per-operation, not a blanket flush) and sets an `X-Cache-Invalidate` header for the PR2 client L1 layer to consume.

## Verification
- ✅ `tsc --noEmit` — clean
- ✅ `eslint` — clean
- ✅ `next build` — compiled successfully, `/api/health` route registered

## Follow-ups (separate PRs)
- **PR2**: client SWR (`client-cache.ts`/`client-fetch.ts` + rewire `lib/api/steem.ts`), deprecate Redux cache skeletons.
- **PR3**: degradation UI banner (`use-service-health` + `DegradationBanner` in root layout).

Not in scope: master's July image-proxy fixes (#3976–#3978), `next/image` adoption, voting-slider UX.

🤖 Generated with Claude Code

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,12 @@ JWT_SECRET=your-secret-key-change-in-production
 # REDIS_DB=0
 # REDIS_KEY_PREFIX=steem:session:
 
+# Content Cache (uses the same REDIS_URL above when set)
+# When REDIS_URL is configured, Steem RPC responses are cached server-side
+# (stale-while-error) to cut latency and survive node outages. Cached keys are
+# namespaced separately from sessions, so both can share one Redis instance.
+# REDIS_KEY_PREFIX=condenser
+
 # Other Configuration
 NEXT_PUBLIC_SIGNUP_URL=https://signup.steemit.com
 
--- a/app/api/health/route.ts
+++ b/app/api/health/route.ts
@@ -1,0 +1,123 @@
+/**
+ * GET /api/health
+ * Steem node health check with stale-while-revalidate caching.
+ *
+ * - Fresh cache (<60s): return immediately, no RPC.
+ * - Stale cache: another instance may be probing; if we cannot acquire the
+ *   probe lock, serve stale data (X-Health-Stale) instead of triggering a
+ *   probe storm.
+ * - No cache / lock acquired: run a single shared in-flight probe, write the
+ *   result to Redis, and return it.
+ *
+ * Returns 200 when healthy, 503 when degraded.
+ */
+
+import { NextResponse } from 'next/server';
+import { checkSteemNodeHealth } from '@/lib/steem/client';
+import {
+  getSteemHealthStale,
+  markSteemHealthy,
+  markSteemUnhealthy,
+  acquireProbeLock,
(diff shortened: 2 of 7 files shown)


ety001:
### 📌 Carry-over to PR3: throttle `/api/health` when Redis is unavailable

Self-audit item #4 (non-blocking for this PR, but matters once PR3 lands):

When `REDIS_URL` is not configured, `acquireProbeLock()` / `getSteemHealthStale()` / `markSteemHealthy()` all no-op, so `/api/health` loses its three gating layers and **every request probes the Steem node directly**. This is inherited from the wallet project's behavior and is acceptable today because nothing polls the endpoint frequently.

**But PR3 introduces `use-service-health`, which polls `/api/health` every ~60s from every client.** With N server instances and no Redis, that becomes N probes/60s against the node — the amplification risk is introduced by PR3, so the throttle fix belongs there.

**When doing PR3, address one of:**
1. Short-circuit at the top of the health route when Redis is off: `if (!getRedis()) return buildResponse(true)` (simplest; local dev just can't observe node status).
2. Add an in-process minimum TTL (memory var recording last-probe time, skip if within e.g. 5s) so the no-Redis path still self-limits.

Recording here so it's not forgotten when PR3 starts.

PR #3986, opened 2026-07-10
feat(cache): degradation UI banner + no-Redis health throttle (PR3)

ety001:
Completes the three-tier cache migration with the user-facing degraded signal and a no-Redis safety net for `/api/health`. This is **PR3** of the cache migration (PR1 #3984 server-side Redis, PR2 #3985 client SWR).

## Degradation UI
- **`hooks/use-service-health.ts`** — polls `/api/health` every 60s, re-checks immediately on tab focus, and pauses polling while the tab is hidden to avoid needless load. Guards `setState` with an `active` flag so async resolves after unmount are dropped (satisfies `react-hooks/set-state-in-effect`).
- **`components/layout/DegradationBanner.tsx`** — amber banner when the node is serving stale cache ("Showing cached content — the Steem network is slow to respond."), red banner on outage. Hidden while healthy or before the first poll resolves.
- **`AppShell.tsx`** — banner mounted just below the header, above main content.

## No-Redis health throttle (carry-over from #3984)
Without Redis, the cache/probe-lock layers in `/api/health` are no-ops, so every request would probe the node directly. PR2's client poll (60s × N instances) turns that into an amplification risk. Added an in-process minimum interval (**5s**) that replays the last probe result within the window, capping the no-Redis path to one probe per instance per interval. The Redis path is unaffected (already throttled by cache + probe lock).

### Trade-off worth noting
On the no-Redis path, if a probe *fails* (throw), the in-memory result isn't updated, so requests within the next 5s replay the previous *successful* probe result. This is acceptable — it's the stale-while-error spirit (briefly show last-known-good rather than re-probing per request), only affects the no-Redis degraded path, and the failing request itself still returns the correct outage signal.

## Verification
- tsc --noEmit — clean
- eslint — clean (fixed set-state-in-effect via active guard)
- next build — Compiled successfully

## Migration status
This closes the cache-layer migration:
| PR | Layer | Status |
|---|---|---|
| #3984 | Redis L2 (server) | merged |
| #3985 | Browser L1 (client) | merged |
| this PR | Degraded UI + no-Redis throttle | review |

### DIFF
--- a/app/api/health/route.ts
+++ b/app/api/health/route.ts
@@ -8,12 +8,16 @@
  *   probe storm.
  * - No cache / lock acquired: run a single shared in-flight probe, write the
  *   result to Redis, and return it.
+ * - No Redis: cache/lock layers are no-ops; an in-process minimum interval
+ *   (5s) replays the last probe result to avoid amplifying the 60s client
+ *   poll into per-request node probes.
  *
  * Returns 200 when healthy, 503 when degraded.
  */
 
 import { NextResponse } from 'next/server';
 import { checkSteemNodeHealth } from '@/lib/steem/client';
+import { getRedis } from '@/lib/cache/redis';
 import {
   getSteemHealthStale,
   markSteemHealthy,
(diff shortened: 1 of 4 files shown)


ety001:
### Self-audit notes (for reviewers)

Ran a self-review before this is looked at. Findings recorded here for context.

**Fixed in this PR:**
- ❌ `useServiceHealth` keyed status on `res.ok`, which collapsed the route's `'degraded'` (HTTP 503, node slow / serving stale) into `'outage'`. The amber "showing cached content" banner would never have appeared — only the red outage one. **Fixed**: status is now derived from the response body's `status` field regardless of HTTP code; `'outage'` is only reached when `/api/health` itself is unreachable or the body is unparseable. (Confirmed the route only ever emits `healthy`/`degraded` in the body, so this is consistent.)

**Known, non-blocking (left as-is):**
- 🟡 **No-Redis cold-start window**: on the no-Redis path, `lastInMemoryHealth` is null right after boot, so the 5s throttle can't kick in until the first probe completes. Cold start thus lets the first request per instance reach the node. Low impact — single-flight still limits each instance to one probe, and cold start is a one-time event.
- 🟡 **visibilitychange interval restart**: hidden→visible recreates the interval only when `intervalRef.current` is null (i.e. it was cleared on hide), so the design avoids double intervals. A theoretical race with the effect cleanup exists but is near-impossible in practice (React runs cleanup synchronously).
- 🟢 **Banner position under `<GlobalModals>`**: mounted below `<Header />` and above content. Needs a visual once-over to confirm `<GlobalModals>` (likely portal/fixed) doesn't push it out of place — a runtime/visual check, not a code defect.

None of these block merge; the first one was the only merge-blocker and it's resolved.

PR #3991, opened 2026-08-16
Add native IPFS video player compatibility for speak-on-steem

Alejos7ven:


ety001:
## Security Review — request changes

I audited this PR end-to-end against the actual pipeline versions (remarkable 1.7.1, xmldom 0.1.27, sanitize-html 1.14.1), running the real sanitize config on attack payloads rather than reasoning statically.

**The good news first:** there is no direct XSS here. `javascript:` / `data:` values in `src` are dropped by sanitize-html's scheme check (`naughtyHref` runs after `transformTags`), attribute breakouts via `"` / `<` are correctly escaped by both xmldom and sanitize-html's serializer, and event handlers are blocked by the attribute whitelist. All of this was verified empirically.

However, I found two medium-severity issues and one medium-low that should be addressed before merging.

---

### 1. [Medium] ReDoS in `replaceIPFSVideoLinks` — verified, amplifies to SSR

The anchor regex has quadratic backtracking:

```js
/<a[^>]*href="([^"]+)"[^>]*>\s*IPFS Video\s*<\/a>/gi
```

For every `<a` start position, `[^>]*` scans forward to the next `>` and backtracks. A naive payload (`'<ax'.repeat(N)`) gets escaped by remarkable, but hiding `<a ` tokens inside a **quoted attribute value** survives markdown rendering verbatim:

```
<a title="<a <a <a ... (repeated) ...">x</a>
```

Measured with the actual regex on remarkable-rendered output:

| post size | regex time (single invocation) |
|---|---|
| ~30 KB | 609 ms |
| ~60 KB | **1368 ms** (STEEM body limit is 65536 chars, so this is reachable) |

Amplification — `preprocessHtml` runs even when `mutate: false` (`HtmlReady.js` calls it before parsing unconditionally), so one on-chain post costs multi-second CPU on **every uncached render**:

- `MarkdownViewer.jsx:104` — SSR + client render of the post page
- `ExtractContent.js:42` (`extractImageLink`) — SSR meta extraction, **and feed cards** via `PostSummary`, so the payload fires on every feed page that lists the post

**Suggested fix:** don't regex-replace on the raw string. After the xmldom parse, iterate `<a>` nodes, compare `textContent === 'IPFS Video'`, and replace the node. If a regex must stay, require `href` to be the first attribute and bound every quantifier (e.g. `[^>]{0,200}`).

### 2. [Medium] CSS injection via the `video@style` substring gate — verified

The transform passes through the **entire** style attribute if it merely *contains* the substring `width`, `margin`, or `aspect-ratio`. sanitize-html 1.14.1 does no style validation (no `allowedStyles` in our config), so this survives end-to-end:

```html
<video style="width:1px;position:fixed;top:0;left:0;width:100vw;height:100vh;z-index:99999;background-image:url(https://evil.example/log)"></video>
```

Impact:

- `position:fixed` full-viewport overlay with an arbitrary background image — visual spoofing over the site chrome and blocking of all interaction beneath. This is the *only* injection point that can escape the document flow (no other tag allows `style`).
- `url()` fetches leak every reader's IP / UA / timestamp to an attacker-controlled server.

Note the precedent in this same file: the `td` transform requires an **exact match** (`attribs.style === 'text-align:right'`). A substring `includes()` is a much weaker gate.

**Suggested fix:** drop the passthrough branch entirely and always apply your fixed default style — there's no functionality loss, the default already exists in this PR.

### 3. [Medium-low] Arbitrary-origin media embed → reader IP disclosure — verified

Three paths let any post author make every visitor's browser fetch from an arbitrary origin (all verified passing the sanitizer):

1. `<a href="https://tracker.example/x">IPFS Video</a>` — when the URL has no `ipfs/` substring and no 46+ char alphanumeric run, the original href is kept verbatim as `src`.
2. Raw `<video><source src="https://any-host/x.mp4">` is now allowed in **all posts, including historical ones** once this ships.
3. Protocol-relative `//evil.com/x.mp4` passes — sanitize-html 1.14.1 defaults to `allowProtocolRelative: true` and our config doesn't override it.

For context, images are proxied server-side specifically to protect reader IPs from third parties (PRs #3976–#3978 just hardened exactly this). Videos bypass all proxying.

**Suggested fix:** enforce a gateway allowlist in the `source` transform — the `iframeWhitelist` pattern already exists in this file. E.g. only allow `https://gateway.pinata.cloud/ipfs/<valid CID>` and reject everything else.

### 4. Minor notes

- The fallback regex `([a-zA-Z0-9]{46,})(\/.*)?` treats **any** 46+ alphanumeric run in **any** URL as a CID (a long YouTube video id would be rewritten to pinata). Please validate CIDv0 / base32 CIDv1 shape.
- `class` is passed through without a whitelist — inconsistent with the `div` transform's `classWhitelist` (sanitize-html only filters classes when `allowedClasses` is configured, which it isn't here).
- `crossorigin` value is passed through verbatim; it should be pinned to `anonymous`.
- The gateway `https://gateway.pinata.cloud` is hardcoded: all reader IPs flow to a single third party and it's a single point of failure. `$STM_Config.ipfs_prefix` already exists in the codebase for exactly this purpose.
- **Architecture suggestion:** the codebase already has a safer, established embed pattern — `~~~ embed:${id} type ~~~` placeholders replaced by React components with **fixed URL templates** (`MarkdownViewer.jsx:142`, used by youtube/vimeo/twitch/dtube/threespeak). Emitting e.g. `~~~ embed:CID ipfs ~~~` and rendering a fixed player component would eliminate issues 2 and 3 entirely, with zero author-controlled URLs.
- The added test only covers the happy path. Please add adversarial cases: `javascript:` href, hostile `style`, non-IPFS arbitrary URLs, protocol-relative URLs.

---

### Summary

| # | Issue | Severity | Verified |
|---|---|---|---|
| 1 | ReDoS — quadratic regex, amplified via SSR + feed cards | Medium | ✅ 1368 ms / 60 KB post |
| 2 | CSS injection — full-viewport overlay + reader tracking beacons | Medium | ✅ survives sanitizer |
| 3 | Arbitrary-origin embeds / reader IP disclosure (incl. historical posts) | Medium-low | ✅ all 3 paths pass |
| 4 | CID validation, class/crossorigin passthrough, hardcoded gateway | Low | code review |

The sanitize-html scheme check does catch the classic XSS vectors, but issues 1–3 are real, exploitable regressions. I'd suggest reworking this onto the `~~~ embed:` component pattern, or applying the per-issue fixes above, before merge. Happy to help review a revised version.

ety001:
## Follow-up: concrete suggestion — use the existing `~~~ embed:` component pattern

Expanding on the architecture note in my review above, since I think it's the cleanest path to resolve findings **2** and **3** outright and shrink the sanitizer diff instead of growing it.

### How the existing pattern works

Every rich embed in this codebase (youtube, vimeo, twitch, dtube, threespeak) already flows through one mechanism:

1. `HtmlReady` replaces the matched markup with a plain-text placeholder — see `embedThreeSpeakNode`: `` `~~~ embed:${id} threespeak ~~~` `` (`src/shared/HtmlReady.js:454`).
2. The placeholder passes through the sanitizer as ordinary text — nothing in it is author-controlled markup.
3. After sanitization, `MarkdownViewer.jsx:142-145` splits on `~~~ embed:` and renders a **React component with a fixed URL template**, e.g. threespeak builds `` `https://3speak.online/embed?v=${id}` `` at line 166. The post author never controls the URL.

### What this would look like for IPFS

- In `preprocessHtml`, replace the `IPFS Video` anchor with `` `~~~ embed:${cid} ipfs ~~~` `` (same shape as 3speak).
- In `MarkdownViewer.jsx:144`, add `ipfs` to the type alternation. The existing ID charset `[A-Za-z0-9?=_\-\/\.]+` already accepts base58 CIDv0 strings, so no other change to the matcher.
- Render an `<IpfsPlayer cid={id} />>` component that builds the source URL itself from a validated CID + a configurable gateway (e.g. `$STM_Config.ipfs_prefix`, which `HtmlReady.js:582` already supports, or a single constant).

### Why this is strictly better than the current approach

- **Finding 2 (CSS injection) disappears**: no `video`/`source`/`style` entries in `SanitizeConfig` at all — the entire 44-line diff to `SanitizeConfig.js` can be reverted. Attack surface shrinks instead of growing, and no historical post gains new allowed markup.
- **Finding 3 (arbitrary-origin embeds) disappears**: there is no author-controlled `src` anywhere; the component constructs the URL, so the gateway is enforceable in exactly one place (and swapping/adding gateways later is a one-line change instead of a sanitizer migration).
- **Consistent UX for free**: fixed dimensions, `videoWrapper` responsive sizing, and identical client + SSR rendering — same as every other embed on the site.
- **CID validation lands in one spot**: the component can reject anything that isn't a well-formed CIDv0 (`Qm...`, 46 chars) or base32 CIDv1, which also fixes the overly-broad `([a-zA-Z0-9]{46,})` fallback regex noted in finding 4.

### One caveat

This does **not** by itself fix finding 1 (ReDoS) — the anchor still has to be detected to emit the placeholder, so whichever detection you use should still be done DOM-side after the xmldom parse (or with bounded quantifiers). But that fix is needed either way, and it composes cleanly with this change.

Happy to review the revised PR if you go this route.

Alejos7ven:
Thank you very much for the detailed and thorough security audit. Your observations were extremely valuable. I have refactored the entire implementation to address all findings by aligning the feature with the codebase's native `~~~ embed:` placeholder architecture.

### Fixes implemented:

1. **ReDoS in preprocessHtml (Issue 1):** Removed the raw regex-replace entirely. The detection is now done DOM-side inside the `link()` traversal function using native `xmldom` APIs, evaluating nodes individually.
2. **CSS Injection & SanitizeConfig changes (Issue 2 & 3):** Reverted all modifications to `SanitizeConfig.js` back to its original secure state. The `<video>` markup is no longer allowed in post bodies. Instead, `HtmlReady.js` emits a safe `~~~ embed:CID ipfs ~~~` placeholder, which is processed and rendered securely by React inside `MarkdownViewer.jsx`.
3. **Arbitrary-Origin Embeds (Issue 3):** The source URL is built internally within the React component using the global `$STM_Config.ipfs_prefix` (falling back safely to the public Pinata gateway if undefined), meaning post authors cannot control or inject custom domains or trackers.
4. **CID Validation (Issue 4):** Replaced the broad alphanumeric fallback regex with strict verification for CIDv0 (base58 `Qm...`) and CIDv1 (base32 `b...`) structures. Any non-valid CID will be treated as a plain text link and skipped.
5. **Responsive Aspect Ratio:** Integrated the player using the existing `.videoWrapper` class layout with absolute positioning rules, maintaining responsive 16:9 consistency.

Unit tests have been updated and expanded to cover adversarial cases (javascript links, styling attempts, and non-IPFS links) and all 163 tests are passing successfully. Please take a look at the revised commit!

ety001:
## Re-review of the rewrite — big improvement, one blocking item left

Thanks for reworking this onto the `~~~ embed:` component pattern — the new architecture is exactly right. I re-ran the new `link()` logic against the real xmldom/remarkable pipeline and re-ran the sanitizer payloads against the current diff.

### What's fixed ✅

1. **ReDoS (finding 1) — resolved.** The raw-string regex is gone; the anchor conversion now happens DOM-side during traversal (`textContent.trim() === 'IPFS Video'`), which is linear and can't backtrack. Verified there's no remaining regex over untrusted HTML in the new path.
2. **The IPFS anchor path is now solid.** Verified end-to-end:
   - strict CID validation (CIDv0 `Qm` + 44 base58, CIDv1 base32) — `javascript:`, arbitrary non-CID URLs, and non-matching link text all correctly fall through to normal link handling (which re-applies the protocol normalization + external-link warning + phishing checks — nice);
   - the placeholder is plain text through the sanitizer;
   - the React component builds the URL from a **fixed template** + re-validated CID, with fixed style/crossOrigin props — no author-controlled attributes;
   - gateway is now configurable via `$STM_Config.ipfs_prefix` with a sane fallback.

### Still blocking ❌ — `SanitizeConfig.js` must be reverted

The `src/app/utils/SanitizeConfig.js` diff is byte-identical to the first revision (adds `video`/`source` to `allowedTags` plus the `style` substring gate and `source.src` passthrough). Under the new architecture **none of it is needed anymore** — the React-rendered `<video>` never passes through the sanitizer, and the placeholder is plain text. But it still keeps both original findings fully exploitable (just re-verified against this diff with sanitize-html 1.14.1):

```html
<!-- finding 2: CSS injection, survives verbatim -->
<video style="width:1px;position:fixed;top:0;left:0;width:100vw;height:100vh;z-index:99999;background-image:url(https://evil.example/log)"></video>

<!-- finding 3: arbitrary-origin embed, survives -->
<video controls="controls"><source src="https://evil.example/x.mp4" type="video/mp4"></video>
```

These require no `IPFS Video` anchor at all — they work via raw HTML in any post (new or historical). Since the feature no longer depends on them, please revert the entire `SanitizeConfig.js` change. That single revert closes both findings and the PR's sanitizer attack surface returns to exactly today's master.

### Minor notes (non-blocking)

- The embed `id` path may contain `../`, `?` and `=` (allowed by the `[A-Za-z0-9?=_\-\/\.]+` matcher). The browser normalizes `/..` so the URL stays on the trusted gateway host, but I'd reject `.`/`..` path segments in `link()` for hygiene.
- `child.textContent.trim() === 'IPFS Video'` is now case-sensitive (the v1 regex was case-insensitive) — just confirm that matches what the speak-on-steem dApp actually emits.
- Converted anchors skip `state.links.add(url)`, so those hrefs no longer appear in the collected links — flagging in case anything downstream relied on that.
- All-reader-IPs-to-Pinata remains a product-level privacy consideration; at least it's now configurable per environment.

With the `SanitizeConfig.js` revert, this LGTM from my side. The new tests are a nice touch, including the adversarial cases — appreciated.

Alejos7ven:
You are completely right. The `SanitizeConfig.js` changes were accidentally preserved from the initial commit when running `git commit --amend` without staging the reverted state. 

I have checked out `SanitizeConfig.js` to match the exact state of `master` and force-pushed the clean commit. The file has been completely removed from the PR diff, which now strictly changes only the logic files: `MarkdownViewer.jsx`, `HtmlReady.js`, and `HtmlReady.test.js`. 

Thank you again for the thorough review and guidance to land this safely. Ready for your final re-review!

ety001:
## Final review — LGTM 🎉

The `SanitizeConfig.js` change has been fully reverted, which was the last blocking item. Verified against the current diff with sanitize-html 1.14.1:

- **Finding 2 (CSS injection)** — raw `<video style="width:1px;position:fixed;...;background-image:url(...)">` is now stripped entirely by the sanitizer (`<p></p>`).
- **Finding 3 (arbitrary-origin embeds)** — raw `<video><source src="https://evil.example/x.mp4">` is likewise stripped, leaving only inner text.

Combined with the previous revision, the final state is:

| Finding | Status |
|---|---|
| 1. ReDoS in anchor detection | ✅ fixed (DOM-side conversion, no raw-HTML regex) |
| 2. CSS injection via `video@style` | ✅ fixed (sanitizer diff reverted) |
| 3. Arbitrary-origin embeds / reader IP disclosure | ✅ fixed (sanitizer diff reverted; component builds URL from fixed template + validated CID) |
| 4. CID validation, configurable gateway, adversarial tests | ✅ addressed (CIDv0/CIDv1 validated twice — at conversion and at render; `$STM_Config.ipfs_prefix`; hostile test cases added) |

The remaining nits from my last review (rejecting `.`/`..` path segments in the embed id, case-sensitivity of the `IPFS Video` text match, converted anchors dropping out of `state.links`) are all non-blocking polish — fine to address in a follow-up if desired.

From a security perspective this is ready to merge. Thanks for the constructive iteration — the end result is materially safer than the first revision (net sanitizer attack surface unchanged from master, zero author-controlled URLs/attributes in the embed).

ety001 (approved):

PR #4008, opened 2026-08-31
fix(security): harden open redirect validation against backslash bypass (F47 follow-up)

ety001:
## Summary

Follow-up to the F47 open redirect fix (#3993). The security team reported that the previous validation in `src/server/utils/RedirectTarget.js` only blocked the literal `//` prefix, and recommended switching to URL-parser-based origin validation.

## Root cause

Browsers normalize `\` to `/` and strip tab/CR/LF before URL parsing, so payloads that pass the old regex checks still navigate to an external host:

| Input | Browser navigates to | Old validator | New validator |
|---|---|---|---|
| `/\evil.com` | `https://evil.com/` | allow (vulnerable) | block |
| `/\/\evil.com` | `https://evil.com/` | allow | block |
| `/<TAB>/evil.com` | `https://evil.com/` | allow | block |
| `//evil.com` | `https://evil.com/` | block | block |

## Fix

Rewrote `isSafeRedirectTarget` per the security team's recommendation:

- target must be a rooted relative path (`raw[0] === '/'`) — rejects absolute URLs and bare schemes
- parse with `new URL(raw, BASE)` against a fixed base (`https://steemit.com`)
- require `url.origin === BASE`

The WHATWG parser applies the same normalization as browsers, so every authority-override form (backslashes, mixed slashes, control characters, encoded variants) is rejected by the single origin comparison instead of by enumerating signatures.

## Notes

- `URL` is imported from the `url` core module since `engines` declares `node >= 8.7.0` (global `URL` only exists from Node 10).
- `/%2F%2Fevil.com` now passes: percent-encoded slashes stay same-origin path content (`https://steemit.com/%2F%2Fevil.com`); the previous rejection was overly strict.
- Same-origin absolute forms such as `//steemit.com` pass the origin check and are harmless.
- No call-site changes needed (`src/server/server.js:246`).

## Test plan

- [x] `npx jest src/server/__tests__/redirect.test.js` — 6/6 pass, including new backslash/control-char bypass cases
- [x] `npx eslint` clean on both changed files
- [x] Side-by-side old-vs-new validation run against bypass payloads (table above)

### DIFF
--- a/src/server/__tests__/redirect.test.js
+++ b/src/server/__tests__/redirect.test.js
@@ -5,8 +5,21 @@ describe('isSafeRedirectTarget', () => {
     it('rejects protocol-relative targets', () => {
         expect(isSafeRedirectTarget('//evil.com')).toBe(false);
         expect(isSafeRedirectTarget('%2F%2Fevil.com')).toBe(false);
-        expect(isSafeRedirectTarget('/%2F%2Fevil.com')).toBe(false);
         expect(isSafeRedirectTarget('///evil.com')).toBe(false);
+    });
+
+    it('rejects backslash forms the browser treats as authority', () => {
+        // `\` is normalized to `/` before URL parsing, so these are
+        // protocol-relative in every browser even though no literal
+        // `//` appears — the gap reported by the security team.
+        expect(isSafeRedirectTarget('/\\evil.com')).toBe(false);
+        expect(isSafeRedirectTarget('\\evil.com')).toBe(false);
+        expect(isSafeRedirectTarget('/\\/\\evil.com')).toBe(false);
+        expect(isSafeRedirectTarget('/\\//evil.com')).toBe(false);
+        // tabs/newlines are stripped by the URL parser just like by browsers
+        expect(isSafeRedirectTarget('/\t/evil.com')).toBe(false);
+        expect(isSafeRedirectTarget('/\\\nevil.com')).toBe(false);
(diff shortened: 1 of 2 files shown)


ety001:
Review feedback addressed in e11b9e42:

1. **Origin constant** — comparison now uses a precomputed `BASE_ORIGIN = new URL(BASE).origin` instead of the raw `BASE` string, so a future BASE edit (trailing slash, cased host) cannot silently break it.
2. **Fixed-BASE rationale** — commented above the constant: the middleware only ever emits rooted relative paths, which resolve against any parse base, so the anchor never rejects legitimate staging/local traffic and exists purely to detect authority overrides; deriving it from the request `Host` header would validate against an attacker-controllable value and re-open the bug.
3. **Unrooted-but-normalizing tests** — added `\t//evil.com`, `\n/\evil.com`, `' //evil.com'` (leading whitespace stripped by browsers → protocol-relative) plus a fail-closed boundary case `\t/trending`, confirming `raw[0] === '/'` rejects before any parsing.

On the non-blocking Node note: CI (CircleCI) only builds the Docker images, so runtime Node comes from the Dockerfiles — prod `node:10`, dev `node:12-alpine`. Both have global `URL`; the `url` core-module import additionally covers Node 8.x per `engines`. Documented at the import line.

`npx jest src/server/__tests__/redirect.test.js` → 7/7 pass, eslint clean.

kuny0707 (approved):

PR #4011, opened 2026-09-01
fix(security): close open redirect via tag-route normalization 301s (F47 follow-up #2)

ety001:
## Summary

Second follow-up to the F47 open redirect finding, implementing the security team's proposed fix (all four points) plus one latent bug fix found during review.

## Root cause

PR #4008 guarded the `ch/cn/r` param-stripping redirect exit only. The two lowercase-normalization 301 blocks in `server.js` were unguarded, and the `<tag>` route token admitted path separators:

- `ResolveRoute.js` expanded `<tag>` to `([\w\W\d-]{1,32})` — `[\w\W]` matches **every** character, including `/` and `\`
- `GET //evil.com/@a/B` therefore matched `routeRegex.Post` with `tag=/evil.com`
- the user-normalization block lowercased it (`B`→`b`, so `p !== originalUrl`) and issued `301 Location: //evil.com/@a/b`
- the browser resolved that to `https://evil.com/@a/b` — open redirect
- `/\evil.com/@a/B` (backslash variant) worked identically

Verified against the actual regex construction before fixing:

| Input | before | after |
|---|---|---|
| `//evil.com/@a/B` (Post) | match → 301 off-site | no match |
| `/\evil.com/@a/B` (Post) | match → 301 off-site | no match |
| `/\/\evil.com/@a/B` (Post) | match | no match |
| `//evil.com/@a/b.json` (PostJson) | match | no match |
| `/hive-148441/@ety001/Some-Post` | match | match ✓ |
| `/trending/HIVE-148441` (CategoryFilters) | match | match ✓ (301 normalization preserved) |

## Changes (defense in depth — either layer blocks independently)

1. **`src/app/ResolveRoute.js`** — tighten `<tag>` to `([\w.-]{1,32})`; no path separators. Covers all chain-legal tags (`[a-z0-9-]`, `hive-NNNNNN`); uppercase kept on purpose so the lowercase 301 normalization still fires. Note this regex is shared with `resolveRoute()` (the actual router), so route behavior and the redirect gate stay consistent.
2. **`src/server/server.js`** — both normalization blocks now require `isSafeRedirectTarget(p)` (`server.js:216`-ish and `:227`-ish). Any future `<tag>`-style regex regression can no longer turn these 301s into exits.
3. **`src/server/app_render.jsx`** — the two router-driven `ctx.redirect` exits (`:153` redirectUrl, `:195` onEnter error path) guarded with `isSafeRedirectTarget`; unsafe targets fall through to normal render / rethrow instead of leaving the site.
4. **`src/server/app_render.jsx` TDZ fix** — `console.error('AppRender error', err, redirect)` referenced `redirect` before its `const` declaration, so every caught render error re-threw a `ReferenceError` and the redirect/error handling below it was dead code. Destructure reordered.
5. **Tests** — separator-swallowing regressions (Post/PostJson/CategoryFilters payloads), legitimate-tag positives (uppercase `hive-148441` normalization pinned), regex-pattern snapshot updated.

## Notes

- Behavior change: tag URLs with chars outside `[\w.-]` (non-ASCII, `%`, spaces) now 404 instead of rendering an empty listing — such tags are not chain-valid anyway.
- `redirects.js` keeps its deliberate external redirect (`/pick_account` → `https://signup.steemit.com`): hardcoded constant, no user input, intentionally **not** guarded — blanket-applying `isSafeRedirectTarget` there would break signup.
- eslint crashes (`Cannot read properties of null (reading 'range')`) on `server.js`/`app_render.jsx` and the 22 pre-existing lint errors in `ResolveRoute.test.js` are present on master unchanged — tooling debt, not introduced here.

## Test plan

- [x] `npx jest src/app/ResolveRoute.test.js src/server/__tests__/redirect.test.js` — 37/37
- [x] `npx jest` (full suite) — 197/197, 21 suites
- [x] eslint clean on `ResolveRoute.js`, `RedirectTarget.js`, `redirect.test.js`
- [x] Old-vs-new regex comparison table above run against real pattern construction

### DIFF
--- a/src/app/ResolveRoute.js
+++ b/src/app/ResolveRoute.js
@@ -11,7 +11,13 @@ const reg = pattern => {
             '<sort>',
             '(hot|trending|promoted|payout|payout_comments|muted|created)'
         )
-        .replace('<tag>', '([\\w\\W\\d-]{1,32})')
+        // <tag> must not admit path separators: a class like [\w\W] matches
+        // every char, letting `//evil.com/@a/b` pose as tag=/evil.com and
+        // reach the lowercase-normalization 301 (open redirect). Letters,
+        // digits, underscore, dot and hyphen cover every chain-legal tag
+        // (incl. hive-NNNNNN communities); case is kept on purpose so the
+        // 301 still normalizes /trending/HIVE-148441.
+        .replace('<tag>', '([\\w.-]{1,32})')
         .replace('<permlink>', '([\\w\\d-]+)')
         .replace('/', '\\/');
     return new RegExp('^\\/' + pattern + '$');
--- a/src/app/ResolveRoute.test.js
+++ b/src/app/ResolveRoute.test.js
@@ -11,13 +11,13 @@ describe('routeRegex', () => {
             ],
             [
                 'CategoryFilters',
-                /^\/(hot|trending|promoted|payout|payout_comments|muted|created)(?:\/([\w\W\d-]{1,32}))?\/?$/,
+                /^\/(hot|trending|promoted|payout|payout_comments|muted|created)(?:\/([\w.-]{1,32}))?\/?$/,
             ],
             ['PostNoCategory', /^\/(@[\w\.\d-]+)\/([\w\d-]+)$/],
-            ['Post', /^\/([\w\W\d-]{1,32})\/(@[\w\.\d-]+)\/([\w\d-]+)\/?$/],
+            ['Post', /^\/([\w.-]{1,32})\/(@[\w\.\d-]+)\/([\w\d-]+)\/?$/],
             [
                 'PostJson',
-                /^\/([\w\W\d-]{1,32})\/(@[\w\.\d-]+)\/([\w\d-]+)(\.json)$/,
+                /^\/([\w.-]{1,32})\/(@[\w\.\d-]+)\/([\w\d-]+)(\.json)$/,
             ],
             ['UserJson', /^\/(@[\w\.\d-]+)(\.json)$/],
         ];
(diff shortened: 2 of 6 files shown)


ety001:
Review feedback addressed in 844dca66 (test/comment only, no runtime change — production `pr4011-test1-20260901055500` already runs the identical logic):

1. **Backslash/protocol-relative payload shapes** — added `/\evil.com/@a/b` (false), `//evil.com/@a/b` (false), `/\evil.com?x=1` (false). Note `/%2F%2Fevil.com` → true and `//evil.com` → false were already pinned in the suite; the encoded-path acceptance now also has a full-path variant (`/%2F%2Fevil.com/@a/b`) and a comment naming the spec-level reason (browser never decodes percent-encoded chars into an authority).

2. **Parser-contract comment** — `RedirectTarget.js` doc now states: Node's WHATWG URL (from the `url` core module; global only from Node 10, engines declares >= 8.7.0) implements the same URL Standard browsers apply to the Location header, and the tests pin that equivalence so parser divergence or a non-compliant runtime fails loudly.

3. **Query/fragment behavior** — new test block mirroring the server.js path (p = full lowercased URL incl. query/fragment): `/trending/HIVE-148441?ref=//evil.com` → true, `/trending/HIVE-148441#//evil.com` → true, `/hive-148441/@a/b?next=%2F%2Fevil.com#f` → true. `//` inside query/fragment is same-origin content, not an authority switch — confirmed and pinned.

4. **supertest integration test (optional)** — deliberately skipped for now: importing `server.js` into jest boots the config/steem-js/session chain (the reason `RedirectTarget.js` documents for testing the module directly), and the middleware is an anonymous generator inside the app setup, so a real integration test needs a refactor to export it. The two layers (routeRegex gate + isSafeRedirectTarget exit) are tested independently, the glue is a one-line `&&`, and the payload was verified end-to-end against the deployed production instance. Happy to do the export-middleware refactor as a follow-up if you want the integration coverage in-repo.

`npx jest src/server/__tests__/redirect.test.js` → 11/11, eslint clean.

kuny0707 (approved):

PR #4031, opened 2026-09-20
feat(settings): optimistic profile overlay to cover the hivemind indexing lag

ety001:
## Summary

Extends the pending-overlay write→read convergence loop (PR #4026, documented in `docs/VOTE_STATE_FLOW.md`) to **profile saves**. Previously, after saving profile settings, a refresh could still show the old profile until hivemind re-indexed the account row and both cache layers expired. This PR closes that window with the same pattern used for votes.

## Problem

`https://steemitdev.com/@ety001.test01/settings` — after a successful profile update broadcast, the page (and the public profile page) keep rendering the old profile during:

1. the hivemind indexing lag (hivemind re-reads the steemd account row on its own schedule), and
2. the server cache TTL (`steem:profile:{account}`) plus the browser L1 SWR entries.

## Solution

Same four-layer closed loop as votes: local state → L1 invalidation → L2 invalidation → server-side pending overlay.

**Write path** (`app/api/steem/broadcast/route.ts`)

- New 5th overlay key `steem:pendingprofile:{account}`: on a successful `account_update2` broadcast, parse `posting_json_metadata` from the op and record the complete new `profile` sub-object via `recordPendingProfile()` (120s TTL).
- Actor extraction fix: `account_update2` ops never matched `voter`/`author`, so the response never carried the `X-Cache-Invalidate` account token. It now reads `op.account`, so browser L1 profile entries evict immediately after the save — same channel as votes.

**Read path** (`lib/steem/client.ts` `getProfile`)

- The overlay merges **after** the cache layer on both the cached and observer-bypass paths, so cached entries keep holding pure chain data. Covers the profile page banner, settings page, and `generateMetadata`.

**Merge semantics** (profile-specific equivalent of the vote sign check)

- **Replace, not patch**: the op carries the complete new `metadata.profile`, so the overlay swaps the sub-object wholesale — fields the user cleared disappear during the window too. `stats` / `reputation` / everything outside `metadata.profile` stays chain-owned.
- **Convergence**: once the chain-side profile equals the saved one (same keys, same values), the overlay is a no-op. A second save rewrites the key. After the 120s TTL the chain data has necessarily landed.
- **Failure modes** (identical philosophy to votes): malformed `posting_json_metadata` → no overlay recorded; Redis off → no-op everywhere; overlay read error → chain data as-is.

## Docs

`docs/VOTE_STATE_FLOW.md`: new `account_update2` row in the write-effects table, new "Profile save overlay semantics" section, sequence diagram and key-files table updated.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run test` — 290 cases across 41 files pass, including 7 new overlay-merge cases and 3 new broadcast-route cases
- [x] `pnpm run lint` — 0 errors (20 warnings are pre-existing in untouched files)
- [x] `pnpm run build`
- [ ] Manual: save profile on `/@user/settings`, hard-refresh both `/@user` and `/@user/settings` within 2 minutes — the new profile shows on both


ety001:
## Self-audit follow-up

An independent audit pass of this PR (full model, repo-wide read) returned **request-changes** with one major finding, which is now fixed in b995d0aa:

**MAJOR — empty/absent `posting_json_metadata` blanked the profile for 120s.** `account_update2` treats an empty metadata field as "leave unchanged" (optional-field semantics — exactly what key-only updates like memo_key rotation emit). The original code recorded a `{}` overlay for those ops, and `applyProfileOverlay` would then replace `metadata.profile` with `{}` until the TTL expired. Fix: record the overlay only when the parsed metadata actually carries a plain-object `profile`; a `!Array.isArray` guard was also added on both the write and read side (`typeof [] === 'object'`).

Also addressed from the audit:
- Doc comment on `recordPendingProfile` corrected — the sub-object is "as broadcast" (posting-authority-signed, same trust as chain data), not "sanitized".
- `docs/VOTE_STATE_FLOW.md` failure-mode list now covers empty/absent metadata and missing-profile-key.
- Tests added: empty metadata, missing profile key, array profile, second-save rewrite (newest wins), corrupted overlay JSON, read-side array guard.

Audit items noted but intentionally not changed (pre-existing patterns shared with the vote overlay, no regression from this PR):
- Convergence check uses reference equality, so nested-object profile values keep the overlay replacing with semantically identical data until TTL (benign).
- `cacheDeleteByPrefix`/L1 substring invalidation over-matches accounts sharing a name prefix (over-invalidation only).
- `website` href / `cover_image` scheme validation at render is a pre-existing chain-path exposure worth separate hardening.
- `getProfile` itself has no unit test file (merge placement verified by reading both call paths).

Verification: `pnpm run typecheck` clean, 296/296 tests pass, `pnpm run lint` 0 errors (20 pre-existing warnings in untouched files).

